PR #1, opened 2019-07-31
copy the code from cli-experimental to cli-utils

Liujingfang1:
Keep the git history of cli-experimental

The last commit is to change the import path from cli-experimental to cli-utils


Liujingfang1:
@barney-s

PR #3, opened 2019-08-08
Readme

Liujingfang1:
This change is to add examples for using apply, status and prune


Liujingfang1:
@barney-s

docs/status.md
```diff
@@ -0,0 +1,58 @@
...
+```
+cli-utils apply config/hello
+```
+
+Then, get the status of the applied resources
+```
+cli-utils apply status config/hello
+```
+
+The output is like the following
```
pwittrock: what is the exit code if it is done vs not done?

docs/prune.md
```diff
@@ -0,0 +1,126 @@
...
+```
+inventory:
+  type: ConfigMap
+  configMap:
+    name: prune-cm-name
+    namespace: some-namespace
+```
+Then the output of the kustomization directory will contain a ConfigMap resource named `prune-cm-name`, which contains a list of all resources in this kustomization directory. 
+
+When resources are removed from this kustomization directory, the `prune-cm-name` will contains a different list of resources. The `prune` compares the old list with the new list to get a diff and deletes the obsolete resources.
```
pwittrock: nit: s/will contains/will contain/

docs/prune.md
```diff
@@ -0,0 +1,126 @@
+# prune
+
+```
+cli-utils prune <dir>
+```
+can prune obsolete resources from a kustomization directory or resources that are with declarative deletion annotation.
+
+## prune kustomize directory
```
pwittrock: Introduce that there are 2 methods for deleting resources:

- prune by lifecycle annotations (preferred)
  - explicit deletion of resources by annotating them with deletion
  - will not delete generated secrets and configmaps
- prune by inventory configmap
  - implicit deletion of resources by removing the resource from the configuration
  - will delete generated secrets and configmaps

docs/status.md
Liujingfang1: The exit code is 0 but the not done case will show the failure.

docs/prune.md
Liujingfang1: prune by inventory configmap doesn't delete generated secrets and configmaps if they are/were refereed in other objects.

docs/prune.md
Liujingfang1: updated

docs/prune.md
Liujingfang1: updated

docs/status.md
```diff
@@ -0,0 +1,58 @@
...
+```
+2019/08/07 16:38:04 nil value at `volumeClaimTemplates[].metadata.labels` ignored in mutation attempt
```
pwittrock: what is this?

docs/prune.md
```diff
@@ -0,0 +1,126 @@
...
+
+The output is like the following
+```
+NAME                                           READY   STATUS    RESTARTS   AGE
+myapp-pod                                      1/1     Running   0         1m
+```
+
+Then run the prune subcommand by
+```
+cli-utils prune config/helloWithInventory
```
pwittrock: is there a dry-run that prints out what will be deleted?

docs/status.md
Liujingfang1: Ah, this is some warning message when applying the example manifests. I can remove it.

docs/prune.md
Liujingfang1: not yet.

PR #7, opened 2019-12-12
WIP: Prune object reference

seans3:
* POC for prune
* This will NOT be submitted
* Prunes by owners references to grouping object

### DIFF
--- a/helloWorld/README.md
+++ b/helloWorld/README.md
@@ -1,0 +1,311 @@
+[base]: ../../docs/glossary.md#base
+[config]: https://github.com/kinflate/example-hello
+[gitops]: ../../docs/glossary.md#gitops
+[hello]: https://github.com/monopole/hello
+[kustomization]: ../../docs/glossary.md#kustomization
+[original]: https://github.com/kinflate/example-hello
+[overlay]: ../../docs/glossary.md#overlay
+[overlays]: ../../docs/glossary.md#overlay
+[patch]: ../../docs/glossary.md#patch
+[variant]: ../../docs/glossary.md#variant
+[variants]: ../../docs/glossary.md#variant
+
+# Demo: hello world with variants
+
+Steps:
+
+ 1. Clone an existing configuration as a [base].
+ 1. Customize it.
+ 1. Create two different [overlays] (_staging_ and _production_)
+    from the customized base.
+ 1. Run kustomize and kubectl to deploy staging and production.
+
+First define a place to work:
+
+<!-- @makeWorkplace @testAgainstLatestRelease -->
+```
+DEMO_HOME=$(mktemp -d)
+```
+
+Alternatively, use
+
+> ```
+> DEMO_HOME=~/hello
+> ```
+
+## Establish the base
+
+Let's run the [hello] service.
+
+To use [overlays] to create [variants], we must
+first establish a common [base].
+
+To keep this document shorter, the base resources are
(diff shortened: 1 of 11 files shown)


helloWorld/README.md
```diff
@@ -0,0 +1,311 @@
+[base]: ../../docs/glossary.md#base
```
Liujingfang1: Is the whole folder helloWorld copied from kustomize? Can you change the README.md to best serve the use case here?

internal/pkg/apply/apply.go
```diff
@@ -43,89 +48,294 @@ type Apply struct {
...
 	Commit *object.Commit
+
+	// Flag; deletes objects previously applied, but not applied currently.
+	Prune bool
+
+	// Flag; don't actually persist created/updated/deleted resources.
+	DryRun bool
+
+	// List of previous inventory objects associated with director.
+	PrevInventory []unstructured.Unstructured
```
Liujingfang1: Suggest change []unstructured.Unstructured to []*unstructured.Unstructured since the sources are defined as that type.

internal/pkg/apply/apply.go
```diff
@@ -43,89 +48,294 @@ type Apply struct {
...
+						fmt.Fprintf(a.Out, "Failure creating inventory director: %v\n", err)
+						break
+					}
+				}
+				a.getPrevInventory(director)
+			}
+			directorRef := createOwnerReference(director)
+			a.addOwnerReference(u, *directorRef)
+			suffix := createUniqueId()
+			u.SetName(u.GetName() + "-" + suffix)
```
Liujingfang1: Based on the change, there is a new suffix created and appended to `u` every time the apply is called. You can use the hash annotation in the Inventory object to skip creating a new `u` when nothing is changed in the apply list.

internal/pkg/apply/apply.go
```diff
@@ -43,89 +48,294 @@ type Apply struct {
...
+	return director, nil
+}
+
+func (a *Apply) createInventoryDirector(inv *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+	director := &unstructured.Unstructured{}
+	director.SetGroupVersionKind(inv.GroupVersionKind())
+	director.SetNamespace(inv.GetNamespace())
+	director.SetName(inv.GetName())
+	inventoryId := createUniqueId()
+	annotations := map[string]string{InventoryId: inventoryId}
```
Liujingfang1: Why not using DeepCopy and then reset the annotations?
```
director := inv.DeepCopy()
inventoryId := createUniqueId()
annotations :=  map[string]string{InventoryId: inventoryId}
director.SetAnnotations(annotations)
```

internal/pkg/apply/apply.go
```diff
@@ -43,89 +48,294 @@ type Apply struct {
...
-	annotations := newObj.GetAnnotations()
-	err = newInv.UpdateAnnotations(annotations)
-	if err != nil {
-		return nil, err
+	return director, nil
+}
+
+func (a *Apply) prune() {
+	for _, inv := range a.PrevInventory {
+		fmt.Fprintf(a.Out, "[deleted:meta] %s/%s\n", inv.GetKind(), inv.GetName())
```
Liujingfang1: The pruning only prints the message for deleting the inventory object. Can we add some explanation that it also deletes the objects that are only owned by this inventory object?

PR #9, opened 2020-01-29
Adds kapply and kstatus packages

seans3:
* Adds `pkg/apply` as new apply code with new prune. 
* Adds `pkg/kstatus` as new status code used in `apply`. 
* Adds unit tests for each of these two new packages.
* Adds a `go.mod` and `go.sum` file at the root of the repository.

/kind cleanup
/sig cli
/priority important-soon

```release-note
NONE
```


mortent:
I find this structure confusing.

We are nesting go modules within each other. I don't actually know how this will work, i.e. will kstatus be considered part of the cli-utils module as well as an independent module? Also, the other modules end up inside the pkg folder, which isn't a place I would expect to find new go modules.

This mixes the old code from cli-utils with the new code we are moving from Kustomize. If we need to keep both around in the same repo, I think we should try to keep them fully separated into individual go modules, all of which should be in a folder on the root of the repo. Something similar to how multiple modules are organized in Kustomize.

PR #18, opened 2020-02-03
Adds file comments to grouping.go

seans3:
* Adds comments at the top of `grouping.go` to explain what this file does.

/sig cli
/priority important-soon

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 main
 cli-utils
 dyctl
+kapply
 
 #intellij
 .idea/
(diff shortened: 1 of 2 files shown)


mortent:
We probably don't want to commit the kapply binary. Can you remove it from the commit and add it to the .gitignore file?

seans3:
> We probably don't want to commit the kapply binary. Can you remove it from the commit and add it to the .gitignore file?

Fixed.

PR #19, opened 2020-02-03
New observer library for status and new status command

mortent:
This adds a new library for computing status of resources based on polling from the cluster. It is in the `observe` package and is meant to replace the `wait` package. It has some key improvements mostly based on feedback from Shopify, including the ability to compute status based on additional resources (like use information from ReplicaSet to compute status for Deployment) and pre-fetch resources before each polling cycle with `LIST` calls. It allows separating the rules for different resource types by implementing different `observers`, each of which will implement the `ResourceObserver` interface.

This PR just contains the basic framework. There will be several follow-up PRs that provides implementations for the interfaces defined here. This approach is based on feedback that the original PR was too big.

@seans3 @pwittrock @monopole @KnVerey


cmd/status/printers/table_printer.go
```diff
@@ -0,0 +1,343 @@
...
+				default:
+					return fmt.Fprintf(w, "%dh", integer.RoundToInt32(age.Round(time.Hour).Hours()))
+				}
+			},
+		},
+		{
+			name:   "message",
+			header: "MESSAGE",
+			width:  40,
+			content: func(w io.Writer, availableWidth int, resource *common.ObservedResource) (int, error) {
```
monopole: can resource be nil (as in L133?)

go.mod
```diff
@@ -8,11 +8,15 @@ require (
...
+	gopkg.in/yaml.v3 v3.0.0-20200121175148-a6ecf24a6d71
+	gotest.tools v2.2.0+incompatible
 	k8s.io/api v0.17.2
 	k8s.io/apimachinery v0.17.2
 	k8s.io/cli-runtime v0.17.2
 	k8s.io/client-go v0.17.2
 	k8s.io/kubectl v0.0.0-20191219154910-1528d4eea6dd
+	k8s.io/utils v0.0.0-20191114184206-e782cd3c129f
 	sigs.k8s.io/controller-runtime v0.4.0
+	sigs.k8s.io/kustomize/kyaml v0.0.9
```
monopole: 0.0.10 is latest

cmd/status/printers/color.go
```diff
@@ -0,0 +1,11 @@
+package printers
```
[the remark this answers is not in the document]
mortent: Agree this might be useful elsewhere, but let's move it when we need to.

cmd/status/status.go
```diff
@@ -0,0 +1,126 @@
...
+			IncludeSubpackages: r.IncludeSubpackages,
+		})
+	}
+	if len(inputs) == 0 {
+		inputs = append(inputs, &kio.ByteReader{Reader: c.InOrStdin()})
+	}
+
+	err = kio.Pipeline{
+		Inputs:  inputs,
+		Filters: filters,
```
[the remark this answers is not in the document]
mortent: No, since this is a pipeline, both the input and output needs to be yaml. It doesn't really fit the model here where we want to capture some data from each resource manifest coming through.

cmd/status/util.go
```diff
@@ -0,0 +1,62 @@
...
+		if err != nil {
+			return nil, err
+		}
+		var namespace string
+		if mapping.Scope.Name() == meta.RESTScopeNameNamespace && id.Namespace == "" {
+			namespace = "default"
+		} else {
+			namespace = id.Namespace
+		}
+		if IsValidKubernetesResource(id) {
```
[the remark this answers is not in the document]
mortent: Maybe, but not clear how we want to handle it. We don't want to print here. This only checks for the basics that needs to be in every resource manifest, like apiVersion and kind.

mortent:
@barney-s 
It might be possible to reduce the boilerplate in the observers. It is something we can iterate on when we start implementing more complex rules for more resource types. Now I'm just trying to find the right trade-off between reducing boilerplate but still keep it readable.
Like I mentioned above, I prefer to move things when we actually need them elsewhere.
The feature to populate a cache with LIST calls is based on suggestions from Shopify. Even with an aggressive polling interval, using broad LIST calls can significantly reduce the number of calls to the API server. Trade-off is that each call might return a large number of resources.

mortent:
@monopole  I have reduced the size of the PR somewhat by only keeping the core observer library. So the collector and the new command is no longer here and I'll do a follow up PR with that. I hope this makes it easier, even if it is still a sizable PR. Breaking it up any further gets hard because there are some dependencies between the packages.

monopole:
> Breaking it up any further gets hard because there are some dependencies between the packages

then they are really one package, and the multiple package arrangement isn't doing us any good

pkg/kstatus/observe/reader/cached_reader.go
```diff
@@ -0,0 +1,205 @@
...
+	for _, u := range cacheList.Items {
+		if selector.Matches(labels.Set(u.GetLabels())) {
+			items = append(items, u)
+		}
+	}
+	list.Items = items
+	return nil
+}
+
+// ListNamespaceScoped lists all resource identifier by the GVK of the list and selector
```
pwittrock: nit: comment has wrong name

pkg/kstatus/observe/observe.go
```diff
@@ -0,0 +1,309 @@
...
+// Observe will create a new StatusObserverRunner that will poll all the resources provided and report their status
+// back on the event channel returned. The StatusObserverRunner can be cancelled at any time by cancelling the
+// context passed in.
+// If stopOnCompleted is set to true, then the runner will stop observing the resources when the StatusAggregator
+// determines that all resources has been fully reconciled. If this is set to false, the observer will keep running
+// until cancelled.
+// If useCache is set to true, the observer will fetch all resources needed using LIST calls before each polling
+// cycle. The observers responsible for computing status will rely on the cached data. This can dramatically reduce
+// the number of calls against the API server.
+func (s *StatusObserver) Observe(ctx context.Context, resources []wait.ResourceIdentifier, stopOnCompleted bool, useCache bool) <-chan Event {
```
pwittrock: nit: consider making `useCache` a field on StatusObserver

pkg/kstatus/observe/observers/common.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	u.SetNamespace(identifier.Namespace)
+	return &u, nil
+}
+
+// ObservedGeneratedResources provides a way to fetch the statuses for all resources of a given GroupKind
+// that match the selector in the provided resource. Typically, this is used to fetch the status of generated
+// resources.
+func (b *BaseObserver) ObserveGeneratedResources(ctx context.Context, observer ResourceObserver, object *unstructured.Unstructured,
+	gk schema.GroupKind, selectorPath ...string) (common.ObservedResources, error) {
```
KnVerey: Could the observer expose which kind it is handling? It looks like technically you could pass this method e.g. a replicaSetObserver with kind="configmap" or whatever, which wouldn't ever make sense afaict. Even the selector path is basically an intrinsic property of the kind, though storing _that_ on the observer definitely feels wrong. 🤔

pkg/kstatus/observe/reader/cached_reader.go
```diff
@@ -0,0 +1,205 @@
...
+// from the cache. If the needed combination of GVK and namespace is not part of the cache,
+// that is considered an error.
+func (c *CachedObserverReader) ListClusterScoped(ctx context.Context, list *unstructured.UnstructuredList, selector labels.Selector) error {
+	return c.ListNamespaceScoped(ctx, list, "", selector)
+}
+
+// Sync loops over the list of gvkNamespace we know of, and uses list calls to fetch the resources.
+// This information populates the cache.
+func (c *CachedObserverReader) Sync(ctx context.Context) error {
+	c.Lock()
```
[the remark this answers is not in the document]
mortent: Each call to `StatusObserver.Observe` will create a new `StatusObserverRunner` that will run in its own go routine. So with there is no way `Sync` can be called concurrently since both the `ObserverReader` and the observers are local to the `StatusObserverRunner`. Everything that happens in the runner is confined to a single goroutine.

Currently all the observers are also running in that same goroutine, so we don't strictly need any synchronization here at all. I am thinking about running multiple observers in parallell when using the cached `ObserverReader`, but I haven't gotten to it yet. It would involve some kind of task queue and a limited number of goroutines to avoid creating too many of them. Either way, the plan is that `Sync` will only ever be called by the main goroutine for the runner.

If we are not using the cached `ObserverReader`, running many observers in parallell just means more load on the APIServer, so it doesn't seem like such a good idea.

pkg/kstatus/observe/common/common.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/wait"
+)
+
+// GetNamespaceForNamespacedResource returns the namespace for the given object,
+// but includes logic for returning the default namespace if it is not set.
```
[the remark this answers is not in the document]
mortent: Yeah, I haven't found a great way to handle this yet. But this is here to handle the situation where we are reading manifests from file that doesn't set the namespace. This is currently allowed with kubectl and the resource will then be put in the default namespace. I'm probably leaning towards moving this issue to whichever process is reading the manifests, and just error out inside this library if we find identifiers for namespaced resources that has an empty namespace.

pkg/kstatus/observe/aggregator/aggregator.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"sigs.k8s.io/cli-utils/pkg/kstatus/observe/common"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/wait"
+)
+
+// AllCurrentOrNotFoundStatusAggregator is an implementation of the StatusAggregator interface
+// that considers the aggregate status as Current whenever all resources have either the Current
+// or the NotFound status.
```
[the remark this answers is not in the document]
mortent: This is another "corner case" where I haven't fully figured out what is the best solution. NotFound is treated as Current here to handle situations with deletions gracefully. When we have an apply function that includes pruning or a teardown/destroy command, we will consider things fully reconciled when all resources have been deleted. But I agree, it seems counterintuitive that NotFound is considered Current.

pkg/kstatus/observe/observers/common.go
```diff
@@ -0,0 +1,140 @@
...
+// SetComputeStatusFunc allows for setting the function used by the observer for computing status. The default
+// value here is to use the status package. This is provided for testing purposes.
+func (b *BaseObserver) SetComputeStatusFunc(statusFunc ComputeStatusFunc) {
+	b.computeStatusFunc = statusFunc
+}
+
+// LookupResource looks up a resource with the given identifier. It will use the rest mapper to resolve
+// the version of the GroupKind given in the identifier.
+// If the resource is found, it will be returned and the observedResource will be nil. If there is an error
+// or the resource is not found, the observedResource will be returned containing information about the resource
```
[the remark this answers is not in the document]
mortent: I agree this isn't great. The advantage of doing it this way is that it allows more of the error handling to be pushed down into the `LookupResource` function and reduce the boilerplate in the observers. I'm still trying to find the best way to reduce boilerplate in the observers and still make them somewhat readable. The `ObserveGeneratedResources` function has the same issue.

pkg/kstatus/observe/observers/deployment.go
```diff
@@ -0,0 +1,78 @@
...
+		return observedResource
+	}
+	return d.ObserveObject(ctx, deployment)
+}
+
+func (d *DeploymentObserver) ObserveObject(ctx context.Context, deployment *unstructured.Unstructured) *common.ObservedResource {
+	identifier := toIdentifier(deployment)
+
+	observedReplicaSets, err := d.ObserveGeneratedResources(ctx, d.RsObserver, deployment,
+		appsv1.SchemeGroupVersion.WithKind("ReplicaSet").GroupKind(), "spec", "selector")
```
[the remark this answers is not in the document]
mortent: This looks up and observe the status for the replicasets, each of which will then look up the pods. This allows the deployment observer to use this information to improve the status (for example, if one or more pods are not Ready, check if they are stuck in some unrecoverable state) and it will also pass references to the generated resources in the event sent back on the go channel. This information is used to show the hierarchy of resources in the screen shot in the PR description.
I have considered looking at the owner references too. Although I think in practice they should always match since I think the controllers will "adopt" resources that matches the selector. But need to verify that.
When you are mentioning current generation here, do you mean the newest replicaset? I agree those are most useful, but the old ones can be interesting to visualize how pods are updated during the rollout.

pkg/kstatus/observe/observers/common.go
mortent: I think letting the function fetch the GroupKind from the observer makes sense. I'll try that out.
The selector path is a property of the `Deployment` kind here, so I think it makes sense to have this information in the Deployment observer. When we had add support for additional resources like Job (which has an optional selector) and CronJob (which have no selector), we will need to see if there is a better way to handle looking up resources of different kinds. Same for looking up all pods that are targeted by a Service.

pkg/kstatus/observe/common/common.go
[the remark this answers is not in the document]
KnVerey: >  moving this issue to whichever process is reading the manifests, and just error out inside this library if we find identifiers for namespaced resources that has an empty namespace.

Makes sense to me

pkg/kstatus/observe/observers/common.go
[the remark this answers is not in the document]
KnVerey: hm, I see. What if we made a function on the base observer that converts errors into observations and called that from both of those places? e.g.:

```go
func (b *BaseObserver) observedResourceError(identifier wait.ResourceIdentifier, err error) *common.ObservedResource {
	if errors.IsNotFound(err) {
		return &common.ObservedResource{
			Identifier: identifier,
			Status:     status.NotFoundStatus,
			Message:    "Resource not found",
		}
	}
	return &common.ObservedResource{
		Identifier: identifier,
		Status:     status.UnknownStatus,
		Error:      err,
	}
}

func (b *BaseObserver) LookupResource(ctx context.Context, identifier wait.ResourceIdentifier) (*unstructured.Unstructured, error) {
	GVK, err := b.GVK(identifier.GroupKind)
	if err != nil {
		return nil, err
	}

	var u unstructured.Unstructured
	u.SetGroupVersionKind(GVK)
	key := common.KeyForNamespacedResource(identifier)
	if err = b.Reader.Get(ctx, key, &u); err != nil {
		return nil, err
	}

	u.SetNamespace(identifier.Namespace)
	return &u, nil
}
```

```go
func (d *DefaultObserver) Observe(ctx context.Context, identifier wait.ResourceIdentifier) *common.ObservedResource {
	u, err := d.LookupResource(ctx, identifier)
	if err != nil {
		return d.observedResourceError(identifier, err)
	}
	return d.ObserveObject(ctx, u)
}
```

pkg/kstatus/observe/aggregator/aggregator.go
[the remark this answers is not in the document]
KnVerey: > When we have an apply function that includes pruning or a teardown/destroy command

That's an interesting case indeed. Krane actually ignores the torn down resources (other than printing a claim that they were pruned) for the purposes of convergence determination. My assumption is that the service working as intended doesn't generally depend on shutdown, and if it does, that will surface other ways (e.g. in Recreate deployments). But if we can do better here, that would be great. Maybe this lib needs to know which resources it expects to converge vs disappear? Because if a resource you asked to watch is not there, I'm not so sure about assuming it was supposed to be removed.

pkg/kstatus/observe/observers/deployment.go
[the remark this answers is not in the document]
KnVerey: >  Although I think in practice they should always match since I think the controllers will "adopt" resources that matches the selector. But need to verify that.

IIRC the case where that won't be true is overlapping selectors: if a resource is already owned, it will not be adopted.

pkg/kstatus/observe/observers/common.go
mortent: So I think I punt on this for now. I tried to address this by adding a `SupportedGroupKind` function to the `ResourceObserver` interface, but that assumes that each observer only handles one GroupKind. That is not a good assumption as we will need some kind of catch-all observer for CRDs. This observer will probably implement the status primarily based on the conditions Erik has proposed. I agree we should try to make this better, but I need to look more into it.

pkg/kstatus/observe/observers/common_test.go
```diff
@@ -0,0 +1,191 @@
...
+`,
+			gk: schema.GroupKind{
+				Group: "custom.io",
+				Kind:  "Custom",
+			},
+			path:        []string{"spec", "selector"},
+			expectError: true,
+			errMessage:  "no matches for kind",
+		},
+		"error listing replicasets": {
```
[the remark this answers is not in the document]
mortent: Yeah, it should... I have added one now.

pkg/kstatus/observe/observe.go
mortent: Since `useCache` is only used within the `Observe` function, I think it makes sense to keep this as a parameter.

pkg/kstatus/observe/observers/common.go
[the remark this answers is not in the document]
mortent: This looks great. I have updated the PR to use this approach.

mortent:
Updated this PR to only contain the core part of the new library. This should make it easier to review and I'll do the rest of the code in smaller follow-up PRs.

cmd/status/status.go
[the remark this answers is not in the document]
barney-s: You could do it with the output as well. It passes a RNode array.

pkg/kstatus/observe/event/event.go
```diff
@@ -0,0 +1,125 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/wait"
+)
+
+// EventType is the type that describes the type of an Event that is passed back to the caller
+// as resources in the cluster are being observed.
+type EventType string
```
monopole: if u want, you can use [stringer](https://blog.golang.org/generate), a utility Rob Pike wrote to convert int enums to strings for printing

E.g. this
https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/loadrestrictions.go
generates this
https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/loadrestrictions_string.go

i suppose the same would apply to pkg/kstatus/status/status.go   

up to you.  keeps channel content lower, and avoids spelling typos by relying on the iota mechanism for uniqueness

pkg/kstatus/observe/event/event_test.go
```diff
@@ -0,0 +1,246 @@
...
+						Kind:  "Deployment",
+					},
+					Namespace: "default",
+					Name:      "Bar",
+				},
+				Status: status.CurrentStatus,
+			},
+			equal: false,
+		},
+		"same resource with same error": {
```
monopole: test case for different err?

pkg/kstatus/observe/event/event_test.go
```diff
@@ -0,0 +1,246 @@
...
+			},
+			equal: false,
+		},
+	}
+
+	for tn, tc := range testCases {
+		t.Run(tn, func(t *testing.T) {
+			res := DeepEqual(&tc.actual, &tc.expected)
+
+			assert.Equal(t, tc.equal, res)
```
monopole: nice test structure

pkg/kstatus/observe/observer/observer.go
```diff
@@ -0,0 +1,197 @@
...
+	ReaderFactoryFunc     ReaderFactoryFunc
+	ObserversFactoryFunc  ObserversFactoryFunc
+}
+
+// Observe will create a new StatusObserverRunner that will poll all the resources provided and report their status
+// back on the event channel returned. The StatusObserverRunner can be cancelled at any time by cancelling the
+// context passed in.
+// If stopOnCompleted is set to true, then the runner will stop observing the resources when the StatusAggregator
+// determines that all resources has been fully reconciled. If this is set to false, the observer will keep running
+// until cancelled.
```
monopole: can this comment explain why someone would set this false?

pkg/kstatus/observe/observer/observer.go
```diff
@@ -0,0 +1,197 @@
...
+// The pollInterval specifies how often the observer should poll the cluster for the latest state of the resources.
+func (s *Observer) Observe(ctx context.Context, identifiers []wait.ResourceIdentifier, pollInterval time.Duration, stopOnCompleted bool) <-chan event.Event {
```
monopole: linefeeds would be nice for reading and reviewing on a small screen (e.g laptop).
fwiw, i follow the cstyle policy: http://go/cstyle#Line_Length, but don't require others too
(just hope they will)

pkg/kstatus/observe/observer/observer.go
```diff
@@ -0,0 +1,197 @@
...
+type ObserversFactoryFunc func(reader ObserverReader, mapper meta.RESTMapper) (resourceObservers map[schema.GroupKind]ResourceObserver, defaultObserver ResourceObserver)
+
+// Observer provides functionality for polling a cluster for status for a set of resources.
+type Observer struct {
+	Reader client.Reader
+	Mapper meta.RESTMapper
+
+	AggregatorFactoryFunc AggregatorFactoryFunc
+	ReaderFactoryFunc     ReaderFactoryFunc
+	ObserversFactoryFunc  ObserversFactoryFunc
```
monopole: please add a comment on why these members are public (who is expected to mutate them)

pkg/kstatus/observe/observer/observer_test.go
```diff
@@ -0,0 +1,227 @@
...
+			identifiers := tc.identifiers
+
+			observer := Observer{
+				AggregatorFactoryFunc: func(identifiers []wait.ResourceIdentifier) StatusAggregator {
+					return newFakeAggregator(identifiers)
+				},
+				ReaderFactoryFunc: func(_ client.Reader, _ meta.RESTMapper, _ []wait.ResourceIdentifier) (ObserverReader, error) {
+					return testutil.NewNoopObserverReader(), nil
+				},
+				ObserversFactoryFunc: func(_ ObserverReader, _ meta.RESTMapper) (resourceObservers map[schema.GroupKind]ResourceObserver, defaultObserver ResourceObserver) {
```
monopole: s                 o              l     o   n        g    t   hi s    l       i   n   e    is!

pkg/kstatus/observe/observer/resourceobserver.go
```diff
@@ -0,0 +1,38 @@
...
+type ResourceObserver interface {
+	// Observe will fetch the resource identified by the given identifier
+	// from the cluster and return an ObservedResource that will contain
+	// information about the latest state of the resource, its computed status
+	// and information about any generated resources.
+	Observe(ctx context.Context, resource wait.ResourceIdentifier) *event.ObservedResource
+
+	// ObserveObject is similar to Observe, but instead of looking up the
+	// resource based on an identifier, it will use the passed in resource.
+	ObserveObject(ctx context.Context, deployment *unstructured.Unstructured) *event.ObservedResource
```
monopole: deployment?

it might be best to just omit names (allowed in interface definitions)

pwittrock:
@mortent Would you add a `doc.go` and `example_test.go` so that how the library should be used is clearly documented.

pkg/kstatus/observe/event/event.go
```diff
@@ -0,0 +1,125 @@
...
+	return idI.Name < idJ.Name
+}
+
+func (g ObservedResources) Swap(i, j int) {
+	g[i], g[j] = g[j], g[i]
+}
+
+// DeepEqual checks if two instances of ObservedResource are identical. This is used
+// to determine whether status has changed for a particular resource.
+func DeepEqual(or1, or2 *ObservedResource) bool {
```
pwittrock: What does identical mean here?  This doesn't seem to check the `Resource *unstructured.Unstructured` which surprised me.

mortent: Yeah, the comment here should clarify this better. This checks whether the status (either the actual status variable or the message) has changed on the resource or any of the generated resources. The `Resource` is not included in the equality check to avoid having events sent back on the eventChannel due to changes that has nothing to do with the status of the resources.

mortent:
@pwittrock I can add the `doc.go` file in a later PR that will include the actual API exposed from the library. This PR just contains the core framework. I can try setting up an `example_test.go` file, but not sure if is practical for this kind of API.

pkg/kstatus/observe/observer/aggregator.go
```diff
@@ -0,0 +1,20 @@
+package observer
+
+import (
+	"sigs.k8s.io/cli-utils/pkg/kstatus/observe/event"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
+)
+
+// StatusAggregator provides the interface the observer uses to compute the aggregate status.
```
pwittrock: // StatusAggregator is implemented by (resource status reporters?) so the Observer can aggregate their status (across what?)

pkg/kstatus/observe/event/event.go
monopole: Good catch Phil. 
a comment isn't sufficient, the method needs a new name.
the current name suggests equality in the sense of reflect.DeepEquals.

maybe make it a non-static object method on ObservedResource too since there's no need to allow a nil object.

pkg/kstatus/observe/observer/observer.go
monopole: actually,`ObserveUntilReconciled` and `ObserveUntilCancelled` would be clearer.

pkg/kstatus/observe/observer/observer.go
```diff
@@ -0,0 +1,197 @@
...
+// a new ObserverReader for each StatusObserverRunner.
+type ReaderFactoryFunc func(reader client.Reader, mapper meta.RESTMapper, identifiers []wait.ResourceIdentifier) (ObserverReader, error)
+
+// ObserversFactoryFunc defines the signature for the function the Observer will use to
+// create the resource observers and the default observer for each StatusObserverRunner.
+type ObserversFactoryFunc func(reader ObserverReader, mapper meta.RESTMapper) (resourceObservers map[schema.GroupKind]ResourceObserver, defaultObserver ResourceObserver)
+
+// Observer provides functionality for polling a cluster for status for a set of resources.
+type Observer struct {
+	Reader client.Reader
```
pwittrock: Are their sane defaults we can use for these if they aren't specified by the user?

pkg/kstatus/observe/observer/observer.go
```diff
@@ -0,0 +1,197 @@
...
+
+// ReaderFactoryFunc defines the signature for the function the Observer will use to create
+// a new ObserverReader for each StatusObserverRunner.
+type ReaderFactoryFunc func(reader client.Reader, mapper meta.RESTMapper, identifiers []wait.ResourceIdentifier) (ObserverReader, error)
+
+// ObserversFactoryFunc defines the signature for the function the Observer will use to
+// create the resource observers and the default observer for each StatusObserverRunner.
+type ObserversFactoryFunc func(reader ObserverReader, mapper meta.RESTMapper) (resourceObservers map[schema.GroupKind]ResourceObserver, defaultObserver ResourceObserver)
+
+// Observer provides functionality for polling a cluster for status for a set of resources.
```
pwittrock: `for the status of a set`

pkg/kstatus/observe/observer/observer.go
```diff
@@ -0,0 +1,197 @@
...
+			eventChannel <- event.Event{
+				EventType: event.ErrorEvent,
+				Error:     err,
+			}
+			return
+		}
+		observers, defaultObserver := s.ObserversFactoryFunc(observerReader, s.Mapper)
+		aggregator := s.AggregatorFactoryFunc(identifiers)
+
+		runner := &StatusObserverRunner{
```
pwittrock: should StatusObserverRunner be package private visibility?

pkg/kstatus/observe/observer/observer.go
```diff
@@ -0,0 +1,197 @@
...
+
+// StatusObserverRunner is responsible for polling of a set of resources. Each call to Observe will create
+// a new StatusObserverRunner, which means we can keep state in the runner and all data will only be accessed
+// by a single goroutine, meaning we don't need synchronization.
+// The StatusObserverRunner uses an implementation of the ObserverReader interface to talk to the
+// kubernetes cluster. Currently this can be either the cached ObserverReader that syncs all needed resources
+// with LIST calls before each polling loop, or the normal ObserverReader that just forwards each call
+// to the client.Reader from controller-runtime.
+type StatusObserverRunner struct {
+	ctx context.Context
```
pwittrock: plz comment each variable for what it is used for

pkg/kstatus/observe/event/event.go
pwittrock: Should this be `OberservedStateEqual`?

pkg/kstatus/observe/observer/observer.go
```diff
@@ -0,0 +1,197 @@
...
+	eventChannel := make(chan event.Event)
+
+	go func() {
+		defer close(eventChannel)
+
+		observerReader, err := s.ReaderFactoryFunc(s.Reader, s.Mapper, identifiers)
+		if err != nil {
+			eventChannel <- event.Event{
+				EventType: event.ErrorEvent,
+				Error:     err,
```
pwittrock: consider wrapping the error so we can get the stack trace

pkg/kstatus/observe/observer/reader.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ObserverReader is the interface provided to the observers to talk to the cluster.
+type ObserverReader interface {
```
pwittrock: Who implements this?

pkg/kstatus/observe/event/event.go
pwittrock: or `ObservedStateChanged`?

pwittrock:
Looks good to me.  Mostly nits that can be addressed in a follow up.

pwittrock:
@mortent are there any outstanding comments from @KnVerey in this part of the PR?  If not, I suggest we move forward.

mortent:
@pwittrock I have addressed some of the comments, but I haven't pushed them yet. I will make an update to this PR so these changes can go in with it.

pkg/kstatus/observe/observer/observer.go
mortent: I have added comments here.

pkg/kstatus/observe/observer/observer.go
mortent: Yeah, it should be package private. I have updated it.

pkg/kstatus/observe/observer/observer.go
mortent: This is the interface to the core framework. I will add a separate interface that is meant for end-users that will provide defaults for many of these fields.

pkg/kstatus/observe/observer/resourceobserver.go
mortent: Good catch. I have updated the name

pkg/kstatus/observe/observer/observer.go
mortent: I have added comments.

pkg/kstatus/observe/observer/observer.go
mortent: `observeUntilCancelled` is a good name. I have updated the PR.

pkg/kstatus/observe/event/event_test.go
mortent: Added.

PR #35, opened 2020-02-11
Aggregator implementation for the Observe framework

mortent:
This is the basic implementation of the aggregator interface. We still need to figure out the best way to handle status when we delete/prune resources.

@monopole @pwittrock

### DIFF
--- a/pkg/kstatus/observe/aggregator/aggregator.go
+++ b/pkg/kstatus/observe/aggregator/aggregator.go
@@ -1,0 +1,74 @@
+package aggregator
+
+import (
+	"sigs.k8s.io/cli-utils/pkg/kstatus/observe/event"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/wait"
+)
+
+// BasicAggregator implements StatusAggregator.
+// Aggregate status will be Current when all observed
+// resources are either Current or NotFound.
+// TODO: Treating resources that doesn't exist as Current is
+// weird. But it kinda does make sense when we track
+// resources that are deleted/pruned. We should see if
+// there is a better way to handle this.
+type BasicAggregator struct {
+	resourceCurrentStatus map[wait.ResourceIdentifier]status.Status
+}
+
+// NewBasicAggregator returns a BasicAggregator that will track
+// resources identified by the argument.
+func NewAllCurrentOrNotFoundStatusAggregator(identifiers []wait.ResourceIdentifier) *BasicAggregator {
+	aggregator := &BasicAggregator{
+		resourceCurrentStatus: make(map[wait.ResourceIdentifier]status.Status),
+	}
+	for _, id := range identifiers {
+		aggregator.resourceCurrentStatus[id] = status.UnknownStatus
+	}
+	return aggregator
+}
+
+// ResourceObserved is called whenever we have an observation of a resource. In this
+// case, we just keep the latest status so we can later compute the aggregate status
+// for all the resources.
+func (d *BasicAggregator) ResourceObserved(r *event.ObservedResource) {
+	d.resourceCurrentStatus[r.Identifier] = r.Status
+}
+
+// AggregateStatus computes the aggregate status for all the resources. In this
+// implementation of the Aggregator, we treat resources with the NotFound status as Current.
+func (d *BasicAggregator) AggregateStatus() status.Status {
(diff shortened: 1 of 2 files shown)


pkg/kstatus/observe/aggregator/aggregator.go
```diff
@@ -0,0 +1,73 @@
...
+	"sigs.k8s.io/cli-utils/pkg/kstatus/wait"
+)
+
+// AllCurrentOrNotFoundStatusAggregator is an implementation of the StatusAggregator interface
+// that considers the aggregate status as Current whenever all resources have either the Current
+// or the NotFound status.
+// TODO: Treating resources that doesn't exist as Current is weird. But it kinda does
+// make sense when we track resources that are deleted/pruned. We should see if there is a better
+// way to handle this.
+type AllCurrentOrNotFoundStatusAggregator struct {
```
monopole: Per the PR description, lets just call this `BasicAggregator` for now.

Please shorten the description, e.g. 

```
BasicAggregator implements StatusAggregator.
Aggregate status will be Current when all observed
resources are either Current or NotFound.
```

pkg/kstatus/observe/aggregator/aggregator.go
```diff
@@ -0,0 +1,73 @@
...
+	resourceCurrentStatus map[wait.ResourceIdentifier]status.Status
+}
+
+// NewAllCurrentOrNotFoundStatusAggregator returns a new StatusAggregator that will track
+// the status of all resources given by the provided identifiers.
```
monopole: shorten:
```
NewBasicAggregator returns a BasicAggregator that will track
resources identified by the argument.
```

pkg/kstatus/observe/aggregator/aggregator.go
```diff
@@ -0,0 +1,73 @@
...
+func NewAllCurrentOrNotFoundStatusAggregator(identifiers []wait.ResourceIdentifier) *AllCurrentOrNotFoundStatusAggregator {
+	aggregator := &AllCurrentOrNotFoundStatusAggregator{
+		resourceCurrentStatus: make(map[wait.ResourceIdentifier]status.Status),
+	}
+	for _, id := range identifiers {
```
monopole: empty list not an error?

pkg/kstatus/observe/aggregator/aggregator.go
```diff
@@ -0,0 +1,73 @@
...
+		aggregator.resourceCurrentStatus[id] = status.UnknownStatus
```
monopole: repeated entries not an error?

pkg/kstatus/observe/aggregator/aggregator.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	return aggregator
+}
+
+// ResourceObserved is called whenever we have an observation of a resource. In this
+// case, we just keep the latest status so we can later compute the aggregate status
+// for all the resources.
+func (d *AllCurrentOrNotFoundStatusAggregator) ResourceObserved(observedResource *event.ObservedResource) {
```
monopole: s/observedResource/r/

http://go/gocomments#variable-names

pkg/kstatus/observe/aggregator/aggregator.go
```diff
@@ -0,0 +1,73 @@
...
+// implementation of the Aggregator, we treat resources with the NotFound status as Current.
+func (d *AllCurrentOrNotFoundStatusAggregator) AggregateStatus() status.Status {
+	// if we are not observing any resources, we consider status be Current.
+	if len(d.resourceCurrentStatus) == 0 {
+		return status.CurrentStatus
+	}
+
+	allCurrentOrNotFound := true
+	anyUnknown := false
+	for _, latestStatus := range d.resourceCurrentStatus {
```
monopole: s/latestStatus/s/

pkg/kstatus/observe/aggregator/aggregator.go
mortent: The framework is set up to handle an empty slice of identifiers. Not clear where that is useful, but it will work and simply return that the aggregate status is Current.

pkg/kstatus/observe/aggregator/aggregator.go
mortent: It might be that the framework should check for repeated entries before starting the polling cycles. But it should behave as expected, it just means the repeated resources will be polled more often than necessary.

PR #36, opened 2020-02-11
Implementations of the ObserverReader interface for the Observe frame…

mortent:
Two implementations of the `ObserverReader` interface. One is a simple callthrough to the underlying `client.Reader`, the other one pre-fetches the data using LIST calls against the cluster for reach poll cycle.

@monopole @pwittrock

### DIFF
--- a/pkg/kstatus/observe/reader/caching_reader.go
+++ b/pkg/kstatus/observe/reader/caching_reader.go
@@ -1,0 +1,225 @@
+package reader
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// This map is hard-coded knowledge that a Deployment contains and
+// ReplicaSet, and that a ReplicaSet in turn contains Pods, etc., and the
+// approach to finding status being used here requires hardcoding that
+// knowledge in the status client library.
+// TODO: These should probably be defined in the observers rather than here.
+var genGroupKinds = map[schema.GroupKind][]schema.GroupKind{
+	schema.GroupKind{Group: "apps", Kind: "Deployment"}: { //nolint:gofmt
+		{
+			Group: "apps",
+			Kind:  "ReplicaSet",
+		},
+	},
+	schema.GroupKind{Group: "apps", Kind: "ReplicaSet"}: { //nolint:gofmt
+		{
+			Group: "",
+			Kind:  "Pod",
+		},
+	},
+	schema.GroupKind{Group: "apps", Kind: "StatefulSet"}: { //nolint:gofmt
+		{
+			Group: "",
+			Kind:  "Pod",
+		},
+	},
+}
+
+// NewCachingClusterReader returns a new instance of the ObserverReader. The
+// ClusterReader needs will use the reader to fetch resources from the cluster,
+// while the mapper is used to resolve the version for GroupKinds. The list of
+// identifiers is needed so the ClusterReader can figure out which GroupKind
+// and namespace combinations it needs to cache when the Sync function is called.
+// We only want to fetch the resources that are actually needed.
+func NewCachingClusterReader(reader client.Reader, mapper meta.RESTMapper, identifiers []wait.ResourceIdentifier) (*CachingClusterReader, error) {
+	gvkNamespaceSet := newGnSet()
+	for _, id := range identifiers {
+		// For every identifier, add the GroupVersionKind and namespace combination to the gvkNamespaceSet and
+		// check the genGroupKinds map for any generated resources that also should be included.
+		err := buildGvkNamespaceSet(mapper, []schema.GroupKind{id.GroupKind}, id.Namespace, gvkNamespaceSet)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return &CachingClusterReader{
+		reader: reader,
+		mapper: mapper,
+		gns:    gvkNamespaceSet.gvkNamespaces,
+	}, nil
+}
+
+func buildGvkNamespaceSet(mapper meta.RESTMapper, gks []schema.GroupKind, namespace string, gvkNamespaceSet *gvkNamespaceSet) error {
+	for _, gk := range gks {
+		mapping, err := mapper.RESTMapping(gk)
+		if err != nil {
+			return err
+		}
+		gvkNamespaceSet.add(gvkNamespace{
+			GVK:       mapping.GroupVersionKind,
+			Namespace: namespace,
+		})
+		genGKs, found := genGroupKinds[gk]
+		if found {
+			err := buildGvkNamespaceSet(mapper, genGKs, namespace, gvkNamespaceSet)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+type gvkNamespaceSet struct {
+	gvkNamespaces []gvkNamespace
+	seen          map[gvkNamespace]bool
+}
+
+func newGnSet() *gvkNamespaceSet {
+	return &gvkNamespaceSet{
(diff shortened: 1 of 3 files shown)


pkg/kstatus/observe/reader/cached_reader.go
```diff
@@ -0,0 +1,207 @@
...
+// CachedObserverReader is an implementation of the ObserverReader interface that will
+// pre-fetch all resources needed before every sync loop. The resources needed are decided by
+// finding all combinations of GroupVersionKind and namespace referenced by the provided
+// identifiers. This list is then expanded to include any known generated resource types.
+type CachedObserverReader struct {
+	sync.RWMutex
+
+	reader client.Reader
+
+	mapper meta.RESTMapper
```
monopole: comment all these fields.

pkg/kstatus/observe/reader/cached_reader.go
```diff
@@ -0,0 +1,207 @@
...
+
+	gns []gvkNamespace
+
+	cache map[gvkNamespace]unstructured.UnstructuredList
```
monopole: comment should say how this helps, and under what conditions is it invalidated

pkg/kstatus/observe/reader/cached_reader.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func (g *gvkNamespaceSet) add(gn gvkNamespace) {
+	if _, found := g.seen[gn]; !found {
+		g.gvkNamespaces = append(g.gvkNamespaces, gn)
+		g.seen[gn] = true
+	}
+}
+
+// CachedObserverReader is an implementation of the ObserverReader interface that will
```
monopole: An `ObserverReader` sounds like something that (oddly)  _reads an observer_.

But the interface is more clearly a `ClusterReader` or `SyncableClusterReader` or ...

It's clearly a `client.Reader` adapter.  Naming it after the value added by the adaptation would be reasonable. But the value add at the _interface_ level isn't clear.

I know the interface is already in, but i'm objecting anyway :)

Here we have `CachedObserverReader`, which sounds like a reader that is kept in a cache.

Suggest the name:  `CachingClusterReader`.

pkg/kstatus/observe/reader/cached_reader.go
```diff
@@ -0,0 +1,207 @@
...
+
+// ListClusterScoped lists all resource identifier by the GVK of the list and selector
+// from the cache. If the needed combination of GVK and namespace (which for clusterscoped resources
+// will always be the empty string) is not part of the cache, that is considered an error.
+func (c *CachedObserverReader) ListClusterScoped(ctx context.Context, list *unstructured.UnstructuredList, selector labels.Selector) error {
+	return c.ListNamespaceScoped(ctx, list, "", selector)
+}
+
+// Sync loops over the list of gvkNamespace we know of, and uses list calls to fetch the resources.
+// This information populates the cache.
```
monopole: Why have this in the interface at all - i.e., why not call a private sync at the top of the other methods?   

Or why not have a caching reader maintain it's own polling/caching loop rather than have some observer do the maintenance?

pkg/kstatus/observe/reader/cached_reader.go
```diff
@@ -0,0 +1,207 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// genGroupKinds captures the generated resources for a given GroupKind. This is used
```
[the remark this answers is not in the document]
mortent: Your explanation is good. I have updated the PR to include this. 

The name "generated" resources is a bit unfortunate, because while the most common situation is that an observer for a resource will look at generated resource (like deployment will look at replicasets), it is possible that we want to compute status for a resource by looking other resources that are not generated by that resource. An example from Shopify is that we might compute status for a service by looking if any available pods are targeted by the service. I will update if I come up with a better naming here.

pkg/kstatus/observe/reader/cached_reader.go
mortent: Agree. I have renamed the caching implementation to `CachingClusterReader` and the implementation which just wraps `client.Reader` to `DirectClusterReader`.
I will do a follow up PR to rename the interface.

pkg/kstatus/observe/reader/cached_reader.go
mortent: So since the whole framework is based around polling with a certain frequency, having the framework call `Sync` before every polling cycle means we get state from the cluster that is as up-to-date as possible, and we only need to do one LIST call per GVK and namespace combination per cycle. 

If we do sync as part of each observer, we either end up with more sync calls than necessary, or the observers will need to be aware of the polling cycle. I was trying to avoid this, partly because it allows the observers to be reused in an implementation based on a watch instead of polling.

If the reader has its own polling loop, it will not be aligned with the framework cycle. It seems to me that would make things more complicated.

monopole:
OK, thanks for the chat.  Let's go forward and try this.
/lgtm

i'm going to merge and ignore the linter out of memory on travis error.

pkg/kstatus/observe/reader/caching_reader.go
```diff
@@ -0,0 +1,225 @@
...
+	return c.ListNamespaceScoped(ctx, list, "", selector)
+}
+
+// Sync loops over the list of gvkNamespace we know of, and uses list calls to fetch the resources.
+// This information populates the cache.
+func (c *CachingClusterReader) Sync(ctx context.Context) error {
+	c.Lock()
+	defer c.Unlock()
+	cache := make(map[gvkNamespace]unstructured.UnstructuredList)
+	for _, gn := range c.gns {
```
KnVerey: FWIW, our implementation parallelizes these to speed up the sync (and locks the cache at the ns-gvk level instead of globally). We also retry server errors and client timeouts a few times (with backoff) on the list requests, since empty caches are problematic for the polling cycle when using this strategy (though perhaps less critically so in your implementation, I'm not sure). In an older implementation, we let the equivalent of the cluster reader fall back on making individual requests when the required cache was empty, but I think that made less sense than just retrying the request we actually wanted to make.

pkg/kstatus/observe/reader/caching_reader.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return err
+	}
+	gn := gvkNamespace{
+		GVK:       gvk,
+		Namespace: key.Namespace,
+	}
+	cacheList, found := c.cache[gn]
+	if !found {
+		return fmt.Errorf("GVK %s and Namespace %s not found in cache", gvk.String(), gn.Namespace)
```
KnVerey: How does this surface at the watcher level? i.e. does it just report errors on the affected resource(s)--errors that will likely go away in the next polling cycle--or is there some way this could end up being fatal to the watch?

mortent: Currently any errors within a single observer are not considered fatal, so the error will be reported as part of an event for the particular resource. The observer polling will keep running.
Any errors encountered in the `Sync` function is considered fatal. This will cause the main observer loop to return an error event and then exit.
I'm not sure if this is the best way. Definitely that we could look more into.

pkg/kstatus/observe/reader/caching_reader.go
mortent: I have considered running this in multiple goroutines, but I consider that as a more advanced feature. I'm not sure if a more fine-grained locking scheme here would make a significant difference, in particular since it is already using Read-Write locks, so it should already allow multiple readers.
Retry is a good point. I think the client already does some selective retries based on the error code returned from the server. I'll have to check with the kubebuilder team on this. I don't want multiple layers of retry.

KnVerey: >  I consider that as a more advanced feature

Do you mean this is something you'd add later in the development of the library if practical feedback warrants it, or would this be something library users would have to implement their own reader for?

mortent: Certainly something we could implement as part of the library. Just doesn't seem like the highest priority right now. Shouldn't be too hard to implement a workqueue to do this based on a go channel and goroutines.

This is something that would actually be hard for users to change since it will require changes in the core library loop.

PR #37, opened 2020-02-11
ResourceObservers for the Observe framework

mortent:
Implementations of the ResourceObserver interface for some of the common Kubernetes resources. 

There is still some boilerplate code needed, but I haven't found a better way to handle it yet while still making them somewhat readable. I plan to keep iterating on this as we add support for more resource types.

@pwittrock @monopole

### DIFF
--- a/pkg/kstatus/observe/observers/common.go
+++ b/pkg/kstatus/observe/observers/common.go
@@ -1,0 +1,171 @@
+package observers
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"sort"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/observe/event"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/observe/observer"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/wait"
+)
+
+// BaseObserver provides some basic functionality needed by the observers.
+type BaseObserver struct {
+	Reader observer.ClusterReader
+
+	Mapper meta.RESTMapper
+
+	computeStatusFunc observer.ComputeStatusFunc
+}
+
+// SetComputeStatusFunc allows for setting the function used by the observer for computing status. The default
+// value here is to use the status package. This is provided for testing purposes.
+func (b *BaseObserver) SetComputeStatusFunc(statusFunc observer.ComputeStatusFunc) {
+	b.computeStatusFunc = statusFunc
+}
+
+// LookupResource looks up a resource with the given identifier. It will use the rest mapper to resolve
+// the version of the GroupKind given in the identifier.
+// If the resource is found, it will be returned and the observedResource will be nil. If there is an error
+// or the resource is not found, the observedResource will be returned containing information about the resource
+// and the problem.
+func (b *BaseObserver) LookupResource(ctx context.Context, identifier wait.ResourceIdentifier) (*unstructured.Unstructured, error) {
+	GVK, err := b.GVK(identifier.GroupKind)
+	if err != nil {
+		return nil, err
+	}
+
+	var u unstructured.Unstructured
+	u.SetGroupVersionKind(GVK)
+	key := keyForNamespacedResource(identifier)
+	err = b.Reader.Get(ctx, key, &u)
+	if err != nil {
+		return nil, err
+	}
+	u.SetNamespace(identifier.Namespace)
+	return &u, nil
+}
+
+// ObservedGeneratedResources provides a way to fetch the statuses for all resources of a given GroupKind
+// that match the selector in the provided resource. Typically, this is used to fetch the status of generated
+// resources.
+func (b *BaseObserver) ObserveGeneratedResources(ctx context.Context, observer observer.ResourceObserver, object *unstructured.Unstructured,
+	gk schema.GroupKind, selectorPath ...string) (event.ObservedResources, error) {
+	namespace := getNamespaceForNamespacedResource(object)
+	selector, err := toSelector(object, selectorPath...)
+	if err != nil {
+		return event.ObservedResources{}, err
+	}
+
+	var objectList unstructured.UnstructuredList
+	gvk, err := b.GVK(gk)
+	if err != nil {
+		return event.ObservedResources{}, err
+	}
+	objectList.SetGroupVersionKind(gvk)
+	err = b.Reader.ListNamespaceScoped(ctx, &objectList, namespace, selector)
+	if err != nil {
+		return event.ObservedResources{}, err
+	}
+
+	var observedObjects event.ObservedResources
+	for i := range objectList.Items {
+		generatedObject := objectList.Items[i]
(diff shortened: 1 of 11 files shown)


pkg/kstatus/observe/observers/common.go
```diff
@@ -0,0 +1,171 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/observe/event"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/observe/observer"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/wait"
+)
+
+// BaseObserver provides some basic functionality needed by the observers.
+type BaseObserver struct {
+	Reader observer.ObserverReader
```
monopole: ClusterReader

rebase?

monopole:
ack, thanks, looking

monopole:
Certain words - reader, writer, parser evoke expectations.
“Observer” evokes the observer pattern - i.e. multiple
observers register with some state holder, the holder
notifies observers of state changes, the observers then
read the state and do something.  Thus we decouple
state from UI components.

I dug around in code that’s already committed expecting
some signatures like
```
func (x foo) AddObserver(o Observer)
```
but didn’t see them.

Likewise not seeing anything here that looks like
```
func (o Observer) HandleStateChange(e Event)
```
I do see the codebase already has 

```
type ResourceObserver interface {
  Observe(resourceId) *ObservedResource
  ObserveObect(unstruct) *ObservedResource
}
```

but this interface behaves like a reader;
pass in an ID, get an object back from a cluster.

An ObservedResource is a struct of unstructured,
Status and ObservedResource subcomponents
(i.e. a ReplicaSet has Pods).

It makes sense as a “read”.

There’s no registration here.  The ObservedResource
returned isn’t actually being “observed” 
in the sense that if it’s state changes, nobody
is notified, because ObservedResource has no storage
that can hold an observer.






This PR has

BaseObserver {
  // a cluster Reader paired with an injected
  // status computation strategy - AWESOME SAUCE!

  LookupResource // decorates reader.Get
  ObserveGeneratedResources // decorates reader.ListNamespaceScoped 
}

Its tests do a fine job of testing the reading
decoration code, but don’t test anything related
to observer pattern.

The PR has

DefaultObserver struct {
  BaseObserver
}

The first part of this struct name makes sense to me
after reading the whole PR, as it’s a BaseObserver
with a _default_ status computation strategy.

This struct is an implementation of ResourceObserver,
but again that’s basically a reader interface,
not an observer.
The tests of DefaultObserver are, again, tests of
a decorated reader.

After this, we get the extensions of BaseObserver

  DeploymentObserver
  JobObserver
  PodObserver
  ReplicaSetObserver
  ServiceObserver
  StatefulSetObserver

I’ve complained that these aren’t observers.  Another surprise is that they are all created with the same `computeStatusFunc`. One expects that function to encapsulate the difference between Pod and Job, etc.  Perhaps that’s why there’s a setter for computeStatusFunc, but the comment says that’s for testing.  

Do they all have the same `Observe` method (code duplication) ?  seems so?

They differ in their implementation of `ObserveObject`, which is really just a method that converts unstructured to an `ObservedResource`.

So sorry to nit pick but this package doesn’t have observers, so it shouldn’t be called that unless i’m reading it wrong. always possible.

The meat of this PR is N (Pod, Deployment, Job, etc) pairings of 
- ClusterReader
- ComputeStatusFunc (all of these are the same at the moment)
- a YAML manipulation function written as a public method that converts an unstructured into an ObserverResource (differentiating Job, Pod etc.)

There are N tests too, which is fine - but these structs are too similar to need different tests that incorporate a reader.

The tests differ only in that YAML manipulation stuff - the content of ObserveObject.

Perhaps  we need a factory here that creates ReconciledResourceReaders (one class, many instances).

These are all instances of the same class, but adopt your _fine idea_ that they have a injected function for computing status, and _also_ an injected function for converting unstructured to ReconciledResource (new name for ObservedResource).

Roughly: 

>  ObservedResource -> ReconciledResource
>  BaseObserver -> ReconciledResourceReader

the test for RRReader is basically your test for BaseObserver.

Then - on the side you have N simple tests for N functions that do yaml conversions (Job, Pod, etc.) --- no need for a fake reader.

And presumably N tests for status computation functions which presumably live in the status package.

mortent:
If I understand your comment correctly, there are two main points.

First, it uses the name `Observer`, but doesn't actually implement the Observer pattern. That is correct as we discussed few days ago. I can see how that can be confusing, so if we can come up with a better name (probably avoiding another interface with Reader in the name?), we can definitely change it. I used the name `Observer` because they are somewhat like a controller that only looks at the state of resources, but never updates them. So instead of a reconciler which is used in kubebuilder, it is only an observer.

Second, the implementation of the observers doesn't seem all that useful and contains some duplicated code between the implementations for different resource types. That is correct right now, as the observers currently does not leverage the capability of computing status not only based on a single resource, but also based on generated or otherwise associated resources. The goal is to expand on the implementation of the observers, so for example a service can look up pods to make sure it is targeting at least one available pod. For resources that doesn't need to look up any other resources, they can be handled by the generic observer that doesn't do anything but delegate to the status library (The pod and job observers should probably be removed from this PR as they doesn't have any value right now).
As we do this, I think the test setup (which is also why the `ComputeStatusFunc` is in the interface) will change. I did it this way to have some test coverage but without too much duplicated code in tests, but it didn't turn out as well as I hoped.

I'm not convinced about the suggested names. Using the word `Reconciled` in the name makes it sound like this contains info about resources that are fully reconciled, which will not be the case in most situations. Maybe if we rename the Observer interface to `ResourceReader`?

pkg/kstatus/observe/observers/default.go
```diff
@@ -0,0 +1,60 @@
...
+// doesn't have a specific observer. It will just delegate computation of
+// status to the status library.
+// This should work pretty well for resources that doesn't have any
+// generated resources and where status can be computed only based on the
+// resource itself.
+type DefaultObserver struct {
+	BaseObserver
+}
+
+func (d *DefaultObserver) Observe(ctx context.Context, identifier wait.ResourceIdentifier) *event.ObservedResource {
```
pwittrock: (non-blocking) consider `GenericObserver`

pkg/kstatus/observe/observers/default.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/observe/event"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/observe/observer"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/wait"
+)
+
+func NewDefaultObserver(reader observer.ClusterReader, mapper meta.RESTMapper) observer.ResourceObserver {
```
pwittrock: (non-blocking) plz add `var _ InterfaceName = &DefaultObserver`

pwittrock:
file an issue to follow up on naming

fwiw the `Observer` naming is consistent with how we refer to the Kubernetes architecture of Observed vs Desired State.

pkg/kstatus/observe/observers/common.go
```diff
@@ -0,0 +1,171 @@
...
+
+// SetComputeStatusFunc allows for setting the function used by the observer for computing status. The default
+// value here is to use the status package. This is provided for testing purposes.
+func (b *BaseObserver) SetComputeStatusFunc(statusFunc observer.ComputeStatusFunc) {
+	b.computeStatusFunc = statusFunc
+}
+
+// LookupResource looks up a resource with the given identifier. It will use the rest mapper to resolve
+// the version of the GroupKind given in the identifier.
+// If the resource is found, it will be returned and the observedResource will be nil. If there is an error
```
KnVerey: This comment doesn't match the function's return values anymore

mortent: Good catch. I have created a PR to fix it.

PR #38, opened 2020-02-11
Collector to help consumers of the Observer library

mortent:
The Observe library provides a stream of events back to the caller. This is useful in many situations, but sometimes the client needs to be able to access the status and state of all resources at a point in time. The Collector listens to the stream of events from the Observer library, and keeps track of the latest event for each resource. It provides a function for clients to fetch the latest state/status at any point. The Collector is threadsafe.

@monopole @pwittrock

### DIFF
--- a/pkg/kstatus/observe/collector/collector.go
+++ b/pkg/kstatus/observe/collector/collector.go
@@ -1,0 +1,111 @@
+package collector
+
+import (
+	"sort"
+	"sync"
+
+	"sigs.k8s.io/cli-utils/pkg/kstatus/observe/event"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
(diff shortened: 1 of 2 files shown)


pkg/kstatus/observe/collector/collector.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	return &ObservedStatusCollector{
+		aggregateStatus: status.UnknownStatus,
+		observations:    observations,
+	}
+}
+
+// ObservedStatusCollector is for use by clients of the observe library and provides
+// a way to keep track of the latest status/state for all the observed resources. The collector
+// is set up to listen to the eventChannel and keep the latest event for each resource. It also
```
monopole: it would be helpful to see a doc.go explaining the big picture of 'observe' package usage,

PR #45, opened 2020-02-13
Unit test for Applier

mortent:
First attempt to write a unit test for the applier. I was hoping we could use this to help guide us when we try to disentangle the different pieces in the applier and make it more testable.


mortent:
This is updated and ready for review. Still many things in here that we want to improve, but at least the main test function is relatively small.

@seans3

PR #54, opened 2020-02-21
Update the flags used to enable and configure wait for status in applier

mortent:
Update the name of the flags for enabling and configuring status during apply. The names are a little longer than I wanted, but both `--wait` and `--timeout` are unavailable because they are already set by existing kubectl code.

@seans3 
@pwittrock


mortent:
Yeah, I had forgotten about that. I have updated the PR to fix the flag used in the e2e tests.

PR #61, opened 2020-02-25
Improve status for pods

mortent:
This improves the status rules for pods. It makes a few changes:
- A pod that has the `Failed` phase (completed unsuccessfully), is considered Current. I think this makes the most sense from the apply perspective.
- This now checks to see if the pod is unschedulable, and if so, returns the `Failed` status. 
- It checks to see if the pod has any containers that is crashlooping. If so, it sets the status to `Failed`.

This change also demonstrates that working with unstructured objects in the status rules are pretty awkward. We should consider either converting to the actual types or use the kyaml library.

@seans3 @pwittrock @monopole @knverey

### DIFF
--- a/pkg/kstatus/status/core.go
+++ b/pkg/kstatus/status/core.go
@@ -6,6 +6,7 @@ package status
 import (
 	"fmt"
 	"math"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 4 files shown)


pkg/kstatus/status/core.go
```diff
@@ -356,43 +357,94 @@ func podConditions(u *unstructured.Unstructured) (*Result, error) {
...
+		containerNames, isCrashLooping, err := getCrashLoopingContainers(obj)
+		if err != nil {
+			return nil, err
 		}
+		if isCrashLooping {
+			return newFailedStatus("ContainerCrashLooping",
+				fmt.Sprintf("Containers in CrashLoop state: %s", strings.Join(containerNames, ","))), nil
+		}
+
+		return newInProgressStatus("PodRunningNotReady", "Pod is running but is not Ready"), nil
```
monopole: have you considered using enums so we can lean on the compiler and IDE to maintain these state names?

mortent: I have thought about it. These roughly corresponds to the `Reason` field in the Conditions, which are of type string and does not have a clearly defined set of possible values. So I have kept them as strings so far.
Longer term I would like to find a way to automatically generate documentation that would enumerate the possible values here with a description of what it actually means.

pkg/kstatus/status/core.go
```diff
@@ -356,43 +357,94 @@ func podConditions(u *unstructured.Unstructured) (*Result, error) {
...
+	case "Succeeded":
 		return &Result{
 			Status:     CurrentStatus,
 			Message:    "Pod has completed successfully",
 			Conditions: []Condition{},
 		}, nil
-	}
+	case "Failed":
+		return &Result{
+			Status:     CurrentStatus,
```
KnVerey: Why isn't this one a FailedStatus?

mortent: The thought here is that since the pod has gotten into the terminal `Failed` phase, it was "applied" successfully, but failed to complete whatever task it was supposed to do. This would analogous to jobs, which I think we have discussed before, where we consider a job Current whenever it starts running and we don't care (for the purpose of computing status for apply) what happens after it starts running.
I realize this might make some assumptions on the lifecycle of pods that might not be true. I'll take a look at this again.

KnVerey: > I realize this might make some assumptions on the lifecycle of pods that might not be true. I'll take a look at this again.

What was your conclusion? It's true the pod applied successfully, but that doesn't say much--aren't we more interested in whether it reached the desired state? It's true that the script the pod runs may just have exited non-zero, but this state can also be caused by configuration problems of various kinds. Regardless of the cause, it's probably safe to say "Failed" did not accomplish the user's intent in creating the pod. 

Job in particular is tricky, because of how long the user intent may take to accomplish. For Krane, we made the decision that waiting for at least one of the job's pods to start running was a good compromise success condition. However, if we see that the job fully failed while we're still watching, we will declare it failed in our results.

PR #62, opened 2020-02-25
Disable travis in favor of prow.

monopole:


seans3:
This PR removes Travis (keeping the old travis config), since we're moving to Prow for our continuous build. Travis was continually crashing by running out of memory.

PR #63, opened 2020-02-25
Create go.yml

pwittrock:


monopole:
prow insufficient?

pwittrock:
Was talking with Sean about this.  GH checks well integrated into the platform and IMHO will be easier for folks to edit.  One nice thing is that the configuration is in the repo, so we don't have to make changes to the prow repo to configure them.

monopole:
prow runs `make prow-presubmit-check`.

our config is already in repo and runnable from the command line.

mortent:
I'm open to trying to GH, but prow is convenient since it supports commands for re-running tests. It also integrates well with tide (for automatic merge) and it allows multiple checks so we could potentially add a new target for e2e tests. The last two might work well for GH too, but I have never tried it.

monopole:
cli-utils now slots into the same dashboard as the rest of k8s.
https://prow.k8s.io/?repo=kubernetes-sigs%2Fcli-utils

also, it's pretty sane to be able run exactly the CI presubmit test from the command line.

PR #67, opened 2020-02-26
Update names for the observe package

mortent:
This change only contains renaming of files, packages and types. Follows the proposal in https://github.com/kubernetes-sigs/cli-utils/pull/56#discussion_r384166639:

- Top level package renamed from `observe` to `polling`
- Main type in that package renamed from `StatusObserver` to `StatusPoller`
- Main function on the type renamed from `Observe` to `Poll`

The package that implements the core functionality (polling loop) of the library is renamed from `Observer` to `PollerEngine` with the main function renamed to `Poll`

What used to be the `ResourceObserver` interface is now the `StatusReader` interface. The different implementations has been renamed accordingly.

The result of polling a resource and computing the status has been renamed from `ObservedResource` to `ResourceStatus`

@monopole @seans3


pkg/kstatus/polling/clusterreader/caching_reader.go
```diff
@@ -114,7 +114,7 @@ func (g *gvkNamespaceSet) add(gn gvkNamespace) {
 type CachingClusterReader struct {
 	sync.RWMutex
 
-	// reader provides functions to read and list resources from the
+	// clusterreader provides functions to read and list resources from the
```
monopole: oops?

PR #72, opened 2020-02-28
Better support for aggregators on deletion

mortent:
Instead of having a single aggregator implementation for both apply and deletion scenarios, this instead splits it up into two separate aggregators. As far as I know, there is no way to apply/delete in a way where we want to both wait for resources to finish reconcile _and_ be deleted with the same command. So this just allows users of the library to specify whether this is a situation where we want to wait for all the provided resources to reconcile or if we are waiting for them all to be deleted. We will use the former option after apply and the latter after we prune.

@monopole @seans3 @knverey

### DIFF
--- a/pkg/kstatus/polling/aggregator/aggregator.go
+++ b/pkg/kstatus/polling/aggregator/aggregator.go
@@ -9,22 +9,34 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/kstatus/wait"
 )
 
-// BasicAggregator implements StatusAggregator.
-// Aggregate status will be Current when all polled
-// resources are either Current or NotFound.
-// TODO: Treating resources that doesn't exist as Current is
-// weird. But it kinda does make sense when we track
-// resources that are deleted/pruned. We should see if
-// there is a better way to handle this.
-type BasicAggregator struct {
-	resourceCurrentStatus map[wait.ResourceIdentifier]status.Status
+// NewAllCurrentAggregator returns a new aggregator that will consider
+// the set to be completed when all resources have reached the Current status.
+func NewAllCurrentAggregator(identifiers []wait.ResourceIdentifier) *genericAggregator {
+	return newGenericAggregator(identifiers, status.CurrentStatus)
 }
 
-// NewBasicAggregator returns a BasicAggregator that will track
-// resources identified by the argument.
-func NewAllCurrentOrNotFoundStatusAggregator(identifiers []wait.ResourceIdentifier) *BasicAggregator {
-	aggregator := &BasicAggregator{
+// NewAllNotFoundAggregator returns a new aggregator that will consider
+// the set to be completed when all resources have reached the NotFound status.
+func NewAllNotFoundAggregator(identifiers []wait.ResourceIdentifier) *genericAggregator {
+	return newGenericAggregator(identifiers, status.NotFoundStatus)
+}
+
+// genericAggregator implements the StatusAggregator interface. It can
+// be customized by providing the desired status that all resources
+// should reach.
+type genericAggregator struct {
+	resourceCurrentStatus map[wait.ResourceIdentifier]status.Status
+	desiredStatus         status.Status
+}
+
+// newGenericAggregator returns a new Aggregator that will track the
+// resources identified by the slice of ResourceIdentifiers. It will consider
+// the change to be complete when all resources have reached the specified
+// desired status.
+func newGenericAggregator(identifiers []wait.ResourceIdentifier, desiredStatus status.Status) *genericAggregator {
+	aggregator := &genericAggregator{
 		resourceCurrentStatus: make(map[wait.ResourceIdentifier]status.Status),
+		desiredStatus:         desiredStatus,
 	}
 	for _, id := range identifiers {
 		aggregator.resourceCurrentStatus[id] = status.UnknownStatus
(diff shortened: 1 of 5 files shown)


KnVerey:
> As far as I know, there is no way to apply/delete in a way where we want to both wait for resources to finish reconcile and be deleted with the same command. So this just allows users of the library to specify whether this is a situation where we want to wait for all the provided resources to reconcile or if we are waiting for them all to be deleted. We will use the former option after apply and the latter after we prune.

I agree that having NotFound universally treated as Current was problematic, and that it makes sense to have to tell the aggregator which statuses to wait for. However, if I were building a CD tool from these packages, what I'd want to do is:
1. create/update the appropriate resources (set A)
2. prune the appropriate resources (set B)
3. wait for convergence of both (sets A and B) based on the appropriate conditions for each set

It's true that technically 1 and 2 were done sequentially, but that doesn't mean the watches should be. I don't want to have pruning delayed by the (potentially very long) wait for set A. So I feel like I'd end up writing another aggregation layer on top of the one provided here. Furthermore, if I wanted to know when all resources reach either FailedStatus _or_ CurrentStatus (krane's exit condition today), this library already keeps track of all the right information, but the main Poll entrypoint wouldn't support it AFAICT. What do you think about making it possible (either at this layer or via some sort of meta aggregator) to provide multiple sets of resources, each with an explicit list of target states? 

Sidenote: I'm about to leave on vacation, so fyi it'll be a couple weeks before I'll respond again. Please keep tagging me as you see fit, and I'll take a look when I'm back.

mortent:
Our current implementation of apply (with prune and status) actually does things in the following order:
 - Apply the resources
 - Wait for all to become Current
 - Prune
The thought is that if something is wrong, so we time out before all applied resources reach the Current status, we could skip the pruning. I think we still need some validation on whether this is the right path.

I do think the ability to set a target status on each resource individually or on a set of resources could be useful.

My longer term goal for the apply functionality is to support something like:
 - Before any resources are applied, wait for all the concerned resources are Current (so the state of the cluster is settled). 
 - Apply the resources. This will in some special cases need to include waiting for some resources to become current before proceeding with applying other resources.
 - Provide an option on whether we should wait to prune until the resources have become Current or just do prune and apply in parallel.
 - Wait for all resources to have become Current or have been fully Deleted (using Foreground propagation policy when pruning).

It would be great in this scenario if the status poller could run throughout this process. It might require both the ability for resources to have different target statuses, and the ability to dynamically add to (and maybe remove from) the set of resources we are computing status for.

pkg/kstatus/polling/aggregator/aggregator.go
```diff
@@ -35,43 +47,41 @@ func NewAllCurrentOrNotFoundStatusAggregator(identifiers []wait.ResourceIdentifi
...
 
-	allCurrentOrNotFound := true
+	allDesired := true
 	anyUnknown := false
-	for _, s := range d.resourceCurrentStatus {
+	for _, s := range g.resourceCurrentStatus {
 		if s == status.FailedStatus {
 			return status.FailedStatus
 		}
 		if s == status.UnknownStatus {
```
monopole: could just return status.UnknownStatus here, right?

pkg/kstatus/polling/aggregator/aggregator.go
```diff
@@ -35,43 +47,41 @@ func NewAllCurrentOrNotFoundStatusAggregator(identifiers []wait.ResourceIdentifi
...
 			anyUnknown = true
 		}
-		if !(s == status.CurrentStatus || s == status.NotFoundStatus) {
-			allCurrentOrNotFound = false
+		if s != g.desiredStatus {
```
monopole: could just return InProgress here, right?

then if we survive the loop, return g.desiredStatus (no need for booleans)

pkg/kstatus/polling/engine/engine.go
```diff
@@ -85,15 +84,35 @@ func (s *PollerEngine) Poll(ctx context.Context, identifiers []wait.ResourceIden
...
+	// PollInterval defines how often the PollerEngine should poll the cluster for the latest
+	// state of the resources.
+	PollInterval time.Duration
+
+	// WaitForDeletion defines which implementation of the aggregator interface
+	// that should be used. If the value is false, then the aggregator will consider
+	// the wait completed when all resources have reached the Current status. If
+	// the value it true, it will stop polling and return when all resources
+	// have reached the NotFound status
+	WaitForDeletion bool
```
monopole: Can we express PollUntilCancelled and WaitForDeletion as distinct items in an enum to eliminate nonsensical options (e.g. both true)?

pkg/kstatus/polling/aggregator/aggregator.go
monopole: So the comment for this method is:
```
// AggregateStatus returns an aggregation of resource status per the following rules.
// - If any of the resources ...
```
just copy paste your fine comment.

mortent: Done

pkg/kstatus/polling/aggregator/aggregator.go
```diff
@@ -9,22 +9,34 @@ import (
...
-// there is a better way to handle this.
-type BasicAggregator struct {
+// NewAllCurrentAggregator returns a new aggregator that will consider
+// the set to be completed when all resources have reached the Current status.
+func NewAllCurrentAggregator(identifiers []wait.ResourceIdentifier) *genericAggregator {
+	return newGenericAggregator(identifiers, status.CurrentStatus)
+}
+
+// NewAllNotFoundAggregator returns a new aggregator that will consider
+// the set to be completed when all resources have reached the Current status.
```
seans3: How is this different than the NewAllCurrentAggregator?

pkg/kstatus/polling/engine/engine.go
```diff
@@ -85,15 +74,57 @@ func (s *PollerEngine) Poll(ctx context.Context, identifiers []wait.ResourceIden
...
+	return nil
+}
+
+// Options contains the different parameters that can be used to adjust the
+// behavior of the PollerEngine.
+type Options struct {
+	// PollUntilCancelled defines whether the engine should stop polling and close the
+	// event channel when the Aggregator implementation considers all resources to have reached
+	// the desired status.
+	PollUntilCancelled bool
```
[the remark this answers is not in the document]
mortent: I have added comments on how timeouts can be specified on the context.

I see your point on enums. I would like to address that in a separate PR.

pkg/kstatus/polling/aggregator/aggregator.go
mortent: That was a cut-and-paste error. I have updated the comment.

PR #73, opened 2020-02-29
Add status command

mortent:
Adds a separate command for status to cli-utils.

This depends on #72.

@seans3 @monopole

### DIFF
--- a/cmd/status/print/print.go
+++ b/cmd/status/print/print.go
@@ -1,0 +1,19 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package print
+
+// Printer defines an interface for outputting information about status of
+// resources. Different implementations allow output formats tailored to
+// different use cases.
+type Printer interface {
+
+	// Print tells the printer to start outputting data. The stop parameter
+	// is a channel that the caller will use to signal to the printer that it
+	// needs to stop and shut down. The channel returned can be used by the
+	// printer implementation to signal that it has outputted all the data it
+	// needs to, and that it has completed shutting down. The latter is important
+	// to make sure the printer has a chance to output all data before the
+	// program terminates.
+	Print(stop <-chan struct{}) <-chan struct{}
+}
--- a/cmd/status/printers/color.go
+++ b/cmd/status/printers/color.go
@@ -1,0 +1,18 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package printers
+
+const (
+	// RESET is the escape sequence for unsetting any previous commands.
+	RESET = 0
+	// ESC is the escape sequence used to send ANSI commands in the terminal.
+	ESC          = 27
+	RED    color = 31
+	GREEN  color = 32
+	YELLOW color = 33
+)
+
+// color is a type that captures the ANSI code for colors on the
+// terminal.
+type color int
--- a/cmd/status/printers/printers.go
+++ b/cmd/status/printers/printers.go
@@ -1,0 +1,25 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package printers
+
+import (
+	"fmt"
+	"io"
+
+	"sigs.k8s.io/cli-utils/cmd/status/print"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/collector"
+)
+
+// CreatePrinter return an implementation of the Printer interface. The
+// actual implementation is based on the printerType requested.
+func CreatePrinter(printerType string,
+	collector *collector.ResourceStatusCollector, w io.Writer) (print.Printer,
+	error) {
+	switch printerType {
+	case "table":
+		return NewTablePrinter(collector, w), nil
+	default:
+		return nil, fmt.Errorf("no printer available for output %q", printerType)
+	}
+}
--- a/cmd/status/printers/table_printer.go
+++ b/cmd/status/printers/table_printer.go
@@ -1,0 +1,408 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package printers
+
+import (
+	"fmt"
+	"io"
+	"strings"
+	"time"
+	"unicode/utf8"
+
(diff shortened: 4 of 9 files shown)


cmd/status/print/print.go
```diff
@@ -0,0 +1,18 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package print
+
+// Printer defines the interface the status command expects
+// for outputting information about status of resources. Different
+// implementations allow output formats tailored to different use cases.
```
monopole: When defining an interface, its odd to call out who's expected to use it.

cmd/status/print/print.go
```diff
@@ -0,0 +1,18 @@
...
+type Printer interface {
+
```
monopole: Shouldn't this have some kind of `Print` signature?

Like `Reader` has a `Read` method, etc.

cmd/status/printers/color.go
```diff
@@ -0,0 +1,18 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package printers
+
+const (
+	// RESET is the escape sequence for unsetting any previous commands.
+	RESET = 0
+	// ESC is the escape sequence used to send ascii commands in the terminal.
```
monopole: a/ascii/ANSI/

cmd/status/printers/printers.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"fmt"
+	"io"
+
+	"sigs.k8s.io/cli-utils/cmd/status/print"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/collector"
+)
+
+func CreatePrinter(printerType string, collector *collector.ResourceStatusCollector, w io.Writer) (print.Printer, error) {
```
monopole: i'll only mention it here, but feel free to keep it at 80 chars.
https://engdoc.corp.google.com/eng/doc/devguide/cpp/styleguide.md?cl=head#Line_Length

Break wherever you like, as long as the Go formater respects it.

cmd/status/printers/table_printer.go
```diff
@@ -0,0 +1,385 @@
...
+var (
+	// columns contains a list of ColumnDefs, one for each of the columns that
+	// is supported by the printer.
+	columns = []ColumnDef{
+		{
+			// Column containing the namespace of the resource.
+			name:   "namespace",
+			header: "NAMESPACE",
+			width:  10,
+			printContent: func(w io.Writer, availableWidth int, resource *event.ResourceStatus) (int, error) {
```
monopole: a/availableWidth/w/

vars with tiny scopes should have tiny names.  We know this is width from the type def.

monopole: ditto below

cmd/status/printers/table_printer.go
```diff
@@ -0,0 +1,385 @@
...
+				if len(s) > availableWidth {
+					s = s[:availableWidth]
+				}
+				color, setColor := colorForTableStatus(resource.Status)
+				var outputStatus string
+				if setColor {
+					outputStatus = sPrintWithColor(color, s)
+				} else {
+					outputStatus = s
+				}
```
monopole:
```
stat := s
if !setColor {
  stat = sPrintWithColor(color, s)
}
```

cmd/status/printers/table_printer.go
```diff
@@ -0,0 +1,385 @@
...
+					case "False":
+						color = RED
+					default:
+						color = YELLOW
+					}
+					remainingWidth := availableWidth - realLength
+					if len(conditionType) > remainingWidth {
+						conditionType = conditionType[:remainingWidth]
+					}
+					_, err := fmt.Fprint(w, sPrintWithColor(color, conditionType))
```
monopole: no if setcolor here?

cmd/status/printers/table_printer.go
```diff
@@ -0,0 +1,385 @@
...
+func (t *tablePrinter) PrintUntil(stop <-chan struct{}) <-chan struct{} {
+	completed := make(chan struct{})
+
+	linesPrinted := t.printTable(t.collector.LatestObservation(), 0)
+
+	go func() {
+		defer close(completed)
+		ticker := time.NewTicker(updateInterval)
+		for {
+			select {
```
monopole: Maybe also add deadline timer here?  Or is that what's sitting behind 'stop'?

cmd/status/print/print.go
mortent: I have updated the comment.

cmd/status/print/print.go
mortent: It had the `PrintUntil` function. The until part seems unnecessary, so I have renamed the function to just `Print`

cmd/status/printers/color.go
mortent: Fixed

cmd/status/printers/printers.go
mortent: I have updated the PR to stay within 80 chars.

cmd/status/printers/table_printer.go
mortent: Renamed this variable to `width`. `w` is already used by the `io.Writer`

cmd/status/printers/table_printer.go
mortent: So I tried something like this earlier, but I felt it ended up being worse. The challenge is that the code needs to keep track of the column width in the output. The ANSI characters are a challenge here since they do count when checking the length of a string in the program, but does not actually matter when we print. But I am interested in finding a cleaner solution here, but I would like to get back to it in a separate PR.

cmd/status/printers/table_printer.go
mortent: In this case we know that we will always print with colors.

cmd/status/printers/table_printer.go
mortent: Yeah, the stop channel is used to break out of the loop here. That allows the caller to stop the printer for whatever reason, not only based on a timeout.

cmd/status/print/print.go
monopole: yay

PR #74, opened 2020-03-01
Remove wait package

mortent:
This PR does four things, each in a separate commit:
 - Remove the dependencies from the `polling` package on the `wait` package. This was only the `ResourceIdentifier` type. It instead uses the `ObjMetadata` type from the `prune` package.
 - Move the `ObjMetadata` type (with functions and tests) out of the `prune` package and into a new `object` package that it is top level under `pkg`.
 - Refactor the `Applier` to use the `polling` package rather than the `wait` package to poll the resources for status.
 - Remove the `wait` package.

I can split this up into multiple PRs if necessary.

@monopole @seans3 @pwittrock @knverey


pkg/apply/applier.go
```diff
@@ -207,7 +209,12 @@ func (a *Applier) Run(ctx context.Context) <-chan event.Event {
 		}
 
 		if a.StatusOptions.wait {
-			statusChannel := a.resolver.WaitForStatusOfObjects(ctx, infosToObjects(infos))
+			statusChannel := a.statusPoller.Poll(ctx, infosToObjMetas(infos), polling.Options{
+				PollUntilCancelled: true,
+				PollInterval:       a.StatusOptions.period,
+				UseCache:           true,
+				DesiredStatus:      status.CurrentStatus,
+			}) //.WaitForStatusOfObjects(ctx, infosToObjects(infos))
```
seans3: Should this comment be removed // .WaitForStatusOfObjects ...

pkg/apply/prune/inventory.go
```diff
@@ -93,7 +95,7 @@ func (is *Inventory) Subtract(other *Inventory) (*Inventory, error) {
 	return result, nil
 }
 
-// Equals returns true if the "other" inventory set is the same
+// EqualsWithNormalize returns true if the "other" inventory set is the same
```
seans3: This comment is different than the name of the function.

pkg/apply/resource_infos.go
```diff
@@ -70,7 +70,7 @@ func getIndexByKind(kind string) int {
 	return m[kind]
 }
 
-// Equals returns true if the GVK's have equal fields.
+// EqualsWithNormalize returns true if the GVK's have equal fields.
```
seans3: Hmmm...doesn't match the comment does not match the function signature.

pkg/kstatus/polling/statusreaders/common.go
```diff
@@ -137,8 +137,8 @@ func toSelector(resource *unstructured.Unstructured, path ...string) (labels.Sel
 	return metav1.LabelSelectorAsSelector(&s)
 }
 
-func toIdentifier(u *unstructured.Unstructured) wait.ResourceIdentifier {
-	return wait.ResourceIdentifier{
+func toIdentifier(u *unstructured.Unstructured) object.ObjMetadata {
```
seans3: Should this function go into the new object package?

pkg/kstatus/polling/statusreaders/common.go
```diff
@@ -161,7 +161,7 @@ func getNamespaceForNamespacedResource(object runtime.Object) string {
 
 // keyForNamespacedResource returns the object key for the given identifier. It makes
 // sure to set the namespace to default if it is not provided.
-func keyForNamespacedResource(identifier wait.ResourceIdentifier) types.NamespacedName {
+func keyForNamespacedResource(identifier object.ObjMetadata) types.NamespacedName {
```
seans3: Should this function go in the new object package?

seans3:
This looks good. I mention a couple nits. Thanks for pulling the ObjMetadata into a common package; very helpful. I think we should work on creating smaller PR's. As you mention this could/should be multiple PR's. I'll wait for a bit before giving the LGTM to give the others a chance for feedback.

pkg/apply/applier.go
mortent: Yeah, I have removed it.

pkg/apply/prune/inventory.go
mortent: This is the goland refactoring command that has been too eager. I have updated it.

pkg/apply/resource_infos.go
mortent: Fixed

pkg/kstatus/polling/statusreaders/common.go
mortent: Yeah, I agree maybe we should move it. If you don't mind, I'll look at that in a separate PR.

pkg/kstatus/polling/statusreaders/common.go
mortent: Will look at this in a separate PR.

PR #77, opened 2020-03-03
Add count of how many resources have been applied/pruned/deleted in the output

mortent:
The output from running the apply and destroy commands could be confusing since they stated that apply/prune was completed even if nothing actually changed. This change addresses this problem by adding additional information such as the number of resources applied and pruned.

Screen Shot 2020-03-02 at 6 35 24 PM

Ref: https://github.com/GoogleContainerTools/kpt/issues/302

@droot @seans3

### DIFF
--- a/pkg/apply/basic_printer.go
+++ b/pkg/apply/basic_printer.go
@@ -23,11 +23,40 @@ type BasicPrinter struct {
 	IOStreams genericclioptions.IOStreams
 }
 
+type applyStats struct {
+	serversideApplied int
+	created           int
+	unchanged         int
+	configured        int
+}
+
+func (a *applyStats) inc(op event.ApplyEventOperation) {
+	switch op {
+	case event.ServersideApplied:
+		a.serversideApplied++
+	case event.Created:
+		a.created++
+	case event.Unchanged:
+		a.unchanged++
+	case event.Configured:
+		a.configured++
+	default:
+		panic(fmt.Errorf("unknown apply operation %s", op.String()))
+	}
+}
+
+func (a *applyStats) sum() int {
+	return a.serversideApplied + a.configured + a.unchanged + a.created
+}
+
 // Print outputs the events from the provided channel in a simple
 // format on StdOut. As we support other printer implementations
 // this should probably be an interface.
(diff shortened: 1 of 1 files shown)


pkg/apply/basic_printer.go
```diff
@@ -23,23 +23,55 @@ type BasicPrinter struct {
...
+	serversideApplied int
+	created           int
+	unchanged         int
+	configured        int
+}
+
+func (a *applyStats) inc(op event.ApplyEventOperation) {
+	switch op {
+	case event.ServersideApplied:
+		a.serversideApplied++
```
droot: I wonder --- how do you get feedback in SSA case ? Does the server response indicates that ?

pkg/apply/basic_printer.go
```diff
@@ -23,23 +23,55 @@ type BasicPrinter struct {
...
 		switch e.Type {
 		case event.ErrorType:
 			cmdutil.CheckErr(e.ErrorEvent.Err)
 		case event.ApplyType:
 			ae := e.ApplyEvent
 			if ae.Type == event.ApplyEventCompleted {
-				fmt.Fprintf(b.IOStreams.Out, "all resources have been applied\n")
+				fmt.Fprintf(b.IOStreams.Out,
+					"%d resource(s) applied. %d created, %d unchanged, %d configured, %d serverside applied\n",
+					applyStats.sum(), applyStats.created, applyStats.unchanged, applyStats.configured, applyStats.serversideApplied)
```
droot: I think end-users may not know about SSA at all so this might leave them wondering.

pkg/apply/basic_printer.go
mortent: The client knows whether it did a serverside or clientside apply. We use the code from upstream kubectl to handle this. AFAIK, we don't know whether a resource was create, configured or unchanged in the serverside apply case.

pkg/apply/basic_printer.go
mortent: Good point. I have updated the PR so we only print information about ssa if at least one of the resources were applied with ssa.

PR #78, opened 2020-03-03
Use the new standard conditions

mortent:
Check for the `Reconciling` and `Stalled` conditions in the generic status check function. This will allow users to adopt these conditions in their CRDs and the status library will be able to handle them correctly.

@ekitson @monopole @seans3


seans3:
This LGTM. I'll let Erik comment before merging.

PR #81, opened 2020-03-04
apply, preview and destroy take exactly one dir arg

monopole:
Fix GoogleContainerTools/kpt#316


pkg/apply/path.go
```diff
@@ -21,3 +24,26 @@ func processPaths(paths []string) genericclioptions.FileNameFlags {
 	fileNameFlags.Recursive = &t
 	return fileNameFlags
 }
+
+func demandOneDirectory(paths []string) (genericclioptions.FileNameFlags, error) {
+	result := processPaths(paths)
```
seans3: Can we just check the number of paths in the slice before calling processPaths, ensuring there is only one?

seans3: Actually 0 or 1

monopole: had it that way initially (an early exit before call to processPaths), but then figured 

- for lowest impact on the calling code, thought this should return a  genericclioptions.FileNameFlags struct (not a ptr) just like processPaths (a method i didn't want to change)

- and it should literally add validation to the existing behavior - so the struct returned is "real" containing whatever processPaths ends up doing, even in the error case, rather than just an empty struct created soley to accompany the returned error to make the signature correct. 

regardless, we're assure just one path.

monopole: BTW, apply will no longer work with stdin (because `-` isn't a directory) till we again open this up.

PR #82, opened 2020-03-05
First version of InitOptions and tests

seans3:
* First version of `InitOptions` and methods in new `pkg/config`. Happy for another name besides `config`; `init` collides with `init` function.
* Added unit tests. File system calls are not unit tested.

/sig cli
/priority important-soon

```release-note
NONE
```


pkg/config/initoptions.go
```diff
@@ -0,0 +1,179 @@
...
+		return false
+	}
+	return !f.IsDir()
+}
+
+func (i *InitOptions) fillInValues() string {
+	now := time.Now()
+	nowStr := now.Format("2006-01-02 15:04:05 MST")
+	manifestStr := configMapTemplate
+	manifestStr = strings.ReplaceAll(manifestStr, "<DATETIME>", nowStr)
```
mortent: go has a templating engine in "text/template". We should use it.

pkg/config/initoptions.go
```diff
@@ -0,0 +1,179 @@
...
+	namespace := "default"
+	for _, node := range nodes {
+		rm, err := node.GetMeta()
+		if err != nil {
+			continue
+		}
+		if len(rm.ObjectMeta.Namespace) > 0 {
+			namespace = rm.ObjectMeta.Namespace
+		}
+		break
```
mortent: So we just take the namespace from the first resource we find? Isn't that a bit fragile?

pkg/config/initoptions.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func NewInitOptions(ioStreams genericclioptions.IOStreams) *InitOptions {
+	return &InitOptions{
+		ioStreams: ioStreams,
+	}
+}
+
+// Complete fills in the InitOptions fields.
+func (i *InitOptions) Complete(args []string) error {
```
mortent: I'm not sure how we will wire this up. But I don't think we should continue this pattern from Kubectl with the options struct and the Complete function. I think the way kpt commands work are cleaner.

pkg/config/initoptions.go
```diff
@@ -0,0 +1,179 @@
...
+			namespace = rm.ObjectMeta.Namespace
+		}
+		break
+	}
+	return namespace, nil
+}
+
+// defaultGroupName returns a string of the package name (directory)
+// with a random number suffix.
+func (i *InitOptions) defaultGroupName() string {
```
mortent: Do we need the random suffix? I think it would be sufficient to just use the name of the package. If that isn't unique or not valid, we can just let the user decide on a name.

pkg/config/initoptions.go
seans3: I looked into it, and it seemed like overkill. Let me add a TODO.

pkg/config/initoptions.go
seans3: I'll add a TODO.

pkg/config/initoptions.go
seans3: Phil asked for it. We name the groupname after the package directory name, And he thought that if the package is under a directory called "prod", then there might be several of them, and we might need the suffix to disambiguate. I'm happy to remove. Adding a TODO.

pkg/config/initoptions.go
seans3: I don't think so. Either we allow multiple namespaces or we don't. If we don't then it will be fine, and the namespace validation will catch any problems. If we do, then it also works, since we just want the grouping object in one of the namespaces.

pkg/config/initoptions.go
mortent: Yeah, I guess it doesn't really matter if the name has any meaning or not as long as it is unique and it is checked in with the rest of the manifests.

pkg/config/initoptions.go
mortent: But do we ever check whether there are multiple namespaces in the package?
Just remembered this now, but I would assume that the namespace is a field that will usually be defined with a setter. Fetching it from the manifests could lead to surprising results if someone uses the setter after running init.

seans3: Good point. It can also be set with Kustomize build.

PR #90, opened 2020-03-08
Simplify the StatusReaders and add better handling for resource types that act as controllers for pods

mortent:
This changes the structure of the StatusReaders. Instead of each resource specific reader embedding the `BaseStatusReader`, it now acts as a wrapper around each of them. This means that the `ReadStatus` function can be defined only in the`BaseStatusReader` and the resource specific implementations only need to implement the `ReadStatusForObject` function. This also simplifies testing since we can test the `BaseStatusReader` and the resource specific logic independently.

It turns out that the logic for resource types that have pods as generated objects turned out very similar. So this also pulls that out into a separate type. This means some of the resource specific status readers ends up with very little code. It might be worth looking into whether the framework should support that each status reader can handle multiple GKs. I'll revisit this in a separate PR.

@monopole @seans3 @pwittrock @knverey


pkg/kstatus/polling/statusreaders/common.go
```diff
@@ -23,60 +23,96 @@ import (
...
+	reader engine.ClusterReader
+
+	// mapper provides a way to look up the resource types that are available
+	// in the cluster.
+	mapper meta.RESTMapper
+
+	// resourceStatusReader is an resource-type specific implementation
+	// of the resourceTypeStatusReader interface. While the baseStatusReader
+	// contains the logic shared between all resource types, this implementation
+	// will contain the resource specific info.
```
monopole: thanks, these field comments make sense.

pkg/kstatus/polling/statusreaders/common.go
```diff
@@ -23,60 +23,96 @@ import (
...
 		return nil, err
 	}
 	u.SetNamespace(identifier.Namespace)
 	return &u, nil
 }
 
-// StatusForGeneratedResources provides a way to fetch the statuses for all resources of a given GroupKind
+// statusForGenResourcesFunc defines the function type used by the statusForGeneratedResource function.
+// TODO: Find a better solution for this. Maybe put the logic for looking up generated resources
+// into a separate type.
```
monopole: TODO: think of a better adjective than "generated".

PR #91, opened 2020-03-08
New structure for the apply cmd

mortent:
This changes the structure of the apply command to the pattern used in kpt. I like this pattern as it gives us a struct onto which we can bind the cobra flags. I ran into this limitation with the current structure when I wanted to add a `--output` flag to let the user select the output format. 

If we think this is a good structure, I can update the other commands as well.

@monopole @seans3 @pwittrock


seans3:
This seems fine to me. It doesn't seem like a tremendous benefit, but I'll go with it. I'll LGTM when others have had a look.

PR #92, opened 2020-03-10
Accomodates cluster-scoped resources when validating namespace

seans3:
* Fixes `validateNamespace()` function, by taking into account cluster-scoped resources when validating that resources are being applied in the same namespace.
* Added unit tests
* Also manually tested by running `kapply preview <TESTDIR>` as well as `kapply apply <TESTDIR>`

/sig cli
/priority important-soon

```release-note
NONE
```

### DIFF
--- a/pkg/apply/applier_test.go
+++ b/pkg/apply/applier_test.go
@@ -263,32 +263,48 @@ var namespace = "test-namespace"
 
 var obj1Info = &resource.Info{
 	Namespace: namespace,
-	Name:      "foo",
+	Name:      "obj1",
+	Mapping: &meta.RESTMapping{
+		Scope: meta.RESTScopeNamespace,
+	},
 }
 
 var obj2Info = &resource.Info{
 	Namespace: namespace,
-	Name:      "bar",
+	Name:      "obj2",
+	Mapping: &meta.RESTMapping{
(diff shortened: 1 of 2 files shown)


pkg/apply/applier_test.go
```diff
@@ -308,28 +324,32 @@ func TestValidateNamespace(t *testing.T) {
 			objects: []*resource.Info{obj1Info, obj2Info},
 			isValid: true,
 		},
-		"Three resources with same namespace is valid": {
-			objects: []*resource.Info{obj1Info, obj2Info, obj3Info},
+		"Two resources with same namespace and cluster-scoped obj is valid": {
```
KnVerey: Out of curiosity, why is namespace-scoped + cluster-scoped supported but not multiple namespaces? Given the extent of the blast-radius difference, in Krane we decided to not only ban these types from being created together, but even forced a different command to be used. In our context, cluster-scoped resources are generally created by a different type of user than namespace-scoped resources, but even in the case where a given user needs to create both, this separation gives a stronger guarantee that they're aware of what they're doing and not unintentionally polluting the global scope with a resource they believed to be namespaced (for example). It definitely reduces convenience for some so I understand it might not be appropriate here, but curious about your reasoning.

PR #93, opened 2020-03-10
Move hard-coded default namespace to proper constants

seans3:
* Replaces hard-coded "default" namespace with const `metav1.NamespaceDefault`. Does the same with missing namespace and `metav1.NamespaceNone`. This PR does not address tests which hard-code default namespace.
* Ran unit tests

/sig cli
/priority important-soon

```release-note
NONE
```


monopole:
/lgtm

hey bot!

PR #95, opened 2020-03-13
Fix issue that causes wait-for-reconcile to never finish

mortent:
The `Poll` function is called with an incorrect parameter. This causes the `apply` command to never finish if using the `--wait-for-reconcile` flag.

@seans3


seans3:
/lgtm

I'm guessing this was manually tested?

PR #97, opened 2020-03-13
Fixes grouping object namespace calculation

seans3:
* Fixes grouping object namespace calculation that did not properly handle cluster-scoped objects when they were handled first. Only returns default namespace if there are no resources with a namespace.
* Adds better documentation.
* Increases the random suffix from six digits to nine digits for collision safety.
* Tested manually.

/sig cli
/priority important-soon

```release-note
NONE
```

### DIFF
--- a/pkg/config/initoptions.go
+++ b/pkg/config/initoptions.go
@@ -60,11 +60,11 @@ metadata:
 // grouping object template ConfigMap.
 type InitOptions struct {
 	ioStreams genericclioptions.IOStreams
-	// Package directory argument
+	// Package directory argument; must be valid directory.
 	Dir string
-	// Namespace for grouping object
+	// Namespace for grouping object; can not be empty.
 	Namespace string
-	// Grouping object label value
+	// Grouping object label value; must be a valid k8s label value.
 	GroupName string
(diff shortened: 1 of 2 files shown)


pkg/config/initoptions.go
```diff
@@ -115,34 +116,37 @@ func isDirectory(path string) bool {
...
 		}
 		if len(rm.ObjectMeta.Namespace) > 0 {
-			namespace = rm.ObjectMeta.Namespace
+			return rm.ObjectMeta.Namespace, nil
 		}
-		break
 	}
-	return namespace, nil
+	// Return the default namespace if none found.
+	return metav1.NamespaceDefault, nil
```
mortent: Maybe we should just ask the user to tell us which namespace in this case rather than just taking the default namespace? Or we could ask the user to confirm the choices as part of the `init` command?

seans3: I think we should definitely put more thought into a solution. But this fixes a bug. Implementing a system for user input, and then validating (currently we only allow one namespace) would be a bigger effort. Let's get this fix in, and implement the bigger solution in another PR.

PR #98, opened 2020-03-14
Update sorting of resources in Applier

mortent:
This moves handling of the grouping object template and ordering of resources into the same flow. This fixes the issue where we first sort the resources and then handle the grouping object, which can lead to incorrect ordering.

@seans3

### DIFF
--- a/pkg/apply/applier.go
+++ b/pkg/apply/applier.go
@@ -76,7 +76,6 @@ func (a *Applier) Initialize(cmd *cobra.Command, paths []string) error {
 	if err != nil {
 		return errors.WrapPrefix(err, "error setting up ApplyOptions", 1)
 	}
-	a.ApplyOptions.PreProcessorFn = prune.PrependGroupingObject(a.ApplyOptions)
 	a.ApplyOptions.PostProcessorFn = nil // Turn off the default kubectl pruning
 	err = a.PruneOptions.Initialize(a.factory)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/apply/applier.go
```diff
@@ -139,6 +138,56 @@ func (a *Applier) newStatusPoller() (poller, error) {
...
+		return nil, fmt.Errorf("objects have differing namespaces")
+	}
+
+	return append([]*resource.Info{groupingObject}, resources...), nil
+}
+
+// splitInfos takes a slice of resource.Info objects and splits it
+// into one slice that contains the grouping object templates and
+// another one that contains the remaining resources.
+func splitInfos(infos []*resource.Info) ([]*resource.Info, []*resource.Info) {
```
seans3: This is nice and clean

pkg/apply/applier.go
```diff
@@ -139,6 +138,56 @@ func (a *Applier) newStatusPoller() (poller, error) {
 	return polling.NewStatusPoller(c, mapper), nil
 }
 
+// readAndPrepareObjects reads the resources that should be applied,
+// handles ordering of resources and sets up the grouping object
+// based on the provided grouping object template.
+func (a *Applier) readAndPrepareObjects() ([]*resource.Info, error) {
```
seans3: Nice

pkg/apply/applier_test.go
```diff
@@ -364,6 +441,80 @@ func TestValidateNamespace(t *testing.T) {
...
+	}{
+		"no grouping object": {
+			resources:     []*resource.Info{obj1Info},
+			expectedError: true,
+		},
+		"multiple grouping objects": {
+			resources:     []*resource.Info{groupingObjInfo, groupingObjInfo},
+			expectedError: true,
+		},
+		"only grouping object": {
```
seans3: Just to be consistent, should we put the `expectedError: false` here (instead of being implied)?

pkg/apply/applier_test.go
```diff
@@ -364,6 +441,80 @@ func TestValidateNamespace(t *testing.T) {
...
+			resources: []*resource.Info{groupingObjInfo},
+		},
+		"grouping object already at the beginning": {
```
seans3: Same

pkg/apply/applier_test.go
```diff
@@ -364,6 +441,80 @@ func TestValidateNamespace(t *testing.T) {
...
+			resources: []*resource.Info{groupingObjInfo, obj1Info,
+				clusterScopedObjInfo},
+		},
+		"grouping object not at the beginning": {
```
seans3: Same

pkg/apply/applier_test.go
```diff
@@ -364,6 +441,80 @@ func TestValidateNamespace(t *testing.T) {
 	}
 }
 
+func TestReadAndPrepareObjects(t *testing.T) {
```
seans3: Nice tests.

pkg/apply/prune/grouping.go
```diff
@@ -76,120 +75,128 @@ func FindGroupingObject(infos []*resource.Info) (*resource.Info, bool) {
...
-// the grouping object; nil otherwise. Each object is in
-// unstructured.Unstructured format.
-func AddInventoryToGroupingObj(infos []*resource.Info) error {
+// AddInventoryToGroupingObj takes a slice of infos, finds the
+// grouping object template among the resources, and returns a
+// new slice of infos that has a fully populated grouping object
+// as the first info object in the slice. It returns an error
+// if no grouping object can be found, or if the resources
+// do not have the expected unstructured format.
+func AddInventoryToGroupingObj(infos []*resource.Info) ([]*resource.Info,
```
seans3: I don't see the reason for re-writing this function `AddInventoryToGroupingObj`. If it's a refactor to clean-up, and not a bug fix, then let's move this to another PR.

pkg/apply/applier_test.go
mortent: Agree

pkg/apply/prune/grouping.go
mortent: Yeah, I've cleaned this up.

PR #101, opened 2020-03-16
Fixes minor issue in init command and current directory (.)

seans3:
* Correctly handles `kapply init` command using current directory `.` as the directory parameter by expanding the current directory.
* Tested manually.

/sig cli
/priority important-soon

```release-note
NONE
```


mortent:
Looks good, but is there a way we can add a test?

seans3:
> Looks good, but is there a way we can add a test?

We'd have to mock out the `os.Getwd()` call, which I think would mean mocking out the whole filesystem. I think they mocked out the filesystem in `Kustomize`, but I don't think it's a trivial effort. Maybe that should be another PR.

PR #102, opened 2020-03-16
Implement the applier as a queue of tasks

mortent:
This approach should be pretty flexible to support different workflows where we might need to run multiple wait operations in between the apply operations. But this PR is focused on just reproducing the existing functionality.

@seans3 @monopole @knverey

### DIFF
--- a/pkg/apply/applier.go
+++ b/pkg/apply/applier.go
@@ -18,10 +18,12 @@ import (
 	"k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/scheme"
 	"sigs.k8s.io/cli-utils/pkg/apply/event"
+	"sigs.k8s.io/cli-utils/pkg/apply/poller"
 	"sigs.k8s.io/cli-utils/pkg/apply/prune"
+	"sigs.k8s.io/cli-utils/pkg/apply/task"
+	"sigs.k8s.io/cli-utils/pkg/apply/taskrunner"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/polling"
 	pollevent "sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
-	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/cli-utils/pkg/object"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -42,14 +44,16 @@ func NewApplier(factory util.Factory, ioStreams genericclioptions.IOStreams) *Ap
 	}
 }
 
-// poller defines the interface the applier needs to poll for status of resources.
-type poller interface {
-	Poll(ctx context.Context, identifiers []object.ObjMetadata, options polling.Options) <-chan pollevent.Event
-}
-
 // Applier performs the step of applying a set of resources into a cluster,
 // conditionally waits for all of them to be fully reconciled and finally
 // performs prune to clean up any resources that has been deleted.
+// The applier performs its function by executing a list queue of tasks,
+// each of which is one of the steps in the process of applying a set
+// of resources to the cluster. The actual execution of these tasks are
+// handled by a StatusRunner. So the taskqueue is effectively a
+// specification that is executed by the StatusRunner. Based on input
+// parameters and/or the set of resources that needs to be applied to the
+// cluster, different sets of tasks might be needed.
 type Applier struct {
 	factory   util.Factory
 	ioStreams genericclioptions.IOStreams
@@ -57,7 +61,7 @@ type Applier struct {
 	ApplyOptions  *apply.ApplyOptions
 	StatusOptions *StatusOptions
 	PruneOptions  *prune.PruneOptions
-	statusPoller  poller
+	statusPoller  poller.Poller
 
 	NoPrune bool
 	DryRun  bool
@@ -119,7 +123,7 @@ func (a *Applier) SetFlags(cmd *cobra.Command) error {
 
 // newStatusPoller sets up a new StatusPoller for computing status. The configuration
 // needed for the poller is taken from the Factory.
-func (a *Applier) newStatusPoller() (poller, error) {
+func (a *Applier) newStatusPoller() (poller.Poller, error) {
 	config, err := a.factory.ToRESTConfig()
 	if err != nil {
 		return nil, errors.WrapPrefix(err, "error getting RESTConfig", 1)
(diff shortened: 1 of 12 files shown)


mortent:
This is ready for review, but putting a hold since we don't need to merge until after this weeks release goes out.
/hold

pkg/apply/applier.go
```diff
@@ -188,6 +185,65 @@ func splitInfos(infos []*resource.Info) ([]*resource.Info, []*resource.Info) {
 	return resources, groupingObjectTemplates
 }
 
+// buildTaskQueue takes the slice of infos and object identifiers, and
+// builds a queue of tasks that needs to be executed.
+func (a *Applier) buildTaskQueue(infos []*resource.Info, identifiers []object.ObjMetadata, eventChannel chan event.Event) chan taskrunner.Task {
```
seans3: The event channel appears tightly coupled with the applier; I think we should make it a field in the applier struct instead of passing it around.

pkg/apply/applier.go
```diff
@@ -188,6 +185,65 @@ func splitInfos(infos []*resource.Info) ([]*resource.Info, []*resource.Info) {
...
+					Type: event.PruneType,
+					PruneEvent: event.PruneEvent{
+						Type: event.PruneEventCompleted,
+					},
+				},
+				EventChannel: eventChannel,
+			})
+	}
+
+	taskQueue := make(chan taskrunner.Task, len(tasks))
```
seans3: Would it be simpler to add the taskQueue as a field of the applier struct?

pkg/apply/applier.go
```diff
@@ -188,6 +185,65 @@ func splitInfos(infos []*resource.Info) ([]*resource.Info, []*resource.Info) {
 	return resources, groupingObjectTemplates
 }
 
+// buildTaskQueue takes the slice of infos and object identifiers, and
```
seans3: The `infos` and the `identifiers` both represent objects (or resources). Can we put more information in this comment on these parameters and which objects they represent?

pkg/apply/applier.go
```diff
@@ -197,12 +253,12 @@ func splitInfos(infos []*resource.Info) ([]*resource.Info, []*resource.Info) {
 // cancellation or timeout will only affect how long we wait for the
 // resources to become current.
 func (a *Applier) Run(ctx context.Context) <-chan event.Event {
-	ch := make(chan event.Event)
+	eventChannel := make(chan event.Event)
```
seans3: As mentioned earlier; it might be easier to make this channel a field of the applier, and build it in the `Initialize()`

pkg/apply/poller/poller.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"context"
+
+	"sigs.k8s.io/cli-utils/pkg/kstatus/polling"
+	pollevent "sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
+	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+// Poller defines the interface the applier needs to poll for status of resources.
```
seans3: Can we put some information on the parameters here?

pkg/apply/applier.go
```diff
@@ -42,11 +44,6 @@ func NewApplier(factory util.Factory, ioStreams genericclioptions.IOStreams) *Ap
 	}
 }
 
-// poller defines the interface the applier needs to poll for status of resources.
-type poller interface {
-	Poll(ctx context.Context, identifiers []object.ObjMetadata, options polling.Options) <-chan pollevent.Event
-}
-
 // Applier performs the step of applying a set of resources into a cluster,
```
seans3: We should probably put a couple paragraphs of comments in the applier describing how the different pieces work together in the "queue of tasks".

seans3: We should probably put a couple paragraphs of detailed comments in the `applier` which describes how all the pieces work together in the "queue of tasks".

pkg/apply/applier.go
```diff
@@ -188,6 +185,65 @@ func splitInfos(infos []*resource.Info) ([]*resource.Info, []*resource.Info) {
...
 
+// buildTaskQueue takes the slice of infos and object identifiers, and
+// builds a queue of tasks that needs to be executed.
+func (a *Applier) buildTaskQueue(infos []*resource.Info, identifiers []object.ObjMetadata, eventChannel chan event.Event) chan taskrunner.Task {
+	tasks := []taskrunner.Task{
+		&task.ApplyTask{
+			Objects:      infos,
+			ApplyOptions: a.ApplyOptions,
+		},
+		&task.SendEventTask{
```
seans3: It's not clear why we're addling the `SendEventTask` at the same time as the `ApplyTask`; maybe a comment?

pkg/apply/applier.go
mortent: I've added some more info here. I think the documentation in the taskrunner package does a decent job of explaining exactly how the tasks are executed.

pkg/apply/applier.go
mortent: I've added additional comments here. It is a bit awkward that we need to work with both the infos and the identifiers, but I can't see any obvious ways to fix this. We should keep looking for ways to simplify this.

pkg/apply/applier.go
mortent: The `eventChannel` is passed in on each call to `Run`, so it doesn't feel right to turn it into a field on the struct. Values that are used to configure the applier should be fields on the struct, while values passed into the applier through the `Run` function should not be fields. 
Also, I would like to keep the statements that create the channel and the statement that closes it as close together as possible.

pkg/apply/applier.go
mortent: I have added comments to all the tasks added to the queue.

pkg/apply/applier.go
mortent: Similar to the `eventChannel`, the queue is specific to each call to the `Run` function, so it doesn't feel right to make it a field on the struct.

pkg/apply/applier.go
mortent: See the comment above. Eventually I was hoping we might be able to move away from the `Initialize` function. We are still a bit away from this, but I think that should be a long-term goal.

pkg/apply/poller/poller.go
mortent: Done.

KnVerey:
👌 This does look very flexible!
Quick question: if I remember correctly, we wanted to support watching pruned objects until they disappear. How does that work in this setup?

```go
Apply(desired_objects)
Watch(desired_objects)
Prune(desired_objects) // incidentally, if I didn't know better,  "prune(thing)" reads like it would delete the passed in things
Watch(???) // am I supposed to extract the targets from the previous task's event stream?
```

PR #103, opened 2020-03-17
Adds error if grouping object namespace does not already exist.

seans3:
* Adds validation step during `apply` which checks if trying to add resources to namespace also being added. This will fail for the grouping object which is applied first.
* Tested manually.

/sig cli
/priority important-soon

```release-note
NONE
```

### DIFF
--- a/pkg/apply/prune/grouping_errors.go
+++ b/pkg/apply/prune/grouping_errors.go
@@ -1,12 +1,15 @@
 // Copyright 2019 The Kubernetes Authors.
 // SPDX-License-Identifier: Apache-2.0
 //
-// The error returned when applying resources, but not
-// finding the required grouping object template.
+// Grouping object errors.
 
 package prune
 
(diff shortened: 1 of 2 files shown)


pkg/apply/prune/grouping_errors.go
```diff
@@ -21,6 +24,13 @@ const multipleGroupingErrorStr = `Package has multiple grouping object templates
 The package should have one and only one grouping object template.
 `
 
+var groupingObjNamespaceError = `Attempting to apply namespace at the same time as the applied resources.
+
+Namespace: %s
+
+Currently, the namespace that applied resources are to be inserted into must exist before applying.
```
mortent: I don't think we need to use the word Currently here. Let's just say that this is not supported.

mortent:
After discussions we have decided to hold off on this change. There are some scenarios we still need to work out.
/hold

seans3:
We'll address this in the future--closing.

PR #112, opened 2020-03-28
BasicPrinter should only print status while waiting

mortent:
With the change to use the taskqueue, the event channel will send status update events throughout the process of apply and pruning. This change updates the printer to only output the status info when we are waiting for resource to reconcile. I'm not sure what is the best behavior here, but this at least keeps it consistent with the previous behavior.

This also does a little bit of cleanup in the printer code.

@seans3

### DIFF
--- a/pkg/apply/basic_printer.go
+++ b/pkg/apply/basic_printer.go
@@ -14,6 +14,7 @@ import (
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"sigs.k8s.io/cli-utils/pkg/apply/event"
 	pollevent "sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
+	"sigs.k8s.io/cli-utils/pkg/object"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/apply/basic_printer.go
```diff
@@ -49,82 +50,139 @@ func (a *applyStats) sum() int {
...
-				name := getName(obj)
-				deleteCount++
-				printFunc("%s %s", resourceIDToString(gvk.GroupKind(), name), "deleted")
-			}
+			b.processDeleteEvent(e.DeleteEvent, deleteStats, printFunc)
 		}
 	}
 }
 
+func (b *BasicPrinter) processApplyEvent(ae event.ApplyEvent, as *applyStats,
```
seans3: It should be easier to write unit tests for each of these `processEvent` functions. We can do that in future PR's.

seans3:
/lgtm

This refactor looks like it makes unit testing easier. We should add those in future PR's.

PR #113, opened 2020-03-31
Updates prune to use UID instead of ObjMetadata

seans3:
* Updates prune to use object UID to determine object equality, instead of four pieces of metadata: group, kind, namespace, name (ObjMetadata). Since objects which are equal can have different group/kind (e.g. extensions/Deployment : apps/Deployment), the ObjMetadata is inadequate for prune set calculation.
* Updated algorithm compares the UID of every previously applied object to the set UID's of currently applied objects; deleting previously applied objects which are not in the current UID set.
* A consequence of this change is more previously appied objects are retrieved from the APIServer (in order to get their UID).

/sig cli
/kind cleanup


pkg/apply/applier.go
```diff
@@ -35,12 +35,15 @@ import (
...
 	return &Applier{
-		ApplyOptions:  apply.NewApplyOptions(ioStreams),
+		ApplyOptions:  applyOptions,
 		StatusOptions: NewStatusOptions(),
-		PruneOptions:  prune.NewPruneOptions(),
-		factory:       factory,
-		ioStreams:     ioStreams,
+		// VisitedUids keeps track of the unique identifiers for all
+		// currently applied objects. Used to calculate prune set.
+		PruneOptions: prune.NewPruneOptions(&applyOptions.VisitedUids),
```
mortent, line 44: I find this a little bit strange. 
First, as far as I can tell, `VisitedUids` is a map, so I don't think we need to make this a pointer.
Also, it seems like this set of uids are only used in the `Prune` function. I think it would be easier to understand what is going on if we could just take the list of Uids from the `ApplyOptions` after the `ApplyOptions.Run` has completed and then pass it into the `PruneOptions.Prune` function as a parameter.

seans3: Updated to not use a pointer to the string set.

Maybe I should add more comments to describe the relationship between the `apply` and the `prune` with this set of UID's. The set of UID's is shared between the `apply` and the `prune`. The `apply` populates the string set with UID's from the current applied objects (this was already happening in the `kubectl apply`), while the prune reads these UID's afterward to determine what has been omitted. I don't know how we would get this set after the `ApplyOptions.Run()`, since that happens in another goroutine in the `ApplyTask`. Additionally, if we start passing this set around, it would add a noticeable amount of complexity by changing function signatures, while the current implementation of sharing the set at initialization is very simple and straightforward. I'm leaning toward keeping this implementation.

mortent: So sharing a data structure like this seems unsafe in a couple of ways:
- It breaks encapsulation, so it is hard to understand the consequences of making changes. Reassigning a new map to the `VisistedUids` variable in `ApplyOptions` would break this, but it is not obvious without understanding how prune and apply interact.
- Sharing a datastructure between code that runs in different goroutines is not safe without synchronization. It can lead to some very subtle bugs.

I just find that this kind of solution is hard to understand since it is not at all obvious without looking pretty deep in the details that `ApplyOptions.Run` must be complete before invoking `PruneOptions.Prune`.

I do agree that it is not obvious how to do this differently. Seems like we might need a way to pass data between the tasks in the taskqueue. I will take a look at it.

seans3: Some of this discussion happened offline, but I'll document it here. The `apply` must always complete before the `prune` starts for any and all conditions. If our code does not ensure this, then the code is fundamentally broken and it must be fixed. But I'm confident this condition is true. So sharing this data structure is safe because writing the UID to the set is complete before the reading of the string set begins. Another check I performed at Morten's suggestion was compiling/running prune with the `-race` flag. This race condition check gave no warnings.

seans3: Additionally, I added comments for the `currentUids` field in the `PruneOptions` to document that the structure is shared and a note of the condition that applying must complete before prune begins.

PR #114, opened 2020-04-01
Change grouping object to inventory object

seans3:
* Resolves confusion about `grouping object` naming by renaming to `inventory object`. 
* Identifiers have been changed to `InventoryID` from `GroupName`.
* Changes `-g` group name flag to `-i` inventory id flag.
* Updates unit tests.

### DIFF
--- a/cmd/initcmd/cmdinit.go
+++ b/cmd/initcmd/cmdinit.go
@@ -24,6 +24,6 @@ func NewCmdInit(ioStreams genericclioptions.IOStreams) *cobra.Command {
 			cmdutil.CheckErr(io.Run())
 		},
 	}
-	cmd.Flags().StringVarP(&io.GroupName, "group-name", "g", "", "Name to group applied resources. Must be composed of valid label characters.")
+	cmd.Flags().StringVarP(&io.InventoryID, "inventory-id", "i", "", "Identifier for group of applied resources. Must be composed of valid label characters.")
 	return cmd
 }
--- a/pkg/config/initoptions.go
+++ b/pkg/config/initoptions.go
@@ -18,21 +18,21 @@ import (
 	"github.com/google/uuid"
 )
 
-const manifestFilename = "grouping-object-template.yaml"
+const manifestFilename = "inventory-template.yaml"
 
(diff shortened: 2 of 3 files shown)


mortent:
So I think I would prefer that we call it the `prune inventory object`. I feel that just `inventory object` is a little too generic and adding prune to the name makes it somewhat clearer what the purpose of the object is.

cmd/initcmd/cmdinit.go
```diff
@@ -24,6 +24,6 @@ func NewCmdInit(ioStreams genericclioptions.IOStreams) *cobra.Command {
 			cmdutil.CheckErr(io.Run())
 		},
 	}
-	cmd.Flags().StringVarP(&io.GroupName, "group-name", "g", "", "Name to group applied resources. Must be composed of valid label characters.")
+	cmd.Flags().StringVarP(&io.InventoryName, "inventory-name", "i", "", "Name to group applied resources. Must be composed of valid label characters.")
```
mortent, line 27: I think we should consider calling the value of the label the `prune inventory id` rather than use `name`. It is the identifier we use to link from the template manifest to the actual objects. And if we move towards setting the value simply to a random generated value, using `name` doesn't quite feel right. Also, it removes the ambiguity between the value in the label and the name of the actual object.

seans3: I have taken your suggestion and changed `InventoryName` to `InventoryID`. Please note that the linter forces both letters of `ID` to be capitalized. I've also changed the flag from `inventory-namel` to `inventory-id`. I believe you are right that `name` does not make sense.

However, I will stick with the `Inventory` instead of the longer `PruneInventory`. I think this is in keeping with golang's shorter identifiers, and Brian Grant has already blessed it.

seans3:
> So I think I would prefer that we call it the `prune inventory object`. I feel that just `inventory object` is a little too generic and adding prune to the name makes it somewhat clearer what the purpose of the object is.

For now I will stick with `Inventory` over the longer `PruneInventory`.

mortent:
/lgtm
We should also make sure we update all the docs to be consistent with this change.

PR #116, opened 2020-04-06
Add TaskContext to be passed between tasks

mortent:
The `TaskContext` encapsulates the `eventChannel` and `taskChannel` that was passed around as part of the task runner. It also includes a synchronized map to allow data to be passed between tasks in a safe way.


mortent:
Removed the value map from the `TaskContext` in this PR.

seans3:
/lgtm

The unit test within `pkg/apply/taskrunner` takes about thirty seconds. We should eventually streamline this test.

PR #119, opened 2020-04-10
Fix handling of clusterscoped resources in status poller

mortent:
This fixes an issue where clusterscoped resource was not handled correctly because the status poller defaulted the namespace to `default` without checking if a resource is clusterscoped. With this change, the poller engine will verify at the beginning that all namespaced resources do have their namespace set so we no longer need to handle this within the status readers.

Ref: https://github.com/GoogleContainerTools/kpt/issues/498

@seans3 @monopole

### DIFF
--- a/pkg/apply/applier.go
+++ b/pkg/apply/applier.go
@@ -294,12 +294,7 @@ func (a *Applier) Run(ctx context.Context) <-chan event.Event {
 		// and handling the grouping object.
 		infos, err := a.readAndPrepareObjects()
 		if err != nil {
-			eventChannel <- event.Event{
-				Type: event.ErrorType,
-				ErrorEvent: event.ErrorEvent{
-					Err: errors.WrapPrefix(err, "error reading resources", 1),
-				},
-			}
+			handleError(eventChannel, err)
 			return
 		}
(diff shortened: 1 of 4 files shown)


pkg/apply/applier.go
```diff
@@ -334,17 +329,21 @@ func (a *Applier) Run(ctx context.Context) <-chan event.Event {
...
-					Err: err,
-				},
-			}
+			handleError(eventChannel, err)
 		}
 	}()
 	return eventChannel
 }
 
+func handleError(eventChannel chan event.Event, err error) {
```
seans3, line 338: Maybe we can create a common or util package to house these common routines in a future PR. Looks like the same function is in the polling engine.

pkg/kstatus/polling/engine/engine.go
```diff
@@ -85,6 +86,13 @@ func (s *PollerEngine) Poll(ctx context.Context, identifiers []object.ObjMetadat
 	return eventChannel
 }
 
+func handleError(eventChannel chan event.Event, err error) {
```
seans3, line 89: Ref.

PR #122, opened 2020-04-13
Break out functionality for table printing into library

mortent:
This PR breaks out the core functionality for printing information about resources into a separate library. In follow-up PRs, the plan is to transition both the table printer for the status command and the apply command to use this instead of duplicating code.

@seans3 @monopole

### DIFF
--- a/pkg/print/common/ansii.go
+++ b/pkg/print/common/ansii.go
@@ -1,0 +1,11 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package common
+
+const (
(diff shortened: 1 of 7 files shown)


mortent:
That is a good point. I'm not sure if the actual printers should be in `pkg`, but since this is just tools for building printers, it makes sense to put in under `pkg`. I have updated the PR.

seans3:
/lgtm

Those that are used to kubectl printing might be confused by your naming of table printing, since they sound similar. We might want to eventually have a comment that in this code which describes the relationship (or lack thereof) to kubectl table printing.

PR #125, opened 2020-04-15
cli-utils diff command with one directory argument

seans3:
* cli-utils `diff` command with one directory argument, instead of default kubectl flags.
* Moves `path.go` and unit test to `pkg/common` directory and exports `DemandOneDirectory()` to make it reusable.
* Removes ability to do diff server-side--only client-side.
* Removes extraneous flags.

### DIFF
--- a/cmd/diff/cmddiff.go
+++ b/cmd/diff/cmddiff.go
@@ -6,10 +6,72 @@ package diff
 import (
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/kubectl/pkg/cmd/apply"
 	"k8s.io/kubectl/pkg/cmd/diff"
 	"k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/util/i18n"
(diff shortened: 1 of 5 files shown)


cmd/diff/cmddiff.go
```diff
@@ -4,12 +4,102 @@
...
+	o.ServerSideApply = false
+	o.ForceConflicts = false
+
+	return nil
+}
+
+// parseDirArg returns the validated directory path
+// or an error if there are more than one arguments,
+// or the path is not a directory.
+func parseDirArg(args []string) (string, error) {
```
mortent, line 85: This looks very similar to the code at https://github.com/kubernetes-sigs/cli-utils/blob/09643102a8fcfd633696a6547e29e44b45c117dc/pkg/apply/path.go#L28-L49

Could we avoid the duplication?

seans3: Yes. I was able to reuse the same function as the `apply` and `destroy` command--`DemandOneDirectory`. I had to move `path.go` and its unit test to a common directory `pkg/common`. Have a look.

PR #139, opened 2020-04-23
Handle errors better in table views

mortent:
Update the BaseTablePrinter with better error handling. It now prints them out properly instead of just calling panic.

@seans3

### DIFF
--- a/pkg/print/table/base.go
+++ b/pkg/print/table/base.go
@@ -28,6 +28,7 @@ type ColumnDefinition interface {
 // that should be printed.
 type ResourceStates interface {
 	Resources() []Resource
+	Error() error
 }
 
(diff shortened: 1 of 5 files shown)


pkg/print/table/base.go
```diff
@@ -136,6 +141,11 @@ func (t *BaseTablePrinter) printSubTable(resources []Resource,
 	return linePrintCount
 }
 
+func (t *BaseTablePrinter) printError(err error) int {
+	t.printOrDie("Fatal error: %v\n", err)
+	return 1
```
seans3, line 146: What does this 1 represent?

mortent: Good catch. I have added a comment.

PR #142, opened 2020-04-27
Support wait for prune in the applier

mortent:


pkg/apply/applier.go
```diff
@@ -159,13 +170,81 @@ func (a *Applier) readAndPrepareObjects() ([]*resource.Info, error) {
 		return nil, err
 	}
 
+	// Fetch the grouping label used for the current set of resources.
+	groupingLabel, err := prune.RetrieveInventoryLabel(inventoryObject.Object)
```
seans3, line 174: Should this be `inventoryLabel` instead of `groupingLabel` ?

pkg/apply/applier.go
```diff
@@ -159,13 +170,81 @@ func (a *Applier) readAndPrepareObjects() ([]*resource.Info, error) {
 		return nil, err
 	}
 
+	// Fetch the grouping label used for the current set of resources.
```
seans3, line 173: Same

seans3:
Looks good. A simple naming (`grouping` to `inventory`) issue.

pkg/apply/applier.go
mortent: fixed

pkg/apply/applier.go
mortent: Fixed.

PR #143, opened 2020-04-27
Add StdIn as an acceptable argument for apply, preview, destroy, and diff commands

seans3:
* Adds StdIn as an acceptable argument to the following commands: `apply`, `preview`, `destroy`, and `diff`.
* Updates `Use:` in command
* Adds unit tests
* Tested manually

### DIFF
--- a/cmd/diff/cmddiff.go
+++ b/cmd/diff/cmddiff.go
@@ -19,7 +19,7 @@ import (
 func NewCmdDiff(f util.Factory, ioStreams genericclioptions.IOStreams) *cobra.Command {
 	options := diff.NewDiffOptions(ioStreams)
 	cmd := &cobra.Command{
-		Use:                   "diff DIRECTORY",
+		Use:                   "diff (DIRECTORY | STDIN)",
 		DisableFlagsInUseLine: true,
 		Short:                 i18n.T("Diff local config against cluster applied version"),
 		Args:                  cobra.MaximumNArgs(1),
--- a/pkg/common/path.go
+++ b/pkg/common/path.go
@@ -13,32 +13,43 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 )
 
+const stdinDash = "-"
+
 func processPaths(paths []string) genericclioptions.FileNameFlags {
 	// No arguments means we are reading from StdIn
 	fileNameFlags := genericclioptions.FileNameFlags{}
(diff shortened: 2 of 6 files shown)


pkg/common/path.go
```diff
@@ -14,30 +14,43 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 )
 
+const stdinDash = "-"
```
mortent, line 17: I'm not sure we should use this convention. I know we do it with kubectl, but I think we can just assume that no arguments means stdin.

seans3: Right now, both "no arguments" and "dash" mean StdIn. If you think it  should only be "no arguments", please let me know.

mortent: I think I slightly prefer that we don't support the dash argument. It is easier to add it later than to remove it once it is in.

seans3: Removed dash as an argument; only "no argument" means stdin now.

pkg/common/path.go
```diff
@@ -13,32 +13,43 @@ import (
...
 	}
 
+	// Must be a single directory here; set recursive flag.
 	t := true
 	fileNameFlags.Filenames = &paths
 	fileNameFlags.Recursive = &t
 	return fileNameFlags
 }
 
+// DemandOneDirectoryOrStdin processes "paths" to ensure the
```
seh, line 34: This doesn't match the function name `DemandOneDirectory`. Did you intend to rename the function to match the comment?

cmd/diff/cmddiff.go
```diff
@@ -19,7 +19,7 @@ import (
 func NewCmdDiff(f util.Factory, ioStreams genericclioptions.IOStreams) *cobra.Command {
 	options := diff.NewDiffOptions(ioStreams)
 	cmd := &cobra.Command{
-		Use:                   "diff DIRECTORY",
+		Use:                   "diff (DIRECTORY | STDIN)",
```
seh, line 22: I find that this fails in what appears to be `common.ExpandPackageDir` and `common.expandDir`:

> % ... | kpt live diff   
> error: lstat -: no such file or directory

It looks like we need to thread this treatment of the single hyphen (meaning standard input) deeper into the call path.

seh: See #170.

PR #144, opened 2020-04-27
Removes inventory.go and inventory_test.go

seans3:
* Removes `inventory.go` and `inventory_test.go`.
* Implements simple union functionality `unionPastObjs()` in `prune.go`.
* Part of effort to cleanup after renaming `grouping object` to `inventory object`.
* Much more cleanup to come in future PRs


pkg/apply/prune/prune_test.go
```diff
@@ -159,18 +159,33 @@ func TestUnionPastInventory(t *testing.T) {
...
 			}
 		})
 	}
 }
 
+// objInArray is a helper function that returns true if passed obj
+// is in array of objects; false otherwise.
+func objInArray(obj *object.ObjMetadata, arr []*object.ObjMetadata) bool {
+	for _, a := range arr {
+		if a.String() == obj.String() {
```
mortent, line 182: I think we should do a proper `Equals` function for this. Or if we can avoid using pointers here, I think we could just use `a == obj`.

seans3: I think I've addressed your concern. I've avoided pointers here, and used `a == obj`. Let me know if this works.

pkg/apply/prune/prune.go
```diff
@@ -161,21 +161,27 @@ func infoToObjMetadata(info *resource.Info) (*object.ObjMetadata, error) {
...
-// grouping objects, or if unable to retrieve the inventory from any
-// grouping object.
-func unionPastInventory(infos []*resource.Info) (*Inventory, error) {
-	inventorySet := NewInventory([]*object.ObjMetadata{})
+// unionPastObjs takes a set of inventory objects (infos), returning the
+// union of the objects referenced by these inventory objects.
+// Returns an error if any of the passed objects are not inventory
+// objects, or if unable to retrieve the referenced objects from any
+// inventory object.
+func unionPastObjs(infos []*resource.Info) ([]*object.ObjMetadata, error) {
```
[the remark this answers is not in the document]
seans3: Sounds good. Changed from returning `[]*object.ObjMetadata` to `[]object.ObjMetadata`.

PR #153, opened 2020-05-01
Handle CRDs and CRs in the same set of resources

mortent:
POC on how we can be able to read manifests with the Builder, even when there are both CRDs and CRs among the set of resources. This uses the local setting on the Builder to get info objects that doesn't have a mapping or a client. We can then create the appropriate tasks for the queue and update the infos for CRs only when we have already applied the CRDs.

This still needs cleanup and tests before it is merged, but wanted to get feedback on whether this approach could work.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -7,9 +7,14 @@ import (
 	"flag"
 	"os"
 	"strings"
+	"sync"
 
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/util/logs"
 	"sigs.k8s.io/cli-utils/cmd/apply"
@@ -35,7 +40,9 @@ func main() {
 	flags := cmd.PersistentFlags()
 	kubeConfigFlags := genericclioptions.NewConfigFlags(true).WithDeprecatedPasswordFlag()
 	kubeConfigFlags.AddFlags(flags)
-	matchVersionKubeConfigFlags := util.NewMatchVersionFlags(kubeConfigFlags)
+	matchVersionKubeConfigFlags := util.NewMatchVersionFlags(&cachingRESTClientGetter{
+		delegate: kubeConfigFlags,
+	})
 	matchVersionKubeConfigFlags.AddFlags(cmd.PersistentFlags())
 	cmd.PersistentFlags().AddGoFlagSet(flag.CommandLine)
 	f := util.NewFactory(matchVersionKubeConfigFlags)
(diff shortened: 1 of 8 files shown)


cmd/main.go
```diff
@@ -79,3 +86,32 @@ func updateHelp(names []string, c *cobra.Command) {
 		c.Example = strings.ReplaceAll(c.Example, "kubectl "+name, "kapply "+name)
 	}
 }
+
+// cachingRESTClientGetter caches the RESTMapper so every call to
```
seans3, line 90: Eventually, we might want to put this in its own file in another package.

pkg/apply/applier.go
```diff
@@ -142,18 +154,130 @@ func (a *Applier) newStatusPoller() (poller.Poller, error) {
...
+
+	var infos []*resource.Info
+	err := r.Visit(func(info *resource.Info, err error) error {
+		infos = append(infos, info)
+		return nil
+	})
+	return infos, err
+}
+
+// setNamespaces verifies that
```
seans3, line 196: Is this comment incomplete?

pkg/apply/applier.go
```diff
@@ -142,18 +154,130 @@ func (a *Applier) newStatusPoller() (poller.Poller, error) {
...
+		FilenameParam(a.ApplyOptions.EnforceNamespace, &a.ApplyOptions.DeleteOptions.FilenameOptions).
+		LabelSelectorParam(a.ApplyOptions.Selector).
+		Flatten().
+		Do()
+	if err := r.Err(); err != nil {
+		return nil, err
+	}
+
+	var infos []*resource.Info
+	err := r.Visit(func(info *resource.Info, err error) error {
```
seans3, line 189: I don't think all of this using "Visit" is necessary. I think it's possible to just call "r.Infos()" to return the infos.

pkg/apply/applier.go
```diff
@@ -142,18 +154,130 @@ func (a *Applier) newStatusPoller() (poller.Poller, error) {
...
+
+		if err != nil && !meta.IsNoMatchError(err) {
+			return err
+		}
+
+		if err == nil {
+			// If we find a mapping for the resource type in the RESTMapper,
+			// we just use it.
+			if mapping.Scope == meta.RESTScopeNamespace {
+				inf.Namespace = defaultNamespace
```
seans3, line 230: Why are we setting to "defaultNamespace" when mapping is found? Where is "defaultNamespace" set?

seans3:
I think we should add some tests that prove the namespace added to the CRD works as we expect. I've run into issues with this before.

pkg/apply/applier.go
```diff
@@ -142,18 +154,130 @@ func (a *Applier) newStatusPoller() (poller.Poller, error) {
...
+// handles ordering of resources and sets up the grouping object
+// based on the provided grouping object template.
 func (a *Applier) readAndPrepareObjects() (*ResourceObjects, error) {
-	infos, err := a.ApplyOptions.GetObjects()
+	infos, err := a.manifestReaderFunc()
 	if err != nil {
 		return nil, err
 	}
+
+	err = a.setNamespaces(infos, a.ApplyOptions.Namespace)
```
seans3, line 276: I believe ApplyOptions.Namespace is only set using the "-n" flag on the command line or "namespace" in the kubeconfig file. I don't think its usually set. If any of these two are set, then we also have to look at "EnforceNamespace" field in the builder. We should probably write some tests for this.

cmd/main.go
mortent: Ack

pkg/apply/applier.go
mortent: Yeah, you are right. This was something that was left over from a previous iteration on this. Fixed now.

pkg/apply/applier.go
mortent: Fixed.

pkg/apply/applier.go
mortent: Mapping found means that we can tell whether the resource is namespace-scoped or cluster-scoped. And if it is namespace-scoped, we need to set the namespace.

pkg/apply/applier.go
mortent: You are correct. I have updated the code to also verify that the namespace matches `ApplyOptions.Namespace` for resources that already have a namespace if `EnforceNamespace` is true. Also added tests for this.

pkg/apply/applier.go
```diff
@@ -142,18 +154,139 @@ func (a *Applier) newStatusPoller() (poller.Poller, error) {
...
+		}
+	}
+
+	return nil
+}
+
 // readAndPrepareObjects reads the resources that should be applied,
-// handles ordering of resources and sets up the inventory object
-// based on the provided inventory object template.
+// handles ordering of resources and sets up the grouping object
```
seans3, line 277: Are we moving back to `grouping` object naming; or should we stick with `inventory` for the time being?

pkg/apply/applier.go
```diff
@@ -142,18 +154,139 @@ func (a *Applier) newStatusPoller() (poller.Poller, error) {
...
+// based on the provided grouping object template.
```
seans3, line 278: Same

pkg/apply/applier.go
mortent: This was a leftover from before we changed it. We definitely want to stick with the 'inventory' name. I have updated the PR

PR #155, opened 2020-05-03
Allow users to specify deletion propagation for prune

mortent:
Allow users to specify the deletion propagation policy for pruning.

### DIFF
--- a/cmd/apply/cmdapply.go
+++ b/cmd/apply/cmdapply.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/util/i18n"
@@ -26,10 +27,9 @@ func GetApplyRunner(f cmdutil.Factory, ioStreams genericclioptions.IOStreams) *A
 		Use:                   "apply (DIRECTORY | STDIN)",
 		DisableFlagsInUseLine: true,
 		Short:                 i18n.T("Apply a configuration to a resource by package directory or stdin"),
-		Run:                   r.Run,
+		RunE:                  r.RunE,
 	}
 
-	cmd.Flags().BoolVar(&r.noPrune, "no-prune", r.noPrune, "If true, do not prune previously applied objects.")
 	cmdutil.CheckErr(r.applier.SetFlags(cmd))
 
 	// The following flags are added, but hidden because other code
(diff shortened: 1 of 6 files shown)


pkg/apply/prune/prune.go
```diff
@@ -190,13 +190,16 @@ type Options struct {
...
+
+	PropagationPolicy metav1.DeletionPropagation
 }
 
 // Prune deletes the set of resources which were previously applied
 // (retrieved from previous inventory objects) but omitted in
 // the current apply. Prune also delete all previous inventory
 // objects. Returns an error if there was a problem.
 func (po *PruneOptions) Prune(currentObjects []*resource.Info, eventChannel chan<- event.Event, o Options) error {
+	fmt.Printf("Pruning with prop policy %s\n", string(o.PropagationPolicy))
```
seans3, line 202: Should this be removed?

cmd/apply/cmdapply.go
```diff
@@ -68,14 +72,20 @@ type ApplyRunner struct {
...
+	wait                   bool
+	period                 time.Duration
+	timeout                time.Duration
+	noPrune                bool
+	prunePropagationPolicy string
 }
 
-func (r *ApplyRunner) Run(cmd *cobra.Command, args []string) {
+func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
+	prunePropPolicy, err := convertPropagationPolicy(r.prunePropagationPolicy)
```
seans3, line 84: Should we add a comment somewhere that the default propagation policy is `background`

pkg/apply/task/prune_task.go
```diff
@@ -13,9 +14,10 @@ import (
 // by using the PruneOptions. The provided Objects is the
 // set of resources that have just been applied.
 type PruneTask struct {
-	PruneOptions *prune.PruneOptions
-	Objects      []*resource.Info
-	DryRun       bool
+	PruneOptions      *prune.PruneOptions
+	Objects           []*resource.Info
+	DryRun            bool
+	PropagationPolicy metav1.DeletionPropagation
```
seans3, line 20: Should this be separate from the `PruneOptions` or part of it?

seans3:
Looks good. Thanks for taking the initiative on this. I had a couple questions.

pkg/apply/prune/prune.go
mortent: Yeah, this shouldn't be here..

cmd/apply/cmdapply.go
mortent: We are setting the default value in the cobra command. So running `apply --help` will show that the default is `Background`. I have also updated this PR so the Options passed into the Applier will also handle defaulting to Background correctly.

pkg/apply/task/prune_task.go
mortent: So we are passing the PropagationPolicy into the `Prune` function as part of the options parameter object. I prefer this solution over setting values directly on the `PruneOptions`. After we moved `DryRun` from being a field on `PruneOptions` to being passed into the `Prune` function, the Applier only interacts with the `PruneOptions` through functions. This means we can use an interface to help with testing.

PR #156, opened 2020-05-03
Print more info on wait timeout

mortent:
This updates the error message printed after a wait timeout to include a list of the resources that failed to reach the desired status. This only covers the output for the table printer, I will do a separate PR to make sure the events printer also includes this information in the error message.

Ref: https://github.com/GoogleContainerTools/kpt/issues/582

### DIFF
--- a/cmd/printers/table/collector.go
+++ b/cmd/printers/table/collector.go
@@ -4,12 +4,15 @@
 package table
 
 import (
+	"bytes"
+	"fmt"
 	"sort"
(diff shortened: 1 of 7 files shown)


seans3:
* I think we should eventually move `cmd/printers` to `pkg`, since there is no `cobra.Command` in that directory, but there is in every other `cmd` dir.
* I think we should eventually add a unit test to `cmd/printers/table/collector.go`

/lgtm

PR #159, opened 2020-05-04
Never delete lifecycle directive

seans3:
* Adds initial implementation of "prevent delete" lifecycle directive.
* This version stores annotation in `pkg/common`, but this will probably change.
* Adds unit tests
* Tested manually as well

### DIFF
--- a/pkg/apply/prune/prune.go
+++ b/pkg/apply/prune/prune.go
@@ -224,7 +224,7 @@ func (po *PruneOptions) Prune(currentObjects []*resource.Info, eventChannel chan
 		namespacedClient := po.client.Resource(mapping.Resource).Namespace(past.Namespace)
 		obj, err := namespacedClient.Get(past.Name, metav1.GetOptions{})
 		if err != nil {
-			// Object not found -- skip it and move to the next object.
+			// Object not found in cluster, so no need to delete it; skip to next object.
 			if apierrors.IsNotFound(err) {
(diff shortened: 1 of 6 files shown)


pkg/apply/prune/prune.go
```diff
@@ -241,6 +241,20 @@ func (po *PruneOptions) Prune(currentObjects []*resource.Info, eventChannel chan
 		if po.currentUids.Has(uid) {
 			continue
 		}
+		// Handle lifecycle directive preventing deletion.
+		if preventDeleteAnnotation(metadata.GetAnnotations()) {
+			eventChannel <- event.Event{
+				Type: event.ApplyType,
```
mortent, line 247: Using the `Apply` event type doesn't seem right here. The annotation is about pruning, not apply.

seans3: Added new `PruneEventType` called `PruneEventSkipped`. My initial pass at printing this looks like:

```
$ kapply apply ~/testdata/testdata-2
configmap/inventory-aa61e798 created
namespace/test-namespace unchanged
pod/pod-b unchanged
pod/pod-c unchanged
pod/pod-d created
5 resource(s) applied. 2 created, 3 unchanged, 0 configured
pod/pod-a pruned skipped
configmap/inventory-1e5824fb pruned
1 resource(s) pruned
```

Let me know what you think.

PR #171, opened 2020-05-19
First pass at inventory polymorphism POC

seans3:
Proof-of-Concept for creating a flexible mechanism for different types of inventory objects beyond the current inventory config map.

* Adds `Inventory` interface to load/store `ObjMetadata` in/out of an inventory object and `GetObject()`.
* Adds `InventoryFactoryFunc()` to wrap the inventory config map and return an `Inventory` interface.
* Adds `InventoryConfigMap` struct to wrap the inventory config map; implements `Inventory`.

### DIFF
--- a/pkg/apply/applier.go
+++ b/pkg/apply/applier.go
@@ -47,10 +47,11 @@ func NewApplier(factory util.Factory, ioStreams genericclioptions.IOStreams) *Ap
 		factory:      factory,
 		ioStreams:    ioStreams,
 	}
-	a.previousInventoriesFunc = a.previousInventories
+	a.previousInventoriesFunc = a.PruneOptions.GetPreviousInventoryObjects
 	a.infoHelperFactoryFunc = a.infoHelperFactory
 	a.manifestReaderFunc = a.readObjects
 	a.InventoryFactoryFunc = prune.WrapInventoryObj
+	a.PruneOptions.InventoryFactoryFunc = prune.WrapInventoryObj
 
 	return a
 }
@@ -77,9 +78,10 @@ type Applier struct {
 	// so we don't need these functions to facilitate testing.
 	//
 	// previousInventoriesFunc is used to fetch the existing inventories
-	// from the apiserver. It is defined here so we can override it in
+	// from the apiserver, taking the current inventory object as the
+	// parameter. It is defined here so we can override it in
 	// unit tests.
-	previousInventoriesFunc func(label, namespace string) ([]*resource.Info, error)
+	previousInventoriesFunc func(*resource.Info) ([]prune.Inventory, error)
 	// infoHelperFactoryFunc is used to create a new instance of the
 	// InfoHelper. It is defined here so we can override it in unit tests.
 	infoHelperFactoryFunc func() info.InfoHelper
@@ -159,12 +161,6 @@ func (a *Applier) newStatusPoller() (poller.Poller, error) {
 	}
 
 	return polling.NewStatusPoller(c, mapper), nil
-}
-
-// previousInventories returns the infos for all the inventory objects
-// in the cluster.
-func (a *Applier) previousInventories(label, namespace string) ([]*resource.Info, error) {
-	return a.PruneOptions.RetrievePreviousInventoryObjects(label, namespace)
 }
 
 // infoHelperFactory returns a new instance of the InfoHelper.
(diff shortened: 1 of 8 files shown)


pkg/apply/applier.go
```diff
@@ -300,7 +305,8 @@ func (a *Applier) readAndPrepareObjects() (*ResourceObjects, error) {
 		}
 	}
 
-	inventoryObject, err := prune.CreateInventoryObj(invs[0], resources)
+	inv := a.InventoryFactoryFunc(invs[0])
```
mortent, line 308: So doesn't the `splitInfos` function also need to adjust so it can find the correct inventory objects?

pkg/apply/prune/inventorycm.go
```diff
@@ -0,0 +1,132 @@
...
+	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+func WrapInventoryObj(info *resource.Info) Inventory {
+	return &InventoryConfigMap{inv: info}
+}
+
+type InventoryConfigMap struct {
+	inv      *resource.Info
```
mortent, line 22: Do we need these implementations to hold state? I was thinking it would be cleaner if we kept whatever state we need in the applier, and just call out to the inventory implementation whenever the code is specific to the type.

pkg/object/objmetadata.go
```diff
@@ -29,6 +29,19 @@ import (
 // resource name.
 const fieldSeparator = "_"
 
+type ObjLoader interface {
```
mortent, line 32: Is it useful to split this into so many separate interfaces? I thought it might be easier to understand if we had it all in the Inventory interface.

pkg/apply/applier.go
seans3: No. `splitInfos` will return the objects with the `inventory-id` label, regardless of the resource type. `splitInfos` returns the inventory template as a `*resource.Info`

pkg/apply/applier.go
```diff
@@ -50,6 +50,8 @@ func NewApplier(factory util.Factory, ioStreams genericclioptions.IOStreams) *Ap
 	a.previousInventoriesFunc = a.previousInventories
 	a.infoHelperFactoryFunc = a.infoHelperFactory
 	a.manifestReaderFunc = a.readObjects
+	a.InventoryFactoryFunc = prune.WrapInventoryObj
```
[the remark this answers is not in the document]
seans3: I don't think we could set a property, since the logic for creating the type might not live in this repository. Let's discuss this in person. If we move in another direction, I think this would be the best first implementation since it is a small amount of code. Whatever we implement, we can not import the overriden type into this code base.

pkg/apply/prune/inventorycm.go
seans3: I don't understand this. Let's talk in person.

pkg/object/objmetadata.go
seans3: The `go` language does one function interfaces quite often, composing the smaller interfaces into larger interfaces (e.g. the `io` stuff with `Reader`, `Writer`, `Closer`, and `ReadWriter`, `ReadWriteCloser`). But I see your point that maybe this belongs in the `inventory.go`.

pkg/apply/prune/inventorycm.go
```diff
@@ -0,0 +1,135 @@
...
+	objMetas []*object.ObjMetadata
+}
+
+func (icm *InventoryConfigMap) Load() ([]*object.ObjMetadata, error) {
+	objs := []*object.ObjMetadata{}
```
mortent, line 30: I don't think we should use pointers here.

seans3: Fixed

pkg/object/objmetadata.go
seans3: Fixed

pkg/apply/applier.go
```diff
@@ -47,9 +47,12 @@ func NewApplier(factory util.Factory, ioStreams genericclioptions.IOStreams) *Ap
 		factory:      factory,
 		ioStreams:    ioStreams,
 	}
-	a.previousInventoriesFunc = a.previousInventories
+	a.previousInventoriesFunc = a.PruneOptions.GetPreviousInventoryObjects
 	a.infoHelperFactoryFunc = a.infoHelperFactory
 	a.manifestReaderFunc = a.readObjects
+	a.InventoryFactoryFunc = prune.WrapInventoryObj
+	a.PruneOptions.InventoryFactoryFunc = prune.WrapInventoryObj
```
mortent, line 54: Not sure I like that we end up setting these function properties in multiple places, but we can continue to improve as we add new implementations of the Inventory interface.

PR #172, opened 2020-05-19
Move CachingRESTClientGetter into its own package

mortent:
We need to use this to set up the Factory in kpt as well, so we need this in a separate package.

@seans3


seans3:
/lgtm

This stuff is hard to do unit tests.

Issue #169, opened 2020-05-13
Feature request: Allow to inject custom poller into Applier


teoyaomiqui:
It would be very useful if Applier could be used as library with custom polling interface by, for example, allowing to inject it via [constructor](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/apply/applier.go#L37-L49) as an option, or by making statusPoller field public in [here](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/apply/applier.go#L67), or making a setter for it.

We have custom wait logic for numerous CRDs, from different projects, that we have no control over. And default Status readers, really can't handle properly waiting for those CRDs to have desired state, since these CRDs don't follow recommended status conditions, as in described in kpt [documentation](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/apply/applier.go#L67). 

So we annotate these CRDs with special annotation, which contains a logic of how to check this status, for example:
```
metadata:
  annotations:
    airshipit.org/status-check: |
      [
        {
          "status": "failed",
          "condition": "@.status.state==\"failed\""
        },
        {
          "status": "current",
          "condition": "@.status.state==\"current\""
        }
      ]
```
Based on this we can implement our own StatusReader for these CRDs, and inject it into [poller interface](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/apply/poller/poller.go#L20-L22) in [here](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/polling/polling.go#L23-L30) and add this poller interface to applier.

mortent:
@teoyaomiqui This is good feedback. I'll try to get a change in this week.

Longer term we also want to make it easier to use custom status rules, but that is not something that we will get to in the near term.

teoyaomiqui:
@mortent thank you very much for taking a look into it.
Also another note, that i thought maybe the theoretically interested to you. We are implementing our own custom poller based on [this one](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/polling/polling.go#L23-L30), however to add non default status readers, we do a full copy of the reader and expand [createStatusReaders](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/polling/polling.go#L66-L80)

So if there would be an ability to merge these status readers or inject additional ones, i think it can even further expand usage of the project. But i do think that this maybe a bit more complicated.

[...]

PR #173, opened 2020-05-19
Make Poller property on Applier public to allow Custom poller implementation

mortent:
This is a quick fix to address #169. Eventually we should try to make it easier to customize the behavior of the Applier, but it will require larger changes.

@seans3 @teoyaomiqui


teoyaomiqui:
thank you, this seems to be great place to start for us!

PR #178, opened 2020-05-28
Support preview when resource set contains both CRD and CR

mortent:
This PR updates the way we handle dry-run in the `ApplyTask`, so we can handle situations where a CRD and a CR is in the same resource set. 

There are some things I don't like in this PR, but I would like to address it in follow-up PRs as it requires larger changes.
- The `InfoHelper` doesn't really seem like the right place to get the `RESTMapper`. I want to rewrite the `InfoHelper` to separate the responsibility of updating Infos from what is needed to work with the `RESTMapper`.
- Tracking any CRDs and namespaces in the resource set is probably useful in multiple places, so we might want to do this somewhere else than in the `ApplyTask`.

### DIFF
--- a/pkg/testutil/restmapper.go
+++ b/pkg/testutil/restmapper.go
@@ -1,0 +1,21 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package testutil
+
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func NewFakeRESTMapper(gvks ...schema.GroupVersionKind) meta.RESTMapper {
+	var groupVersions []schema.GroupVersion
+	for _, gvk := range gvks {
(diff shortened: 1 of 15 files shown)


pkg/testutil/restmapper.go
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package testutil
+
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func NewFakeRESTMapper(gvks ...schema.GroupVersionKind) meta.RESTMapper {
```
seans3, line 11: Not sure if it's helpful or something we want to move to, but there is functionality for a testing RESTMapper here:

`/staging/src/k8s.io/apimachinery/pkg/api/meta/testrestmapper/test_restmapper.go:func TestOnlyStaticRESTMapper(scheme *runtime.Scheme, versionPatterns ...schema.GroupVersion) meta.RESTMapper {
`

PR #180, opened 2020-05-29
Fix output from lifecycle directive with the destroy command

mortent:
This fixes an issue where the output from the destroyer when a resource has the lifecycle directive is not correct. A resource will show up as deleted, even if it is not actually deleted. The reason is that we were missing the corresponding event type for delete that we had for prune.

This changes the structure of the event for the prune and delete events to include an `Operation` field that matches the structure for apply events. It keeps the structure of the different event types consistent.

@seans3 @phanimarupaka


seans3:
Looks good. Should we add some tests ?

PR #181, opened 2020-06-03
Separate reading manifests into Info objects from the apply logic

mortent:
This separates the logic from reading manifests from the applier. This should simplify the `Applier`, provide better separation of concerns and make it easier to use the Applier as a library.

### DIFF
--- a/cmd/apply/cmdapply.go
+++ b/cmd/apply/cmdapply.go
@@ -16,12 +16,15 @@ import (
 	"k8s.io/kubectl/pkg/util/i18n"
 	"sigs.k8s.io/cli-utils/cmd/printers"
 	"sigs.k8s.io/cli-utils/pkg/apply"
+	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/cli-utils/pkg/manifestreader"
 )
 
 func GetApplyRunner(f cmdutil.Factory, ioStreams genericclioptions.IOStreams) *ApplyRunner {
 	r := &ApplyRunner{
 		Applier:   apply.NewApplier(f, ioStreams),
 		ioStreams: ioStreams,
+		factory:   f,
 	}
 	cmd := &cobra.Command{
 		Use:                   "apply (DIRECTORY | STDIN)",
@@ -71,6 +74,7 @@ type ApplyRunner struct {
 	command   *cobra.Command
 	ioStreams genericclioptions.IOStreams
 	Applier   *apply.Applier
+	factory   cmdutil.Factory
 
 	output                 string
 	period                 time.Duration
@@ -86,7 +90,7 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	cmdutil.CheckErr(r.Applier.Initialize(cmd, args))
+	cmdutil.CheckErr(r.Applier.Initialize(cmd))
 
 	// Only emit status events if we are waiting for status.
 	//TODO: This is not the right way to do this. There are situations where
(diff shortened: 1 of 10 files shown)


pkg/manifestreader/common.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+type ReaderOptions struct {
+	Factory          util.Factory
+	Validate         bool
+	Namespace        string
+	EnforceNamespace bool
+}
+
+func setNamespaces(factory util.Factory, infos []*resource.Info,
```
seans3, line 27: This is a big function. Even though it's not exported, should we put a comment above it to summarize what it's doing?

pkg/manifestreader/common.go
```diff
@@ -0,0 +1,105 @@
...
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/cli-runtime/pkg/resource"
+	"k8s.io/kubectl/pkg/cmd/util"
+	"sigs.k8s.io/cli-utils/pkg/apply/solver"
+)
+
+type ManifestReader interface {
```
seans3, line 16: Can we put a comment what the definition of manifest is?

pkg/manifestreader/path.go
```diff
@@ -0,0 +1,53 @@
...
+// and returns them as Info objects. The returned Infos will not have
+// client or mapping set.
+type PathManifestReader struct {
+	Path string
+
+	ReaderOptions
+}
+
+// Read reads the manifests and returns them as Info objects.
+func (p *PathManifestReader) Read() ([]*resource.Info, error) {
```
seans3, line 20: We might want to add that the mapping is not set in the Info (since the builder is `Local()`).

pkg/manifestreader/stream.go
```diff
@@ -0,0 +1,51 @@
...
+// client or mapping set.
+type StreamManifestReader struct {
+	ReaderName string
+	Reader     io.Reader
+
+	ReaderOptions
+}
+
+// Read reads the manifests and returns them as Info objects.
+func (r *StreamManifestReader) Read() ([]*resource.Info, error) {
```
seans3, line 23: Same--no mapping in the Info.

pkg/apply/applier.go
```diff
@@ -168,128 +159,10 @@ func (a *Applier) infoHelperFactory() info.InfoHelper {
...
-		case "Namespaced":
-			inf.Namespace = defaultNamespace
-			accessor.SetNamespace(defaultNamespace)
-		}
-	}
-
-	return nil
-}
-
 // readAndPrepareObjects reads the resources that should be applied,
```
seans3, line 162: nit: comment should now be `prepareObjects`, since the name of the function changed.

cmd/apply/cmdapply.go
```diff
@@ -97,9 +101,36 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
 		emitStatusEvents = true
 	}
 
-	// Run the Applier. It will return a channel where we can receive updates
+	_, err = common.DemandOneDirectory(args)
```
seans3, line 104: We should eventually clean this up, since we're just ignoring the return value.

pkg/apply/applier.go
```diff
@@ -96,13 +88,12 @@ type Applier struct {
 // Initialize sets up the Applier for actually doing an apply against
 // a cluster. This involves validating command line inputs and configuring
 // clients for communicating with the cluster.
-func (a *Applier) Initialize(cmd *cobra.Command, paths []string) error {
-	fileNameFlags, err := common.DemandOneDirectory(paths)
-	if err != nil {
-		return err
+func (a *Applier) Initialize(cmd *cobra.Command) error {
```
seans3, line 91: Eventually, we want to remove the `cobra.Command` parameter, since it couples the `Applier` to the cli, and we want to use the `Applier` as a library.

pkg/manifestreader/path.go
```diff
@@ -0,0 +1,53 @@
...
+		Filenames: []string{p.Path},
+		Recursive: true,
+	}
+
+	result := p.Factory.NewBuilder().
+		Local().
+		Unstructured().
+		Schema(validator).
+		ContinueOnError().
+		FilenameParam(false, fileNameOptions).
```
seans3, line 36: Should we name this `false` parameter as a variable first to document it; othewise, we can just add a comment.

pkg/manifestreader/common.go
mortent: Done. I just missed copying the comments from the applier.

pkg/apply/applier.go
mortent: Yeah, after this change we are almost at a point where we no longer need to call `ApplyOptions.Complete`. This means we will no longer have to deal with cobra flags set inside `ApplyOptions` and we might be able to hide our usage of `ApplyOptions` inside the `ApplyTask`.

pkg/manifestreader/path.go
mortent: Done

cmd/apply/cmdapply.go
mortent: Done

pkg/apply/applier.go
mortent: Done

PR #182, opened 2020-06-04
Exit with non-zero exit code on error

fsommar:
Hi 👋

A change in exit code behavior was introduced in #166, where printing an error used to also cause the application to exit with a non-zero exit code. As a result, errors in e.g. `kpt live apply` exits the process with a 0 exit code.

This change should bring it in line with how it works before, but I'm open to other solutions if there's a cleaner way to accomplish this without having an error log automatically mean that the application needs to exit.

### DIFF
--- a/pkg/apply/basic_printer.go
+++ b/pkg/apply/basic_printer.go
@@ -5,6 +5,7 @@ package apply
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -15,6 +16,11 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/apply/taskrunner"
 	pollevent "sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
 	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+const (
+	defaultExitErrorCode int = 1
+	timeoutExitErrorCode int = -1
 )
 
 // BasicPrinter is a simple implementation that just prints the events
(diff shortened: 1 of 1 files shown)


fsommar:
I'm working on getting the CLA signed on my end :)

fsommar:
CLA should be set now.

fsommar:
I signed it

fsommar:
I signed it

mortent:
@fsommar Thanks for reporting this. Changing the exit code was not an intentional change.

It seems like returning an exit code whenever something didn't work as expected would be useful. We could maybe start with using just exit code 1 for general errors and use a different error code for timeouts. Any particular reason why you suggested exit code `-1` here?

Ideally we don't want to call `os.Exit` from the printer, but rather have the printer make sure it prints the error message, then have the printer return and have the main program do an orderly shutdown before calling exit. I think the current structure of the printers make this hard, so we can do a shutdown like the one in this PR temporarily until we can clean it up.

fsommar:
@mortent No reason why I chose `-1`, so I'm open to changing it to something else if that would work better for you.

I agree with the unfortunate placing of the exit. It'd indeed be good to have it in the interim until it can be structured better. (If you have a suggestion for how to do it I might want to give it a try.)

mortent:
So it seems like exit code 124 signals a timeout, although this usually means it was terminated by SIGTERM. I'm thinking we just decide on an exit code that isn't reserved already and make sure it is documented. Seems like everything between 3 and 126 is usually considered non-reserved, so how about we just use 3.

@fsommar It's great if you want to contribute. I'll take a look at how we might do it. It mostly involves making sure we shut down all the different goroutines in the right order.

PR #183, opened 2020-06-05
Only mark pods as unschedulable if they have been in that state for a while

mortent:
Some times the scheduler needs a little bit of time before scheduling the pods, so this makes sure we only mark pods as failed (unschedulable) if they haven't been scheduled within 15 seconds.

@seans3

### DIFF
--- a/pkg/kstatus/status/core.go
+++ b/pkg/kstatus/status/core.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math"
 	"strings"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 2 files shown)


pkg/kstatus/status/core.go
```diff
@@ -400,6 +401,11 @@ func podConditions(u *unstructured.Unstructured) (*Result, error) {
 	case "Pending":
 		c, found := getConditionWithStatus(objc.Status.Conditions, "PodScheduled", corev1.ConditionFalse)
 		if found && c.Reason == "Unschedulable" {
+			if time.Now().Add(-15 * time.Second).Before(u.GetCreationTimestamp().Time) {
```
seans3, line 404: Should we make this -15 a constant?

seans3: i.e. `const scheduleWindowSeconds = -15` or something like that

PR #189, opened 2020-06-12
Initial move of Inventory to its own package

seans3:
* First step in making `Inventory` its own package.
* Moves `inventory.go` and associated files to `pkg/inventory`.
* Updates files which use `inventory` functionality to reference this new package.

### DIFF
--- a/pkg/apply/applier.go
+++ b/pkg/apply/applier.go
@@ -23,6 +23,7 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/apply/prune"
 	"sigs.k8s.io/cli-utils/pkg/apply/solver"
 	"sigs.k8s.io/cli-utils/pkg/apply/taskrunner"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/polling"
 	"sigs.k8s.io/cli-utils/pkg/object"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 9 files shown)


pkg/inventory/inventorycm.go
```diff
@@ -6,7 +6,7 @@
 // ConfigMap resource which stores the set of inventory
 // (object metadata).
 
-package prune
+package inventory
```
mortent, line 9: I think we should keep this in a package separate from both prune and inventory. This is just one of the (possibly many) implementations of the inventory.

pkg/apply/applier.go
```diff
@@ -46,8 +47,8 @@ func NewApplier(factory util.Factory, ioStreams genericclioptions.IOStreams) *Ap
 	}
 	a.previousInventoriesFunc = a.PruneOptions.GetPreviousInventoryObjects
 	a.infoHelperFactoryFunc = a.infoHelperFactory
-	a.InventoryFactoryFunc = prune.WrapInventoryObj
-	a.PruneOptions.InventoryFactoryFunc = prune.WrapInventoryObj
+	a.InventoryFactoryFunc = inventory.WrapInventoryObj
+	a.PruneOptions.InventoryFactoryFunc = inventory.WrapInventoryObj
```
mortent, line 51: Is there any way we can avoid doing this? Not necessarily in this PR. But it makes it impossible to hide PruneOptions behind an interface.

pkg/inventory/inventory.go
```diff
@@ -9,7 +9,7 @@
 // objects, we can correctly prune and teardown sets
 // of resources.
 
-package prune
+package inventory
```
mortent, line 12: Shouldn't we also move `GetPreviousInventoryObjects` into the inventory package? It doesn't seem like it belongs with prune.

seans3: Exactly. In future PR, there will be `InventoryClient` interface, which will implement this (unit tests are easier too).

pkg/apply/applier.go
seans3: Yes -- in a future PR.

pkg/inventory/inventorycm.go
seans3: I think this properly belongs in the `inventory` package, but I'm not wedded to it. Please let me know what other package this could be put in.

mortent: I just think it is useful to keep the implementation separate from the interface. This is just the implementation of the `Inventory` interface using a configmap, but we will have an implementation with ResourceGroup that must live in a package in another repo.

PR #190, opened 2020-06-13
Creates InventoryClient to refactor prune

seans3:
* Creates new `InventoryClient` interface to abstract retrieving inventory from cluster.
* Implements `GetPreviousInventoryObjects` and `GetStoredObjRefs` methods in new interface.
* Creates `ClusterInventoryClient` concrete implementation and `FakeInventoryClient`, which implement `InventoryClient`.
* Prune and applier use `InventoryClient` instead of `prevInventoryFunc`

### DIFF
--- a/pkg/inventory/inventory-client.go
+++ b/pkg/inventory/inventory-client.go
@@ -1,0 +1,184 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package inventory
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/cli-runtime/pkg/resource"
+	"k8s.io/klog"
+	"k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/validation"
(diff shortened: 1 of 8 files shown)


pkg/inventory/inventory-client.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+// InventoryClient expresses an interface for retrieving Inventory
+// objects in the clus
+type InventoryClient interface {
+	// GetPreviousInventoryObjects returns the inventory objects (as *resource.Info)
+	// that are currently stored in the cluster, or an error if one occurred. Uses
+	// information from the current (recently) created inventory object to find
+	// the previously applied inventory object in the cluster.
+	GetPreviousInventoryObjects(currentInv *resource.Info) ([]*resource.Info, error)
```
mortent, line 25: Longer term I don't think we should have a resource.Info as the parameter here, but rather something like an `InventoryID` that just captures the info we need to look up the previous inventories.

mortent:
/hold
Can we squash the commits?

seans3:
> /hold
> Can we squash the commits?

Squashed

PR #193, opened 2020-06-23
Allow colons in object names

chrigl:
In most objects, kubernetes does not allow colons, but especially in
Roles and Bindings they are widely used. On the other hand, colons are
not allowed as keys of ConfigMaps. When we try to write e.g. a
ClusterRole named `psp:rook`, `ObjectMetadata.String` results in a name
like `_psp:rook_rbac.authorization.k8s.io_ClusterRole`, which
consequently is refused by the apiserver.

```
Fatal error: error when creating "generated": ConfigMap "inventory-b5237659" is invalid: data[_psp:rook_rbac.authorization.k8s.io_ClusterRole]: Invalid value: "_psp:rook_rbac.authorization.k8s.io_ClusterRole": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')
```

The corresponding object:
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: 'psp:rook'
rules:
- resources:
  - podsecuritypolicies
  apiGroups:
  - policy
  resourceNames:
  - rook-privileged
  verbs:
  - use
```

This PR translates the colon into a underscore, so that the result
of `ObjectMetadata.String` is
`_psp_rook_rbac.authorization.k8s.io_ClusterRole`.

### DIFF
--- a/pkg/object/objmetadata.go
+++ b/pkg/object/objmetadata.go
@@ -91,7 +91,7 @@ func (o *ObjMetadata) Equals(other *ObjMetadata) bool {
 func (o *ObjMetadata) String() string {
 	return fmt.Sprintf("%s%s%s%s%s%s%s",
 		o.Namespace, fieldSeparator,
-		o.Name, fieldSeparator,
+		strings.ReplaceAll(o.Name, ":", "_"), fieldSeparator,
 		o.GroupKind.Group, fieldSeparator,
 		o.GroupKind.Kind)
 }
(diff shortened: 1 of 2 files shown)


seans3:
/hold

My worry is that this will break our field parsing code. I'll get back shortly.

chrigl:
Thanks @seans3! Indeed it will break when parsing back. There is no easy way by just replacing a character by another valid one, because every valid one could have been used by the user. The only thing that comes to my mind would be to base64 encode the keys (while maintaining current inverse parse). But this would IMHO make the the inventory non human readable.

However, I will come up with a PR, so that such a change will fail the test.

PR #194, opened 2020-06-24
Updates prune algorithm to use one inventory object

seans3:
* Updates prune algorithm to use only one inventory object (instead of multiple). We now modify the current inventory object instead of creating a new one for each apply.
* Backwards compatible previous algorithm as follows: 1) Uses the existing inventory object as the one to modify, 2) Merges multiple inventory objects into one in the (rare) case that there are multiple inventory objects.
* Updates the inventory object in two locations: the very beginning of the apply algorithm, and the very end of the prune operation.
* The first inventory update modifies the inventory to store the union of the previous apply and the current apply. The last inventory update modifies the inventory to store only the object references for the current apply.
* Errors out of the apply algorithm if unable to perform the initial inventory union update.
* This algorithm is tolerant to failures in the middle of the algorithm as long as the first inventory union update succeeds.
* Adds `Merge`, `Replace`, `Init`, and `DeleteInventoryObj` methods within the `InventoryClient` interface to update the inventory object.
* Updates e2e tests to remove lines about inventory object updates.

### DIFF
--- a/examples/alphaTestExamples/MultipleServices.md
+++ b/examples/alphaTestExamples/MultipleServices.md
@@ -59,6 +59,7 @@ the namespace and inventory id used by apply to create inventory objects.
 ```
 kapply init $BASE/mysql > $OUTPUT/status
 expectedOutputLine "namespace: default is used for inventory object"
+
 kapply init $BASE/wordpress > $OUTPUT/status
 expectedOutputLine "namespace: default is used for inventory object"
 ```
@@ -79,8 +80,6 @@ expectedOutputLine "deployment.apps/mysql is Current: Deployment is available. R
 
 expectedOutputLine "secret/mysql-pass is Current: Resource is always ready"
 
-expectedOutputLine "configmap/inventory-57005c71 is Current: Resource is always ready"
-
 expectedOutputLine "service/mysql is Current: Service is ready"
 
 # Verify that we have the mysql resources in the cluster.
@@ -96,8 +95,6 @@ And the apply the wordpress service
 <!-- @RunWordpress @testE2EAgainstLatestRelease -->
 ```
 kapply apply $BASE/wordpress --reconcile-timeout=120s > $OUTPUT/status;
-
-expectedOutputLine "configmap/inventory-2fbd5b91 is Current: Resource is always ready"
 
 expectedOutputLine "service/wordpress is Current: Service is ready"
 
@@ -116,8 +113,6 @@ kapply destroy $BASE/wordpress > $OUTPUT/status;
 expectedOutputLine "service/wordpress deleted"
 
 expectedOutputLine "deployment.apps/wordpress deleted"
-
-expectedOutputLine "configmap/inventory-2fbd5b91 deleted"
 
 # Verify that we still have the mysql resources in the cluster.
 kubectl get all --no-headers --selector=app=mysql | wc -l | xargs > $OUTPUT/status
--- a/pkg/apply/applier.go
+++ b/pkg/apply/applier.go
@@ -157,42 +157,38 @@ func (a *Applier) infoHelperFactory() info.InfoHelper {
 	return info.NewInfoHelper(a.factory, a.ApplyOptions.Namespace)
 }
 
-// prepareObjects handles ordering of resources and sets up the inventory object
-// based on the provided inventory object template.
+// prepareObjects merges the currently applied objects into the
+// set of stored objects in the cluster inventory. In the process, it
+// calculates the set of objects to be pruned (pruneIds), and orders the
+// resources for the subsequent apply. Returns the sorted resources to
+// apply as well as the objects for the prune, or an error if one occurred.
 func (a *Applier) prepareObjects(infos []*resource.Info) (*ResourceObjects, error) {
-	resources, invs := splitInfos(infos)
-
-	if len(invs) == 0 {
-		return nil, inventory.NoInventoryObjError{}
-	}
-	if len(invs) > 1 {
-		return nil, inventory.MultipleInventoryObjError{
-			InventoryObjectTemplates: invs,
-		}
-	}
-
-	inv := a.InventoryFactoryFunc(invs[0])
-	inventoryObject, err := inventory.CreateInventoryObj(inv, resources)
+	localInv, localInfos, err := inventory.SplitInfos(infos)
 	if err != nil {
 		return nil, err
 	}
-
-	// Fetch all previous inventories.
-	previousInventories, err := a.invClient.GetPreviousInventoryObjects(inventoryObject)
+	currentObjs, err := object.InfosToObjMetas(localInfos)
 	if err != nil {
 		return nil, err
 	}
-
-	sort.Sort(ordering.SortableInfos(resources))
-
-	if !validateNamespace(resources) {
+	// returns the objects (pruneIds) to prune after apply. The prune
+	// algorithm requires stopping if the merge is not successful. Otherwise,
+	// the stored objects in inventory could become inconsistent.
+	pruneIds, err := a.invClient.Merge(localInv, currentObjs)
+	if err != nil {
+		return nil, err
+	}
+	// Sort order for applied resources.
+	sort.Sort(ordering.SortableInfos(localInfos))
+	// TODO(seans3): Remove this single namespace requirement once we've
+	// implemented multi-namespace apply.
+	if !validateNamespace(localInfos) {
 		return nil, fmt.Errorf("objects have differing namespaces")
 	}
-
 	return &ResourceObjects{
-		CurrentInventory:    inventoryObject,
-		PreviousInventories: previousInventories,
-		Resources:           resources,
+		LocalInv:  localInv,
+		Resources: localInfos,
+		PruneIds:  pruneIds,
 	}, nil
 }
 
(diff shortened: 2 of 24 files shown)


mortent:
@seans3 So does this change completely remove the possibility of using multiple inventory objects like we are doing today? We need this to be backwards compatible so users who already have existing inventory objects in their clusters can keep using them. We want to deprecate and remove this eventually, but we need to give users time to transition.

pwittrock:
Does this address the issue of the keys not supporting “:”?

pkg/inventory/inventory-client.go
```diff
@@ -79,79 +71,176 @@ func NewInventoryClient(factory util.Factory) (*ClusterInventoryClient, error) {
...
 	if err != nil {
-		return nil, err
+		return pruneIds, err
 	}
-	return UnionPastObjs(prevInventories)
+	if clusterInv == nil {
+		// Initial create, we must add a unique suffix to the name. Otherwise,
+		// we'll get collisions. The other way to implement this is a unique
+		// suffix appended to name on "init" command.
+		objsHashStr, err := object.Hash(objs)
```
mortent, line 108: Why would we need to add a suffix in this case? We no longer create more than one inventory object, so it seems like it should be ok to just just the name and namespace from the inventory template directly.

seans3: Removed here. I will add the suffix in a separate PR in the init to generate a unique inventory name that does not collide with other inventory objects.

pkg/apply/prune/prune.go
```diff
@@ -72,6 +71,10 @@ func (po *PruneOptions) Initialize(factory util.Factory, invClient inventory.Inv
 	if err != nil {
 		return err
 	}
+	po.InvClient, err = inventory.NewInventoryClient(factory, dryRun)
```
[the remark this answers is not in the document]
seans3: Fixed.

pkg/inventory/inventory-client.go
```diff
@@ -5,71 +5,63 @@ package inventory
...
-	retrievedInventoryObjects bool
+	builderFunc func() *resource.Builder
+	mapper      meta.RESTMapper
+	validator   validation.Schema
+	clientFunc  func(*meta.RESTMapping) (resource.RESTClient, error)
+	dryRun      bool
+	// localInv is the locally read inventory object. It contains the
+	// necessary fields to retrieve the cluster inventory object, including:
+	// namespace, label, GVK.
+	localInv *resource.Info
```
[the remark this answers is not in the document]
seans3: Fixed: removed `localInv` and `clusterInv` state in the `InventoryClient`. Updated tests to use `FakeBuilder`

pkg/apply/applier.go
```diff
@@ -82,7 +82,7 @@ type Applier struct {
 // Initialize sets up the Applier for actually doing an apply against
 // a cluster. This involves validating command line inputs and configuring
 // clients for communicating with the cluster.
-func (a *Applier) Initialize(cmd *cobra.Command) error {
+func (a *Applier) Initialize(cmd *cobra.Command, dryRun bool) error {
```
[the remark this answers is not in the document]
seans3: Cleaned-up the `dryRun`.

pkg/inventory/inventory-client.go
```diff
@@ -165,20 +226,147 @@ func (cic *ClusterInventoryClient) retrievePreviousInventoryObjects(current *obj
...
+		retainedObjs = object.Union(retainedObjs, mergeObjs)
+	}
+	if err := wrapRetained.Store(retainedObjs); err != nil {
+		return nil, err
+	}
+	retainInfo, err := wrapRetained.GetObject()
+	if err != nil {
+		return nil, err
+	}
+	if err := cic.applyInventoryObj(retainInfo); err != nil {
```
phanimarupaka, line 282: What if this step fails ? We have deleted all the past inventory objects in the cluster at this point.

pkg/inventory/inventory-client.go
```diff
@@ -79,79 +62,157 @@ func NewInventoryClient(factory util.Factory) (*ClusterInventoryClient, error) {
...
+// getClusterInventoryObj returns a pointer to the cluster inventory object, or
+// an error if one occurred. Returns the cached cluster inventory object if it
+// has been previously retrieved. Uses the ResourceBuilder to retrieve the
+// inventory object in the cluster, using the namespace, group resource, and
+// inventory label. Merges multiple inventory objects into one if it retrieves
+// more than one (this should be very rare).
+//
+// TODO(seans3): Remove the special case code to merge multiple cluster inventory
+// objects once we've determined that this case is no longer possible.
+func (cic *ClusterInventoryClient) getClusterInventoryInfo(localInv *resource.Info) (*resource.Info, error) {
```
phanimarupaka, line 194: The method name says get but it is deleting past inventory objects silently in the background. Though it is a one time step, we need to think of the failure scenarios. 

Nit: get prefix is not recommended as per go readability. Instead it can just be  ClusterInventoryInfo.

pkg/inventory/inventory-client.go
```diff
@@ -165,20 +226,147 @@ func (cic *ClusterInventoryClient) retrievePreviousInventoryObjects(current *obj
...
+}
+
+// mergeClusterInventory merges the inventory of multiple inventory objects
+// into one inventory object, and applies it. Deletes the remaining unnecessary
+// inventory objects. There should be only one inventory object stored in the
+// cluster after this function. This special case should be very rare.
+//
+// TODO(seans3): Remove this code once we're certain no customers have multiple
+// inventory objects in their clusters.
+func (cic *ClusterInventoryClient) mergeClusterInventory(invInfos []*resource.Info) (*resource.Info, error) {
```
phanimarupaka, line 248: How about doing this as part of kpt fix command so that we are transparent to the users about this migration? Agree that this is easier approach but just a thought.

phanimarupaka:
@seans3 What are the limitations of using this approach ? What is the maximum number of resources that can be tracked by single inventory object in cluster? We need to make this clear in our documentation. It might be trivial now as there are not many power users, but eventually users might run into these limitations.

pkg/inventory/inventory-client.go
```diff
@@ -79,79 +62,157 @@ func NewInventoryClient(factory util.Factory) (*ClusterInventoryClient, error) {
...
-func (cic *ClusterInventoryClient) GetStoredObjRefs(currentInv *resource.Info) ([]object.ObjMetadata, error) {
-	prevInventories, err := cic.GetPreviousInventoryObjects(currentInv)
+// Merge stores the union of the passed objects with the objects currently
+// stored in the cluster inventory object. Retrieves and caches the cluster
+// inventory object. Returns the set differrence of the cluster inventory
+// objects and the currently applied objects. This is the set of objects
+// to prune. Creates the initial cluster inventory object storing the passed
+// objects if an inventory object does not exist. Returns an error if one
+// occurred.
+func (cic *ClusterInventoryClient) Merge(localInv *resource.Info, objs []object.ObjMetadata) ([]object.ObjMetadata, error) {
```
phanimarupaka, line 83: Merge is also calculating diff and returning pruneIds, maybe a better name which reflects entire behavior of method would be great.

pkg/inventory/inventory-client.go
```diff
@@ -79,79 +62,157 @@ func NewInventoryClient(factory util.Factory) (*ClusterInventoryClient, error) {
...
-		past, err := infoToObjMetadata(pastInfo)
+	if !cic.dryRun {
+		clusterInv, err = wrappedInv.GetObject()
 		if err != nil {
-			return nil, err
+			return err
 		}
-		if !current.Equals(past) {
-			pastInventoryInfos = append(pastInventoryInfos, pastInfo)
+		klog.V(4).Infof("replace cluster inventory: %s/%s", clusterInv.Namespace, clusterInv.Name)
```
phanimarupaka, line 160: So the new single inventory object will have the name namespace/name? What if we extend the functionality to allow objects of different namespaces in an inventory object ? The naming of the inventory object/cm in the cluster is critical.

examples/alphaTestExamples/MultipleServices.md
```diff
@@ -58,7 +58,9 @@ the namespace and inventory id used by apply to create inventory objects.
 <!-- @createInventoryTemplate @testE2EAgainstLatestRelease-->
 ```
 kapply init $BASE/mysql
+sed -i 's/name: inventory/name: inventory-mysql/g' $BASE/mysql/inventory-template.yaml
```
phanimarupaka, line 61: We need to have an e2e test where the initial state has multiple inventory objects in the cluster, and apply consolidated them to single object and preview let it as is.

pkg/inventory/inventory-client.go
seans3: Good catch. The apply must happen before the deletes. Changed and added unit tests.

pkg/inventory/inventory-client.go
seans3: Let's keep this as is. We'd have to revert numerous function signatures to return a slice of inventory infos, instead of assuming just one inventory info is returned. This would be very awkward.

pkg/inventory/inventory-client.go
seans3: This is a large, long PR. Let' keep this as is for now.

pkg/apply/task/apply_task.go
```diff
@@ -109,10 +109,18 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
 		// applied.
 		//TODO: This isn't really needed if we are doing dry-run.
 		for _, obj := range objects {
-			id := object.InfoToObjMeta(obj)
-			acc, _ := meta.Accessor(obj.Object)
-			gen := acc.GetGeneration()
-			taskContext.ResourceApplied(id, gen)
+			id, err := object.InfoToObjMeta(obj)
+			if err != nil {
+				continue
```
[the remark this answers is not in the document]
seans3: It was being ignored before.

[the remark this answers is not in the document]
seans3: So this won't be a change. But I could add a TODO.

pkg/apply/applier.go
```diff
@@ -157,89 +157,78 @@ func (a *Applier) infoHelperFactory() info.InfoHelper {
...
 // InfosForApply returns the infos representation for all the resources
 // that should be applied, including the inventory object. The
 // resources will be in sorted order.
 func (r *ResourceObjects) InfosForApply() []*resource.Info {
-	return append([]*resource.Info{r.CurrentInventory}, r.Resources...)
+	return r.Resources
+}
+
+func (r *ResourceObjects) InfosForPrune() []*resource.Info {
+	return append([]*resource.Info{r.LocalInv}, r.Resources...)
```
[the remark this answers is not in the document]
seans3: Correct.

Issue #197, opened 2020-07-06
Server-side dry-run support for live preview


fsommar:
Hi 👋

We're using `kpt live preview` for PR feedback purposes. We recently ran into a resource-naming issue where the name was not matching the expected Kubernetes regex for metadata.name. This would've been caught earlier if preview supported server-side dry-run.

It'd be great if the server-side dry-run for preview could echo the semantics of `kubectl apply --dry-run=server`.

My guess is that implementing something like this would require 1) update of kubectl dependency and 2) adding a server-side dry-run to preview that delegates to the server-side dry-run strategy in the updated kubectl dependency. Does that sound about right?

I'd be willing to give to give this a try, but I'm not familiar with how to update dependencies (i.e. using `go get` locally adds a ton of new transitive dependencies).

seans3:
Hi @fsommar . Server-side dry-run requires an APIServer version >= 1.16 (it requires server-side apply which is turned on by default in 1.16). Currently, GKE is in the range of 1.14 - 1.16, with 1.17 as the rapid channel. So we're not currently in a position to use server-side apply/dry-run, since some clusters will not understand the server-side apply. Once we can rely on server-side apply, we will change `kpt live apply` and `kpt live preview` to use server-side apply. I'm going to close this issue shortly.

seans3:
...as far as the change needed (once we intend to make it), the import of `kubectl` does not need to change (since it is about 1.18). But we'll need to update the `ApplyOptions` to fill in the server-side flags.

fsommar:
Ok, that makes sense. I saw that there were some server-side-apply-related flags, and figured that maybe at least the server-side dry-run could be exposed as a hidden flag? I created a PR for this in case that'd be an acceptable short-term fix.

fsommar:
I'll close this and then revisit it in the future when GKE is on more recent versions of Kubernetes.

/close

PR #200, opened 2020-07-07
Add hidden server-dry-run flag

fsommar:
This flag would need to be updated when the kubectl library is updated to `--dry-run=server`, and as server-side apply is made into a first-class citizen in kapply.

I think the logic for the DryRun option is correct, since we're fine to send anything to the server as long as server-side dry-run is enabled.

Closes https://github.com/kubernetes-sigs/cli-utils/issues/197.

### DIFF
--- a/cmd/preview/cmdpreview.go
+++ b/cmd/preview/cmdpreview.go
@@ -101,7 +101,7 @@ func NewCmdPreview(f cmdutil.Factory, ioStreams genericclioptions.IOStreams) *co
 	_ = cmd.Flags().MarkHidden("validate")
 	// Server-side flags are hidden for now.
 	cmdutil.AddServerSideApplyFlags(cmd)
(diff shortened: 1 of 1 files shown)


seans3:
/ok-to-test
/hold

Even with the server-side flag hidden, this is a dangerous approach. If an APIServer does not understand server-side dry-run, it could actually apply a resource instead of dry-run it. This is bad bug, and it has already happened before in `kubectl`. I'm allowing testing, but I suggest we drop this PR. I suggest we wait until we move all `kpt live` commands to server-side versions before attempting to enable any server-side functionality.

fsommar:
Ok, that's unfortunate - I'm closing this then. Do you know a rough timeline of when you feel confident with GKE version 1.16+ adoption to go ahead with server-side-apply-support?

PR #201, opened 2020-07-09
Add LastTransitionTime into kstatus/status.Compute return value

haiyanmeng:


mortent:
@haiyanmeng Can you add a high-level explanation about how we want to compute `lastTransitionTime` in the general case? When I'm looking at this now, I realize that since kstatus only looks at the state of a resource at a certain point in time, it doesn't seem obvious how we can tell when one of the kstatus conditions does a transition.

haiyanmeng:
> @haiyanmeng Can you add a high-level explanation about how we want to compute `lastTransitionTime` in the general case? When I'm looking at this now, I realize that since kstatus only looks at the state of a resource at a certain point in time, it doesn't seem obvious how we can tell when one of the kstatus conditions does a transition.

If kstatus computes a condition based on an existing condition in an `unstructured.Unstructured` object, this PR sets `lastTransitionTime` using the `lastTransitionTime` field in `unstructured.Unstructured`;
Otherwise, this PR sets `lastTransitionTime` to current time.

mortent:
> If kstatus computes a condition based on an existing condition in an `unstructured.Unstructured` object, this PR sets `lastTransitionTime` using the `lastTransitionTime` field in `unstructured.Unstructured`;
> Otherwise, this PR sets `lastTransitionTime` to current time.

So the `lastTransitionTime` in `unstructured.Unstructured` means taking it from the condition I assume?

My concern here is that this doesn't really seem to accurately reflect when the condition actually do change. From what I can tell, it seems like if the resource doesn't have an existing condition, we will just always set the `lastTransitionTime` to current time. But it seems like we in most situations are able to get the time from an existing condition.

pkg/kstatus/status/core.go
```diff
@@ -262,8 +264,14 @@ func replicasetConditions(u *unstructured.Unstructured) (*Result, error) {
 	for _, c := range objc.Status.Conditions {
 		// https://github.com/kubernetes/kubernetes/blob/a3ccea9d8743f2ff82e41b6c2af6dc2c41dc7b10/pkg/controller/replicaset/replica_set_utils.go
 		if c.Type == "ReplicaFailure" && c.Status == corev1.ConditionTrue {
+			reason := "ReplicaFailure"
 			message := "Replica Failure condition. Check Pods"
-			return newInProgressStatus("ReplicaFailure", message), nil
```
mortent, line 266: Do we ever call this function without the `lastTransitionTime`? It seems like it should always be set, so should we change the function signature to accept a timestamp?

haiyanmeng: `newInProgressStatus` and other `newXXXStatus` should set the `LastTransitionTime` to current time.
Overall, `newXXXStatus` are called much more frequently than returning a `Result` whose `LastTransitionTime` is from  an existing condition in an `unstructured.Unstructured` object.

haiyanmeng:
> So the `lastTransitionTime` in `unstructured.Unstructured` means taking it from the condition I assume?

Yes.

> From what I can tell, it seems like if the resource doesn't have an existing condition, we will just always set the `lastTransitionTime` to current time. But it seems like we in most situations are able to get the time from an existing condition.

In this PR, `lastTransitionTime` is only set to current time when `NewXXXStatus` (e.g. `NewInProgressStatus`) is called. 
If kstatus computes a condition based on an existing condition in an unstructured.Unstructured object, this PR sets lastTransitionTime using the `lastTransitionTime` field in unstructured.Unstructured, even if the `lastTransitionTime` field is empty.

> My concern here is that this doesn't really seem to accurately reflect when the condition actually do change. 

What is your suggestions on how we should set `lastTransitionTime`?

mortent:
> What is your suggestions on how we should set `lastTransitionTime`?

So I'm wondering if handling the `lastTransitionTime` would be better handled by libraries using kstatus? So if the goal is to use the library in a controller, presumably the kstatus conditions will be added to the status object of resources. So it seems the controller could make better decisions about when the condition changed by comparing the result from the kstatus library with the previous condition on the resource. Would this be an option?

haiyanmeng:
> So I'm wondering if handling the `lastTransitionTime` would be better handled by libraries using kstatus? So if the goal is to use the library in a controller, presumably the kstatus conditions will be added to the status object of resources. So it seems the controller could make better decisions about when the condition changed by comparing the result from the kstatus library with the previous condition on the resource. Would this be an option?

I think kstatus is a better place to handle `lastTransitionTime` for two reasons:
1) avoiding code duplication. `resourcegroup` may not be the only project using `kstatus`;
2) to be consistent with https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1623-standardize-conditions.

PR #204, opened 2020-07-15
Added parsing to objmetadata_test.TestCreateObjMetadata

chrigl:
This commit extends TestCreateObjMetadata. The test function does not
just verify CreateObjMetadata, but also uses ParseObjMetadata to verify
back and forth between both functions. I left specific testing of
ParseObjMetadata in place, because there might be edge cases, that
should be tested on its own.

I'm aware, that the TestCreateObjMetadata now tests two things, but I
think it makes life of maintainers and new contributors far easier.
E.g. the change I introduced in #193 would have been catched by tests,
rather than someone with a mental model of all of the code.

### DIFF
--- a/pkg/object/objmetadata_test.go
+++ b/pkg/object/objmetadata_test.go
@@ -4,6 +4,7 @@
 package object
 
 import (
+	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


seans3:
/approve
/lgtm

Thanks for the extra tests.

PR #207, opened 2020-07-17
Prune/destroy resources in reverse order from apply

mortent:
Similar to how we use a set of rules to determine the order in which resources are applied, this uses the same rules to prune/destroy resources in reverse order. This doesn't take care of all potential issues around ordering, but will make sure that things like RBAC rules are removed after namespaced resources, which is a common issue when deleting resources in a namspace.

Fixes: https://github.com/GoogleContainerTools/kpt/issues/745

@seans3 @phanimarupaka


Aitozi:
Is this can be used to delete in reverse order used with `kubectl delete -f filename` ?

PR #209, opened 2020-07-31
Adds e2e test for basic pruning functionality

seans3:
* Adds end-to-end test for basic pruning functionality

### DIFF
--- a/examples/alphaTestExamples/pruneBasic.md
+++ b/examples/alphaTestExamples/pruneBasic.md
@@ -1,0 +1,175 @@
+[kind]: https://github.com/kubernetes-sigs/kind
+
+# Demo: Basic Prune
+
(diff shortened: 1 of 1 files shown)


mortent:
So looks good, but can we add some more checks? Currently we are only verifying that we see the desired print statements, but we should also check that the state in the cluster is what we expect. We have had these kinds of bugs in the past.

seans3:
> So looks good, but can we add some more checks? Currently we are only verifying that we see the desired print statements, but we should also check that the state in the cluster is what we expect. We have had these kinds of bugs in the past.

Good Point. Added tests for resources in the cluster. Have a look.

mortent:
Looks good to me. Can we squash the commits?

seans3:
> Looks good to me. Can we squash the commits?

Squashed.

PR #222, opened 2020-08-18
Fix name of --inventory-namespace flag in init error message

Niksko:
Minor fix, this flag is incorrectly named


mortent:
@Niksko Thanks for the contribution!
/lgtm
/approve

PR #225, opened 2020-08-26
Handle the --namespace flag

mortent:
The `--namespace` flag is currently not used by any of the commands, but it is available as a flag. This PR updates the commands to actually use both the namespace provided in the context as well as the namespace provided with the flag. It behaves as follows:

For `kapply init`:
 * If the user has not provided a namespace for the inventory with the `--inventory-namespace` flag and none of the resources in the package has the namespace set, it will fall back to use the namespace provided with the `--namespace` flag and then to the namespace provided in the context.

For `kapply apply/preview`:
 * If a resource does not have the namespace set and the resource has namespace scope, it will fall back to use the namespace provided with the `--namespace` flag and then to the namespace provided in the context. The inventory template is ignored in this process, so this will never affect the namespace for the inventory
 * If the namespace is provided with the `--namespace` flag, it is considered an error if any of the resources (except the inventory) is in a different namespace. This is aligned with the behavior in kubectl. 

Fixes: https://github.com/GoogleContainerTools/kpt/issues/975, https://github.com/GoogleContainerTools/kpt/issues/826

### DIFF
--- a/cmd/apply/cmdapply.go
+++ b/cmd/apply/cmdapply.go
@@ -117,10 +117,21 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Fetch the namespace from the configloader. The source of this
+	// either the namespace flag or the context. If the namespace is provided
+	// with the flag, enforceNamespace will be true. In this case, it is
+	// an error if any of the resources in the package has a different
+	// namespace set.
+	namespace, enforceNamespace, err := r.provider.Factory().ToRawKubeConfigLoader().Namespace()
+	if err != nil {
+		return err
+	}
+
 	var reader manifestreader.ManifestReader
 	readerOptions := manifestreader.ReaderOptions{
-		Factory:   r.provider.Factory(),
-		Namespace: metav1.NamespaceDefault,
+		Factory:          r.provider.Factory(),
+		Namespace:        namespace,
+		EnforceNamespace: enforceNamespace,
 	}
 	if len(args) == 0 {
 		reader = &manifestreader.StreamManifestReader{
--- a/cmd/preview/cmdpreview.go
+++ b/cmd/preview/cmdpreview.go
@@ -7,7 +7,6 @@ import (
 	"context"
 
 	"github.com/spf13/cobra"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/util/i18n"
@@ -117,10 +116,21 @@ func (r *PreviewRunner) RunE(cmd *cobra.Command, args []string) error {
 			return err
 		}
 
+		// Fetch the namespace from the configloader. The source of this
+		// either the namespace flag or the context. If the namespace is provided
+		// with the flag, enforceNamespace will be true. In this case, it is
+		// an error if any of the resources in the package has a different
+		// namespace set.
+		namespace, enforceNamespace, err := r.provider.Factory().ToRawKubeConfigLoader().Namespace()
+		if err != nil {
+			return err
+		}
+
 		var reader manifestreader.ManifestReader
 		readerOptions := manifestreader.ReaderOptions{
-			Factory:   r.provider.Factory(),
-			Namespace: metav1.NamespaceDefault,
+			Factory:          r.provider.Factory(),
+			Namespace:        namespace,
+			EnforceNamespace: enforceNamespace,
 		}
 		if len(args) == 0 {
 			reader = &manifestreader.StreamManifestReader{
--- a/pkg/config/initoptions.go
+++ b/pkg/config/initoptions.go
@@ -11,10 +11,9 @@ import (
 	"strings"
 	"time"
 
-	"github.com/go-errors/errors"
 	"github.com/google/uuid"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 )
@@ -62,6 +61,8 @@ metadata:
 // InitOptions contains the fields necessary to generate a
 // inventory object template ConfigMap.
 type InitOptions struct {
+	factory cmdutil.Factory
+
 	ioStreams genericclioptions.IOStreams
 	// Package directory argument; must be valid directory.
 	Dir string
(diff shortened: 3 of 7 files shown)


wpwoodjr:
>If the user has not provided a namespace for the inventory with the --inventory-namespace flag and none of the resources in the package has the namespace set, it will fall back to use the namespace provided with the --namespace flag and then to the namespace provided in the context.

What happens if the user has not provided a namespace for the inventory with the --inventory-namespace flag and a resource **does** have the namespace set?

pkg/config/initoptions.go
```diff
@@ -140,21 +147,21 @@ func isDirectory(path string) bool {
 // config files. Assumes all namespaced resources are in the
 // same namespace. Returns the default namespace if none of the
 // config files has a namespace.
-func calcPackageNamespace(packageDir string) (string, error) {
+func calcPackageNamespace(packageDir, defaultNamespace string) (string, error) {
```
seans3, line 150: We probably want to add this new parameter (and what it means) to the function documentation.

cmd/apply/cmdapply.go
```diff
@@ -114,10 +114,16 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+
 	var reader manifestreader.ManifestReader
 	readerOptions := manifestreader.ReaderOptions{
-		Factory:   r.factory,
-		Namespace: metav1.NamespaceDefault,
+		Factory:          r.factory,
+		Namespace:        namespace,
+		EnforceNamespace: enforceNamespace,
```
seans3, line 126: Are we using the `EnforceNamespace` field anywhere?

cmd/preview/cmdpreview.go
```diff
@@ -114,10 +113,16 @@ func (r *PreviewRunner) RunE(cmd *cobra.Command, args []string) error {
 			return err
 		}
 
+		namespace, enforceNamespace, err := r.factory.ToRawKubeConfigLoader().Namespace()
```
seans3, line 116: Maybe a comment here for those not familiar with namespace within the kubeconfig file.

cmd/apply/cmdapply.go
```diff
@@ -114,10 +114,16 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	namespace, enforceNamespace, err := r.factory.ToRawKubeConfigLoader().Namespace()
```
seans3, line 117: Same about a comment here.

wpwoodjr:
I'm not clear why a separate `--inventory-namespace` flag is needed.

For `kpt apply init` why can't the `--namespace` flag set the inventory namespace?  That's what I expected initially until I saw the `--inventory-namespace` flag.

mortent:
@wpwoodjr I did consider using the `--namespace` flag directly, but it is a flag that is used across all `kpt live` commands as well as in kubectl. The way it is handled now, it should behave in a way that is consistent with how it is used elsewhere. The way the namespace flag interacts with the namespace set in the context also makes it hard to use it instead of the `--inventory-namespace` flag.

So with this change, the `--inventory-template` flag is a way to force the namespace to a specific value, while the `--namespace` flag and the namespace set on the context provides a way to change the default when none of the resources have the namespace set.

cmd/apply/cmdapply.go
mortent: The goal is to use it the same way here as it is used in kubectl. So the if the value of the namespace is coming from the context, the value of `EnforceNamespace` will be `false`, while if the namespace is set with the `--namespace` flag, the `EnforceNamespace` value will be `true`.
The value here is used in the `setNamespaces` function.

wpwoodjr:
> @wpwoodjr I did consider using the `--namespace` flag directly, but it is a flag that is used across all `kpt live` commands as well as in kubectl. The way it is handled now, it should behave in a way that is consistent with how it is used elsewhere. The way the namespace flag interacts with the namespace set in the context also makes it hard to use it instead of the `--inventory-namespace` flag.
> 
> So with this change, the `--inventory-template` flag is a way to force the namespace to a specific value, while the `--namespace` flag and the namespace set on the context provides a way to change the default when none of the resources have the namespace set.

I think its confusing.  Why can't it just ignore the namespace if set on other resources. `kpt live init .` could first use the `--namespace` flag, if that's not present fall back to the context, if that's not present fall back to `default` (or even an error)?

mortent:
So an alternative solution here is the following:
 * If the namespace is provided with the `--namespace` flag, that namespace will be used for the inventory no matter which namespaces are in the resources.
 * If the `--namespace` flag is not set and all resources have a namespace set and it is the same namespace, that will also be used for the inventory.
 * In all other cases, the namespace from the context will be used. This is always specified, and in most cases it will be `default`.

If we adopted this solution, we should also change the name of the `--inventory-name` flag to `--name` to be consistent. 

@wpwoodjr Do you think this approach is less confusing?

wpwoodjr:
@mortent yes I think it is less confusing, thanks!

seans3:
Some more context on the `--namespace` flag. As mentioned, it is used by all the `kpt live` commands that speak to an API Server (`kpt live apply`, `kpt live preview`). It is defined in the same place as all the `kubeconfig` flags that are used to communicate with an API Server. Here is the set of `kubeconfig` flags (including Namespace): `k8s.io/kubernetes/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go`
```
/ ConfigFlags composes the set of values necessary                                                                                                                                                           
// for obtaining a REST client config                                                                                                                                                                         
type ConfigFlags struct {
    CacheDir   *string
    KubeConfig *string

    // config flags                                                                                                                                                                                           
    ClusterName      *string
    AuthInfoName     *string
    Context          *string
    Namespace        *string
    APIServer        *string
    TLSServerName    *string
    Insecure         *bool
    CertFile         *string
    KeyFile          *string
    CAFile           *string
    BearerToken      *string
    Impersonate      *string
    ImpersonateGroup *[]string
    Username         *string
    Password         *string
    Timeout          *string

    clientConfig clientcmd.ClientConfig
    lock         sync.Mutex
    // If set to true, will use persistent client config and                                                                                                                                                  
    // propagate the config to the places that need it, rather than                                                                                                                                           
    // loading the config multiple times                                                                                                                                                                      
    usePersistentConfig bool
}
```

The reason the `--namespace` flag can be re-used (to specify the inventory namespace) in the `kpt live init` command because we do not contact the API server for this command. If the `kpt live init` command ever needs to interact with the cluster, then this `--namespace` flag reuse will not work.

seans3:
I'll LGTM when the merge conflict is resolved.

mortent:
@seans3 This is updated
/hold cancel

wpwoodjr:
> The reason the `--namespace` flag can be re-used (to specify the inventory namespace) in the `kpt live init` command because we do not contact the API server for this command. If the `kpt live init` command ever needs to interact with the cluster, then this `--namespace` flag reuse will not work.

Its hard to comment on a future, unspecified need; but why would the namespace need to be different if you were contacting the server in that instance?

seans3 (approved):
/lgtm

PR #227, opened 2020-08-27
New Provider wraps kubectl factory and inventory client

seans3:
* Creates new `Provider` interface specifically for `cli-utils` which wraps the `kubectl Factory` and the `InventoryClient`.
* Creates two concrete implementations of this new `Provider` interface: `InventoryProvider` and `FakeProvider`.
* Moves closer to dependency injection, since this `cli-utils Provider` is created at the beginning of the `cli-utils` commands.

### DIFF
--- a/cmd/apply/cmdapply.go
+++ b/cmd/apply/cmdapply.go
@@ -17,15 +17,18 @@ import (
 	"sigs.k8s.io/cli-utils/cmd/printers"
 	"sigs.k8s.io/cli-utils/pkg/apply"
 	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
 	"sigs.k8s.io/cli-utils/pkg/manifestreader"
+	"sigs.k8s.io/cli-utils/pkg/provider"
 	"sigs.k8s.io/kustomize/kyaml/setters2"
 )
 
 func GetApplyRunner(f cmdutil.Factory, ioStreams genericclioptions.IOStreams) *ApplyRunner {
+	provider := provider.NewProvider(f, inventory.WrapInventoryObj)
 	r := &ApplyRunner{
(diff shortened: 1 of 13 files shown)


pkg/factory/factory.go
```diff
@@ -0,0 +1,48 @@
...
+package factory
+
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/kubectl/pkg/cmd/util"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
+)
+
+// Factory is an interface which wraps the kubectl factory and
+// the inventory client.
```
monopole, line 13: In java guice land the name used for an uber-factory or factory-factory used in DI is _provider_.

Might want to call this package `provider`, and this struct `Provider` or `DepProvider`.
Avoids the collision with `sigs.k8s.io/cli-utils/pkg/util/factory`

pkg/factory/fake-factory.go
```diff
@@ -0,0 +1,35 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/kubectl/pkg/cmd/util"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
+	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+type FakeInventoryFactory struct {
+	factory util.Factory
+	objs    []object.ObjMetadata
+}
```
monopole, line 16: To help document the interface this implements, can we add 

```
var _ whatever.NameOfInterface = (*FakeInventoryFactory)(nil)
```

seans3: Done.

pkg/factory/factory.go
seans3: Renamed factory to provider for package and interface. Please have a look.

monopole:
/lgtm
thanks!

PR #233, opened 2020-09-08
Applies inventory namespace first if necessary

seans3:
* If the inventory namespace is one of the applied resources, apply this namespace before the inventory object.
* Fixes problem where inventory fails to apply because the namespace it lives in does not yet exist.
* Adds end-to-end test to validate the namespace is created before the inventory object is applied.
* Refactors common code to create the `resource.Helper` in `helperFromInfo()`

### DIFF
--- a/pkg/apply/applier.go
+++ b/pkg/apply/applier.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/go-errors/errors"
 	"github.com/spf13/cobra"
-	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
@@ -141,10 +140,11 @@ func (a *Applier) prepareObjects(infos []*resource.Info) (*ResourceObjects, erro
 		return nil, err
 	}
 
-	// Verify that the set of resources does not include the namespace
-	// in which we will place the inventory.
-	if err := validateNamespace(localInv, localInfos); err != nil {
-		return nil, err
+	// Ensures the namespace exists before applying the inventory object into it.
+	if invNamespace := inventoryNamespaceInSet(localInv, localInfos); invNamespace != nil {
+		if err = a.invClient.ApplyInventoryNamespace(invNamespace); err != nil {
+			return nil, err
+		}
 	}
 
 	currentObjs, err := object.InfosToObjMetas(localInfos)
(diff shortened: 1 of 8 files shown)


mortent:
Does this make sure that we never end up pruning the namespace if it contains the inventory object for the package?

seans3:
> Does this make sure that we never end up pruning the namespace if it contains the inventory object for the package?

Should that be a separate issue? It does not appear to deal with the ordering of the inventory namespace apply. Currently, we are pruning the inventory namespace if it is omitted.

mortent:
> > Does this make sure that we never end up pruning the namespace if it contains the inventory object for the package?
> 
> Should that be a separate issue? It does not appear to deal with the ordering of the inventory namespace apply. Currently, we are pruning the inventory namespace (and everything in it) if it is omitted in a subsequent apply.

I think we should address this here. This becomes an issue when we allow including the namespace for inventory object inside the package.

seans3:
> > > Does this make sure that we never end up pruning the namespace if it contains the inventory object for the package?
> > 
> > 
> > Should that be a separate issue? It does not appear to deal with the ordering of the inventory namespace apply. Currently, we are pruning the inventory namespace (and everything in it) if it is omitted in a subsequent apply.
> 
> I think we should address this here. This becomes an issue when we allow including the namespace for inventory object inside the package.

Addressed by skipping prune for inventory namespace (if not destroy command; destroy still deletes namespace). Added e2e test. Output now looks like:

```
$ kapply apply /home/sean/testdata/testdata-namespace-2
pod/pod-b unchanged
pod/pod-c configured
pod/pod-d created
3 resource(s) applied. 1 created, 1 unchanged, 1 configured
pod/pod-a pruned
namespace/test-namespace prune skipped
1 resource(s) pruned, 1 skipped
```

Let me know what you think.

pkg/apply/applier.go
```diff
@@ -342,22 +342,21 @@ func handleError(eventChannel chan event.Event, err error) {
...
-var coreV1Namespace = v1.SchemeGroupVersion.WithKind("Namespace")
-
-// validateNamespace returns true if all the objects in the passed
-// infos parameter have the same namespace; false otherwise. Ignores
-// cluster-scoped resources.
-func validateNamespace(inv *resource.Info, infos []*resource.Info) error {
+// inventoryNamespaceInSet returns the the namespace the passed inventory
+// object will be applied to, or nil if this namespace object does not exist
+// in the passed slice "infos" or the inventory object is cluster-scoped.
+func inventoryNamespaceInSet(inv *resource.Info, infos []*resource.Info) *resource.Info {
```
mortent, line 348: Not something we should address here, but we have at least two different types now that requires "special" handling and we end up looking for them among the slice of infos. I think we should probably create a new type that wraps the infos slice so we can put this functionality there.

pkg/inventory/inventory-client.go
```diff
@@ -36,6 +39,7 @@ type InventoryClient interface {
 	DeleteInventoryObj(inv *resource.Info) error
 	// SetDryRunStrategy sets the dry run strategy on whether this we actually mutate.
 	SetDryRunStrategy(drs common.DryRunStrategy)
+	ApplyInventoryNamespace(invNamespace *resource.Info) error
```
mortent, line 42: Is this really the right place for this function? It doesn't seem like applying a namespace has any dependency on the inventory object.

seans3: It's not a perfect fit, but I think this is the right place. This is related to the inventory object (since it is it's namespace). I don't think any other location would be especially better suited.

pkg/apply/applier.go
seans3: Ack.

PR #234, opened 2020-09-14
Refactor inventory template intialization

seans3:
* Adds `InitRunner` similar to the other commands for `init` command.
* Makes the inventory template dynamic, moving configmap-specific code into configmap location.


pkg/config/initoptions.go
```diff
@@ -250,7 +218,7 @@ func (i *InitOptions) fillInValues() string {
 	now := time.Now()
 	nowStr := now.Format("2006-01-02 15:04:05 MST")
 	randomSuffix := common.RandomStr(now.UTC().UnixNano())
-	manifestStr := configMapTemplate
+	manifestStr := i.Template
```
mortent, line 221: nit: I think we should just use the builtin golang templating engine here.

mortent (approved):
/lgtm

PR #235, opened 2020-09-14
Move ManifestReader creation into Provider

seans3:
* Moves `ManifrestReader` into Provider.
* Use `ManifestReader` to create `Infos` instead of `apply.GetObjects` within the `destroy` command like other commands.
* Several `Runner` commands now take `Provider` instead of `Factory`.
* Tested manually

### DIFF
--- a/cmd/destroy/cmddestroy.go
+++ b/cmd/destroy/cmddestroy.go
@@ -15,11 +15,11 @@ import (
 )
 
 // GetDestroyRunner creates and returns the DestroyRunner which stores the cobra command.
-func GetDestroyRunner(f cmdutil.Factory, ioStreams genericclioptions.IOStreams) *DestroyRunner {
-	provider := provider.NewProvider(f, inventory.WrapInventoryObj)
+func GetDestroyRunner(provider provider.Provider, ioStreams genericclioptions.IOStreams) *DestroyRunner {
 	r := &DestroyRunner{
 		Destroyer: apply.NewDestroyer(provider, ioStreams),
 		ioStreams: ioStreams,
+		provider:  provider,
 	}
 	cmd := &cobra.Command{
 		Use:                   "destroy (DIRECTORY | STDIN)",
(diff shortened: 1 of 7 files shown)


cmd/destroy/cmddestroy.go
```diff
@@ -66,7 +66,7 @@ func (r *DestroyRunner) RunE(cmd *cobra.Command, args []string) error {
 
 	// Run the destroyer. It will return a channel where we can receive updates
 	// to keep track of progress and any issues.
-	ch := r.Destroyer.Run()
+	ch := r.Destroyer.Run(cmd, args)
```
mortent, line 69: We should not pass cobra commands into the code that lives in `/pkg`

pkg/apply/destroyer.go
```diff
@@ -80,13 +80,14 @@ func (d *Destroyer) Initialize(cmd *cobra.Command, paths []string) error {
...
 // on progress and any errors are reported back on the event channel.
-func (d *Destroyer) Run() <-chan event.Event {
+func (d *Destroyer) Run(cmd *cobra.Command, args []string) <-chan event.Event {
 	ch := make(chan event.Event)
 
 	go func() {
 		defer close(ch)
 		d.invClient.SetDryRunStrategy(d.DryRunStrategy)
-		infos, err := d.ApplyOptions.GetObjects()
+		reader := d.provider.ManifestReader(cmd.InOrStdin(), args)
```
mortent, line 89: So this is all just to find the inventory information. We should look at how we can make this the responsibility of the client rather than the library, so in this case this should happen in the code that lives under `/cmd`. This should help us make it easier to make the apply logic available as a library.

seans3: I think I've fixed this. The `ManifestReader.Read()` is now in the `cmd` like the other commands. Have a look.

cmd/destroy/cmddestroy.go
seans3: I think I've fixed this. The ManifestReader.Read() is now in the cmd like the other commands. Have a look

seans3:
/retest 
...timeout

cmd/destroy/cmddestroy.go
```diff
@@ -64,9 +66,20 @@ func (r *DestroyRunner) RunE(cmd *cobra.Command, args []string) error {
...
+	}
+	inv, _, err := inventory.SplitInfos(infos)
+	if err != nil {
+		return err
+	}
+
 	// Run the destroyer. It will return a channel where we can receive updates
 	// to keep track of progress and any issues.
-	ch := r.Destroyer.Run()
+	ch := r.Destroyer.Run(inv)
```
mortent, line 82: I think this is an example where if we could just let the command accept the information needed to look up the inventory object, that would be a better api.

mortent (approved):
/lgtm

PR #245, opened 2020-09-30
[WIP] Define the public API for cli-utils

mortent:
Define the API for apply-as-a-library

### DIFF
--- a/docs/api/apply/apply.go
+++ b/docs/api/apply/apply.go
@@ -1,0 +1,75 @@
+package apply
+
+import (
+	"context"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/cli-utils/docs/api/event"
+	"sigs.k8s.io/cli-utils/docs/api/inventory"
+	"sigs.k8s.io/cli-utils/docs/api/provider"
+	"sigs.k8s.io/cli-utils/pkg/common"
+)
+
+// NewApplier creates a new applier based on the given provider.
+func NewApplier(provider provider.Provider) *applier {
+	return nil
+}
+
+// applier can apply resources to a cluster. Each instance of the applier
+// can be used with a single cluster and using a single strategy for how
+// objects for the inventory should be created.
+type applier struct {
+
+}
+
+// Options defines the different properties that can be set to adjust
+// the behavior of a call to Apply.
+type Options struct {
+	// ReconcileTimeout defines whether the applier should wait
+	// until all applied resources have been reconciled, and if so,
+	// how long to wait.
+	ReconcileTimeout time.Duration
+
+	// PollInterval defines how often we should poll for the status
+	// of resources.
+	PollInterval time.Duration
+
+	// EmitStatusEvents defines whether status events should be
+	// emitted on the eventChannel to the caller.
+	EmitStatusEvents bool
+
+	// NoPrune defines whether pruning of previously applied
+	// objects should happen after apply.
+	NoPrune bool
+
+	// DryRunStrategy defines whether changes should actually be performed,
+	// or if it is just talk and no action.
+	DryRunStrategy common.DryRunStrategy
+
+	// PrunePropagationPolicy defines the deletion propagation policy
+	// that should be used for pruning. If this is not provided, the
+	// default is to use the Background policy.
+	PrunePropagationPolicy metav1.DeletionPropagation
+
+	// PruneTimeout defines whether we should wait for all resources
+	// to be fully deleted after pruning, and if so, how long we should
+	// wait.
+	PruneTimeout time.Duration
+}
+
+// Apply will apply the provided set of resources to the cluster given by
+// the provider used when the applier was created. The inventoryInfo parameter
+// contains the data needed to create a new resource for the inventory or to
+// look up an existing one. The resource type for the inventory is given by
+// the provider used when creating the applier. The options parameter contains
+// the properties that can be set to adjust the behavior of Apply, for example
+// to make a dry run.
+// The return value is a go channel that will provide updates on the progress
+// of applying (and optionally pruning) resources. When the operation is
+// complete, the channel will be closed.
+func (a *applier) Apply(ctx context.Context, inventoryInfo inventory.InventoryInfo,
+	resources []*unstructured.Unstructured, options Options) <-chan event.Event {
+	return nil
+}
--- a/docs/api/destroy/destroy.go
+++ b/docs/api/destroy/destroy.go
@@ -1,0 +1,54 @@
+package destroy
+
+import (
+	"context"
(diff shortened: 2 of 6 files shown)


docs/api/apply/apply.go
```diff
@@ -0,0 +1,75 @@
...
+	// of resources.
+	PollInterval time.Duration
+
+	// EmitStatusEvents defines whether status events should be
+	// emitted on the eventChannel to the caller.
+	EmitStatusEvents bool
+
+	// NoPrune defines whether pruning of previously applied
+	// objects should happen after apply.
+	NoPrune bool
```
monopole, line 45: It's tempting to turn this and EmitStatusEvents into enums (like DryRunStrategy) to allow for more prune and status emission styles.

docs/api/event/event.go
```diff
@@ -0,0 +1,158 @@
...
+	// DeleteEvent contains information about object that have been
+	// deleted.
+	DeleteEvent DeleteEvent
+}
+
+// Type defines the go type used to enumerate the possible types of events.
+type Type int
+
+const (
+	InitType Type = iota
```
monopole, line 45: how about reserving the first value of every enum to be the Unknown entry, e.g. 
```
UnknownType
UnknownAction
```
helpful in error message, avoids accidentally defaulting to the first thing, etc.

docs/api/apply/apply.go
mortent: Yeah, that is an interesting suggestion. One of the challenges with using boolean variable, is that we can't distinguish between the value being false and the caller not having specified it at all, which makes defaulting fields harder. If the default value of the enum is something like `Unknown..`, it would solve this issue.

docs/api/event/event.go
mortent: Yeah, it does have some benefits. My concern is that it will require callers of the api to handle these values, and it is not clear that there is any reasonable action here other than to just panic.

docs/api/apply/apply.go
```diff
@@ -0,0 +1,75 @@
...
+// contains the data needed to create a new resource for the inventory or to
+// look up an existing one. The resource type for the inventory is given by
+// the provider used when creating the applier. The options parameter contains
+// the properties that can be set to adjust the behavior of Apply, for example
+// to make a dry run.
+// The return value is a go channel that will provide updates on the progress
+// of applying (and optionally pruning) resources. When the operation is
+// complete, the channel will be closed.
+func (a *applier) Apply(ctx context.Context, inventoryInfo inventory.InventoryInfo,
+	resources []*unstructured.Unstructured, options Options) <-chan event.Event {
```
seans3, line 73: We may want to turn this `resources` slice into a simple union of 1) a dir string, or 2) the `io.Reader` from stdin.

seans3, line 73: I disagree that we expose the `Event` channel. We can wrap the `Event` channel and return a simple `Error` instead. If we expose the `Event` channel, we have now made this API significantly more complex and difficult to use.

docs/api/destroy/destroy.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Destroy will delete all resources from the cluster listed in the inventory
+// of the resource identified in the inventoryInfo. The options parameter
+// contains the properties that can be set to adjust the behavior of Destroy,
+// for example to make a dry run.
+// The return value is a go channel that will provide updates on the progress
+// of destroying the resources. When the operation is complete, the channel
+// will be closed.
+func (d *destroyer) Destroy(ctx context.Context, inventoryInfo inventory.InventoryInfo,
+	options Options) <-chan event.Event {
```
seans3, line 52: Same issue with exposing the `Event` channel. This significantly increases the complexity of the API.

docs/api/event/event.go
```diff
@@ -0,0 +1,158 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	pollevent "sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
+	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+// Event describes something that has happened during the process or either
+// applying resources or removing resources from the cluster.
+// The Type property defines which event has taken place and the
+// corresponding property will contain more information about the details.
+type Event struct {
```
seans3, line 13: Same. I don't think we should expose the `Event`

docs/api/apply/apply.go
seans3: Another strategy for the resource config parameter: pass in a single `io.Reader`. We would build a `FilePathReader` which implements the `io.Reader` interface, and it takes a single parameter, directory, as a string. It's `Read` function would recursively walk the directory to return the resource config YAML or JSON. This `io.Reader` could also be `stdin` to fit in with the current implementation.

seans3: Additionally, we could expose two functions: a more complex one with the events on the channel, and a simpler one which only returns an error, and which does the printing automatically when reading from the `Event` channel.

mortent: So I think we absolutely should expose the API that returns the event channel. Apply with status is a long-running operation, so providing updates on progress is very useful. We can provide a wrapper around that API that just turns it into a synchronous call that just returns an error if something fails. I don't think the library should handle printing.

mortent: Providing a path as a string is not a great solution in my opinion. We should expect the caller of the library to read the manifests from whatever source (git, web, disk...) they chose. 
An `io.Reader` is more reasonable, but I still don't see why it is better than just accepting a slice of unstructured. Using unstructured allows users to create resources in code and apply it directly with the applier. And turning resources from an `io.Reader` into unstructured is pretty straight-forward (and I do think we can provide the ManifestReaders as helpers with the Applier library).

It might be that we should use a slice of `Object` (from https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/object.go#L26) instead of unstructured. That would allow users to pass in both unstructured and any concrete types.

Liujingfang1, line 73: I want to understand more on the even being returned. Assume `resources` contain three objects, how may event does the event channel return? Is there a 1-to-1 mapping between the event and the resource being applied? After all resources are applied, does the event channel close?

docs/api/event/event.go
Liujingfang1, line 13: Does `Event` need to have a field for runtime.Object or at least for the group, kind, name, namespace?

docs/api/event/event.go
```diff
@@ -0,0 +1,158 @@
...
+
+// ResourceAction defines the go type used to enumerate the different actions
+// that can be described in the Init event.
+type ResourceAction int
+
+const (
+	ApplyAction ResourceAction = iota
+	PruneAction
+	DeleteAction
+)
```
seaneagan, line 67: Would it make sense to add a `StatusAction` here? Or if not is there another plan for how to expose the status functionality?

Liujingfang1:
@mortent I think we can close this one.

PR #246, opened 2020-10-02
Move ApplyOptions into ApplyTask

mortent:
This moves the use of the `ApplyOptions` from the `Applier` into the `ApplyTask`. This avoids calling the `NewApplyOptions` and `ao.Complete` functions, but instead builds the `ApplyOptions` instance directly by populating the fields. The advantage of this is that we no longer need to pass in cobra commands when creating the `ApplyOptions` object and we no longer get all the cobra flags from `ApplyOptions`.

### DIFF
--- a/pkg/apply/task/apply_task.go
+++ b/pkg/apply/task/apply_task.go
@@ -4,10 +4,16 @@
 package task
 
 import (
+	"io/ioutil"
+
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/resource"
 	"k8s.io/kubectl/pkg/cmd/apply"
+	"k8s.io/kubectl/pkg/cmd/delete"
+	"k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/util/slice"
 	"sigs.k8s.io/cli-utils/pkg/apply/event"
 	"sigs.k8s.io/cli-utils/pkg/apply/info"
@@ -15,17 +21,6 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/cli-utils/pkg/object"
 )
-
-// ApplyTask applies the given Objects to the cluster
-// by using the ApplyOptions.
-type ApplyTask struct {
-	ApplyOptions   applyOptions
-	InfoHelper     info.InfoHelper
-	Mapper         meta.RESTMapper
-	Objects        []*resource.Info
-	CRDs           []*resource.Info
-	DryRunStrategy common.DryRunStrategy
-}
 
 // applyOptions defines the two key functions on the ApplyOptions
 // struct that is used by the ApplyTask.
@@ -40,6 +35,21 @@ type applyOptions interface {
 	SetObjects([]*resource.Info)
 }
 
+// ApplyTask applies the given Objects to the cluster
+// by using the ApplyOptions.
+type ApplyTask struct {
+	Factory        util.Factory
+	InfoHelper     info.InfoHelper
+	Mapper         meta.RESTMapper
+	Objects        []*resource.Info
+	CRDs           []*resource.Info
+	DryRunStrategy common.DryRunStrategy
+}
+
+// applyOptionsFactoryFunc is a factory function for creating a new
+// applyOptions implementation. Used to allow unit testing.
+var applyOptionsFactoryFunc = newApplyOptions
+
 // Start creates a new goroutine that will invoke
 // the Run function on the ApplyOptions to update
 // the cluster. It will push a TaskResult on the taskChannel
@@ -50,9 +60,6 @@ type applyOptions interface {
 // the desired state of a resource is changed.
 func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
 	go func() {
-		// Update the dry-run field on the Applier.
-		a.setApplyOptionsFields(taskContext.EventChannel())
-
 		objects := a.Objects
 
 		// If this is a dry run, we need to handle situations where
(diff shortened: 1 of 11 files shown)


pkg/apply/task/apply_task.go
```diff
@@ -128,22 +142,48 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
...
+			ErrOut: ioutil.Discard, // TODO: Warning for no lastConfigurationAnnotation
+			// is printed directly to stderr in ApplyOptions. We
+			// should turn that into a warning on the event channel.
+		},
+		DeleteOptions: &delete.DeleteOptions{},
+		PrintFlags: &genericclioptions.PrintFlags{
+			OutputFormat: &emptyString,
+		},
+		ServerSideApply: strategy.ServerDryRun(),
+		FieldManager:    "kubectl", // TODO: Make this configurable
```
seans3, line 173: Until we implement `server-side apply` this will not do anything.

pkg/apply/task/apply_task.go
```diff
@@ -128,22 +142,48 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
...
+		DryRun:          strategy.ClientOrServerDryRun(),
+		ServerDryRun:    strategy.ServerDryRun(),
+		ToPrinter: (&KubectlPrinterAdapter{
```
seans3, line 176: Nice.

seans3 (commented):
I believe this is largely the way to go: manually fill in the `ApplyOptions`, and get rid of the `cobra.Command`. I'm not sure of the utility of moving this `ApplyOptions` creation down into to the `ApplyTask` though. Let's get together and talk about it. My take on this same strategy is here: https://github.com/kubernetes-sigs/cli-utils/pull/251

pkg/apply/task/apply_task.go
seans3: Disregard this. You're right--we need to set it, otherwise the apply can crash for server-side dry run.

pkg/apply/task/apply_task.go
seans3: When looking at the `ApplyOptions` in kubectl apply, the `Complete` function has the following code for the `ToPrinter` to append a string for dry-run:
```
  o.ToPrinter = func(operation string) (printers.ResourcePrinter, error) {
        o.PrintFlags.NamePrintFlags.Operation = operation
        if o.DryRun {
            o.PrintFlags.Complete("%s (dry run)")
        }
        if o.ServerDryRun {
            o.PrintFlags.Complete("%s (server dry run)")
        }
        return o.PrintFlags.ToPrinter()
    }
```

Do we need to add this, or is it not used in `kpt live apply` ?

mortent: We don't use it. We replace it with a different implementation of the `ToPrinter` function.

pkg/apply/task/apply_task.go
```diff
@@ -128,22 +142,48 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
...
+		Recorder:          genericclioptions.NoopRecorder{},
+		IOStreams: genericclioptions.IOStreams{
+			Out:    ioutil.Discard,
+			ErrOut: ioutil.Discard, // TODO: Warning for no lastConfigurationAnnotation
+			// is printed directly to stderr in ApplyOptions. We
+			// should turn that into a warning on the event channel.
+		},
+		DeleteOptions: &delete.DeleteOptions{},
+		PrintFlags: &genericclioptions.PrintFlags{
+			OutputFormat: &emptyString,
```
seans3, line 170: Note to self: Clean up the `kubectl apply` code to not use this field `PrintFlags.OutputFormat` in calculations.

pkg/apply/task/apply_task.go
```diff
@@ -128,22 +142,48 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
...
 	}
-}
 
-func (a *ApplyTask) setApplyOptionsFields(eventChannel chan event.Event) {
-	if ao, ok := a.ApplyOptions.(*apply.ApplyOptions); ok {
-		ao.DryRun = a.DryRunStrategy.ClientDryRun()
-		ao.ServerDryRun = a.DryRunStrategy.ServerDryRun()
-		adapter := &KubectlPrinterAdapter{
+	emptyString := ""
+	return &apply.ApplyOptions{
```
seans3, line 156: Not technically necessary, but maybe useful for documentation: we may want to set the pre-processor and post-processor hooks to nil here.

pkg/apply/task/apply_task.go
```diff
@@ -128,22 +142,48 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
...
+		Overwrite:         true, // Normally set in apply.NewApplyOptions
+		OpenAPIPatch:      true, // Normally set in apply.NewApplyOptions
+		Recorder:          genericclioptions.NoopRecorder{},
+		IOStreams: genericclioptions.IOStreams{
+			Out:    ioutil.Discard,
+			ErrOut: ioutil.Discard, // TODO: Warning for no lastConfigurationAnnotation
+			// is printed directly to stderr in ApplyOptions. We
+			// should turn that into a warning on the event channel.
+		},
+		DeleteOptions: &delete.DeleteOptions{},
```
seans3, line 168: We may want to put a comment that we don't need the `DeleteOptions.FilenameOptions` because we use the `ManifestReader` to create infos and set them with `kubectl.SetObjects()`.

pkg/apply/task/apply_task.go
```diff
@@ -128,22 +142,48 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
...
+		PrintFlags: &genericclioptions.PrintFlags{
+			OutputFormat: &emptyString,
+		},
+		ServerSideApply: strategy.ServerDryRun(),
```
seans3, line 172: Should we hard-code the `ForceConflicts` field if we're setting the related `ServerSideApply` and `FieldManager`?

seans3, line 172: Setting this field is a bit subtle, so we may want to add a comment that we're only currently doing server-side apply for dry-run (still client-side apply)>

pkg/apply/task/apply_task.go
seans3, line 156: It doesn't look like we're setting the `ApplyOptions.Namespace` or `ApplyOptions.EnforceNamespace` here, but I know we set it elsewhere. The `kubectl apply` sets it here using:
```
applyOptions.Namespace, applyOptions.EnforceNamespace, err = factory.ToRawKubeConfigLoader().Namespace()
	if err != nil {
		return nil, err
	}
```
Does it make sense to set it here?

seans3 (commented):
This looks good. Manually filling in the `kubectl.ApplyOptions` is subtle, so we may need more comments to document what we're doing there. I'll LGTM this soon.

pkg/apply/task/apply_task.go
mortent: I thought about it, but I think we want to either set all properties, or just set the ones that are needed. Since the pre-processor and post-processor hooks are nil by default, I decided not to set them here.

pkg/apply/task/apply_task.go
mortent: So I think we should just leave it for now since we don't officially support SSA yet. As far as I can tell, it defaults to `false` in kubectl.

pkg/apply/task/apply_task.go
mortent: I don't think it is needed. Those are used by the Builder and we don't use it (or set it on the ApplyOptions).

pkg/apply/task/apply_task.go
mortent: Added comment

pkg/apply/task/apply_task.go
mortent: Added comment

seans3 (approved):
/lgtm

PR #248, opened 2020-10-05
Update to the latest version of golint

mortent:
Updates to the latest version of golint as it is supposed to be faster, so hopefully it will resolve the timeout issues. This also fixes some lint issues found by the new version.


seans3 (approved):
/lgtm
/approve

PR #250, opened 2020-10-09
Stores InventoryClient in FakeProvider

seans3:
* Store the `FakeInventoryClient` the `FakeProvider` provides.
* Allows tests to retrieve and query the `FakeInventoryClient` when using a `FakeProvider`


mortent (approved):
/lgtm

PR #253, opened 2020-10-15
Clean up the event hierarchy

mortent:
This cleans up the event hierarchy used by the Applier:
 * There is now a `StatusEvent` similar to the other event types instead of reusing the event types from the kstatus package. This means the `StatusEvent` now looks the same as the other event types.
 * Removes the `CompletedEvent` from kstatus as it is no longer in use. Polling is completed when the channel is closed.


seans3 (approved):
/lgtm

PR #258, opened 2020-10-20
Use Unstructured instead of Info

mortent:
First step to replace the use of `resource.Info` with `unstructured.Unstructured`. It will require several PRs to complete this, but this makes the changes for the main `Applier` logic. 

With `resource.Info`, the passed in resources are updated with the state from the cluster after apply. This is no longer the case now. We currently only rely on this for fetching the generation of the resources after apply, but we still handle this in the `ApplyTask`.


Liujingfang1:
Looks good to me! Delegate Sean to approve it.

seans3 (approved):
/lgtm

PR #261, opened 2020-10-24
Update the ManifestReader interface and implementations to use Unstructured

mortent:


pkg/manifestreader/common.go
```diff
@@ -87,12 +81,11 @@ func SetNamespaces(factory util.Factory, infos []*resource.Info,
 		// namespace-scoped. If it is the latter, we set the namespace
 		// to the provided default.
 		var scope string
-		for _, crdInf := range crdInfos {
-			u, _ := crdInf.Object.(*unstructured.Unstructured)
-			group, _, _ := unstructured.NestedString(u.Object, "spec", "group")
-			kind, _, _ := unstructured.NestedString(u.Object, "spec", "names", "kind")
+		for _, crdObj := range crdObjs {
+			group, _, _ := unstructured.NestedString(crdObj.Object, "spec", "group")
```
seans3, line 85: Better to use `unstructured.GroupVersionKind().Group` here.

pkg/manifestreader/common.go
```diff
@@ -87,12 +81,11 @@ func SetNamespaces(factory util.Factory, infos []*resource.Info,
...
+			kind, _, _ := unstructured.NestedString(crdObj.Object, "spec", "names", "kind")
```
seans3, line 86: Better to use `unstructured.GetKind()` here.

pkg/object/infos.go
```diff
@@ -6,19 +6,38 @@ package object
...
+		return nil, err
+	}
+
+	source := "unstructured"
+	if found {
+		path, ok := annos[kioutil.PathAnnotation]
+		if ok {
+			source = path
+			delete(annos, kioutil.PathAnnotation)
+			err := unstructured.SetNestedStringMap(obj.Object, annos, "metadata", "annotations")
```
seans3, line 28: Too low level here: better to use `unstructure.SetAnnotations()`.

pkg/manifestreader/common.go
```diff
@@ -102,26 +95,56 @@ func SetNamespaces(factory util.Factory, infos []*resource.Info,
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// kyamlNodeToUnstructured take a resource represented as a kyaml RNode and
+// turns it into an Unstructured object.
+func kyamlNodeToUnstructured(n *yaml.RNode) (*unstructured.Unstructured, error) {
```
seans3, line 135: This is new and looks like it's the foundation of all our `ManifestReaders`. Should we add a unit test?

pkg/manifestreader/common.go
```diff
@@ -102,26 +95,56 @@ func SetNamespaces(factory util.Factory, infos []*resource.Info,
...
+			filteredObjs = append(filteredObjs, obj)
 		}
 	}
-	return filterInfos
+	return filteredObjs
+}
+
+// removeAnnotations removes any Path or Index annotations from the
+// resource.
+func removeAnnotations(n *yaml.RNode, annotations ...kioutil.AnnotationKey) error {
```
seans3, line 123: This is new and looks like it's the foundation of all our `ManifestReaders`. Should we add a unit test?

seans3 (changes requested):
Looks good. I think there are a couple places that you're too low level for `unstructured`, using `GetNestedString` instead of the higher level functions. `unstructured` has a good set of metadata functions (e.g. `GetKind()`). Now that we're using the kyaml as the basis for the `ManifestReaders`, we probably want a couple simple unit tests for the new functions.

pkg/manifestreader/common.go
mortent: This is not the GVK of the resource. The resource we are looking at there is a CRD and we are looking up information about the resource type that is defined by this CRD.

pkg/manifestreader/common.go
mortent: Same as above.

pkg/manifestreader/common.go
mortent: So this is used by the tests in both the `PathManifestReader` and the `StreamManifestReader`, so I don't think we need to test this separately.

pkg/manifestreader/common.go
mortent: Yeah, added some unit tests for this.

pkg/object/infos.go
seans3: actually, there is: `obj.Object.SetAnnotations(annos)`.

seans3: https://github.com/kubernetes/kubernetes/blob/323f34858de18b862d43c40b2cced65ad8e24052/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured/unstructured.go#L412

pkg/object/infos.go
```diff
@@ -4,21 +4,34 @@
...
+	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 )
 
 func InfoToUnstructured(info *resource.Info) *unstructured.Unstructured {
 	return info.Object.(*unstructured.Unstructured)
 }
 
-func UnstructuredToInfo(obj *unstructured.Unstructured) *resource.Info {
+func UnstructuredToInfo(obj *unstructured.Unstructured) (*resource.Info, error) {
+	accessor, _ := meta.Accessor(obj)
```
seans3, line 18: Minor nit: we don't need the `meta.Accessor`. `unstructured` has a simple `GetAnnotations` and `SetAnnotations` function. So it would just be `obj.GetAnnotations()` and `obj.SetAnnotations(annots)`.

pkg/manifestreader/common_test.go
```diff
@@ -147,10 +153,10 @@ func TestSetNamespaces(t *testing.T) {
...
 			assert.NoError(t, err)
 
-			for i, inf := range tc.infos {
+			for i, obj := range tc.objs {
 				expectedNs := tc.expectedNamespaces[i]
-				assert.Equal(t, expectedNs, inf.Namespace)
-				accessor, _ := meta.Accessor(inf.Object)
+				assert.Equal(t, expectedNs, obj.GetNamespace())
+				accessor, _ := meta.Accessor(obj)
 				assert.Equal(t, expectedNs, accessor.GetNamespace())
```
seans3, line 160: Is this the same test as the one two lines up?

seans3 (approved):
Only a couple minor nits; approving

/lgtm

PR #262, opened 2020-10-26
Export ExpandDir function and add an extra return value

seans3:
* Exports `common.ExpandDir` function.
* Adds return value of inventory template filepath to this function, which was already being calculated.
* Adds unit tests for this newly exported function.

### DIFF
--- a/pkg/common/path.go
+++ b/pkg/common/path.go
@@ -80,7 +80,7 @@ func ExpandPackageDir(f genericclioptions.FileNameFlags) (genericclioptions.File
 		return f, fmt.Errorf("expand package directory should pass one package directory. "+
 			"Passed the following paths: %v", f.Filenames)
 	}
-	configFilepaths, err := expandDir((*f.Filenames)[0])
+	_, configFilepaths, err := ExpandDir((*f.Filenames)[0])
 	if err != nil {
 		return f, err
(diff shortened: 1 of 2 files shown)


pkg/common/path.go
```diff
@@ -137,28 +137,31 @@ func FilterInputFile(in io.Reader, tmpDir string) error {
...
+		return "", filepaths, err
 	}
+	var invFilepath string
 	for _, node := range nodes {
 		meta, err := node.GetMeta()
 		if err != nil {
 			continue
 		}
+		path := meta.Annotations[kioutil.PathAnnotation]
+		path = filepath.Join(dir, path)
```
mortent, line 157: Does this work for packages with subfolders? It might be worth adding a test for this.

pkg/common/path.go
```diff
@@ -137,28 +137,31 @@ func FilterInputFile(in io.Reader, tmpDir string) error {
 }
 
 // expandDir takes a single package directory as a parameter, and returns
```
mortent, line 139: Update the name of the function in the comment to be capitalized.

pkg/common/path.go
seans3: Added unit test for this case.

PR #263, opened 2020-10-29
Add output flag to destroy and preview commands

fsommar:
Hi 👋 

This PR adds output flag support to the `destroy` and `preview` commands, following suite with the output functionality already present in the `apply` command.

The flag is duplicated across the three commands right now. What do you think, is it fine as is or should I refactor it with code sharing in mind?


mortent:
This looks good. Thanks @fsommar!
/lgtm
/approve

PR #264, opened 2020-10-29
split inventory and resources in the apply interface

Liujingfang1:
Change the `Applier.Run` from
```
func (a *Applier) Run(ctx context.Context, objects []*unstructured.Unstructured,
options Options) <-chan event.Event {}
```
to 
```
func (a *Applier) Run(ctx context.Context, inventory *unstructured.Unstructured, objects []*unstructured.Unstructured,
options Options) <-chan event.Event {
```

### DIFF
--- a/cmd/apply/cmdapply.go
+++ b/cmd/apply/cmdapply.go
@@ -17,6 +17,7 @@ import (
 	"sigs.k8s.io/cli-utils/cmd/printers"
 	"sigs.k8s.io/cli-utils/pkg/apply"
 	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
 	"sigs.k8s.io/cli-utils/pkg/provider"
 	"sigs.k8s.io/kustomize/kyaml/setters2"
 )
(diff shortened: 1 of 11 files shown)


mortent:
This makes sense. Long term I don't think we want the type of the inventory object to be `*unstructured.Unstructured`, but rather just the information needed to create/lookup the object containing the inventory. The details of how it is created and updated should be part of the Provider.

Liujingfang1:
This is an intermediate state. I plan to change it to `inventory.Inventory`.

mortent:
Sounds good. This looks good to me. I'll let @seans3 take a look too.

Liujingfang1:
The e2e test failed, I'm working on it.

Liujingfang1:
The tests all passed. Please take a look @seans3

seans3:
This looks good. I'm going to submit the migrate PR first if that is OK.

Liujingfang1:
Yes, that's OK.

Liujingfang1:
Rebased on recent change. PTAL @seans3

seans3 (approved):
/lgtm

PR #268, opened 2020-11-01
Fixes namespace calculation for cluster-scoped resources

seans3:
* Fixes namespace calculation for Inventory in `allInSameNamespace` function, by ignoring cluster-scoped resources.
* Example: So if N resources all have the same namespace and another resource (Type: Namespace) is cluster-scoped, this function should return the same namespace of the N resources.
* Adds unit test.

### DIFF
--- a/pkg/config/initoptions.go
+++ b/pkg/config/initoptions.go
@@ -13,11 +13,13 @@ import (
 
 	"github.com/google/uuid"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/klog"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/cli-utils/pkg/inventory/configmap"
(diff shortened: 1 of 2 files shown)


pkg/config/initoptions.go
```diff
@@ -148,16 +154,24 @@ func allInSameNamespace(packageDir string) (string, bool, error) {
 		if err != nil {
 			return "", false, err
 		}
+		// Skip cluster-scoped resources
+		namespaced, found := openapi.IsNamespaceScoped(rm.TypeMeta)
```
mortent, line 158: We should just add a comment here that if we can't determine is a resource is namespace-scoped, we will default to assuming it is namespaced. That sounds reasonable. Also we should make sure this is reflected in the docs.

mortent (commented):
One small comment, but otherwise it looks good.

pkg/config/initoptions.go
seans3: Updated comment. Docs will be updated in separate PR.

mortent (approved):
/lgtm

PR #269, opened 2020-11-01
Adds more server-side flags to preview command

seans3:
* Adds server-side preview flags `--field-manager` and `--force-conflicts` to make preview consistent with apply.

### DIFF
--- a/cmd/preview/cmdpreview.go
+++ b/cmd/preview/cmdpreview.go
@@ -23,7 +23,6 @@ import (
 
 var (
 	noPrune        = false
-	serverDryRun   = false
 	previewDestroy = false
 )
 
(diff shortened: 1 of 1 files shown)


mortent:
Does the serverside dry-run use these additional parameters, i.e. are there situations where changing the value of the field manager changes the result of a dry-run?
Otherwise it looks good.

seans3:
> Does the serverside dry-run use these additional parameters, i.e. are there situations where changing the value of the field manager changes the result of a dry-run?
> Otherwise it looks good.

Yes it does. I works the same as the apply. Here is the output for a field that I modified to conflict and changing to a different field manager:
```
$ kapply preview --server-side --field-manager=sean ~/testdata/testdata-namespace-2
error: Apply failed with 1 conflict: conflict with "kubectl": .metadata.labels.name

Please review the fields above--they currently have other managers. Here
are the ways you can resolve this warning:
* If you intend to manage all of these fields, please re-run the apply
  command with the `--force-conflicts` flag.
* If you do not intend to manage all of the fields, please edit your
  manifest to remove references to the fields that should keep their
  current managers.
* You may co-own fields by updating your manifest to match the existing
  value; in this case, you'll become the manager if the other manager(s)
  stop managing the field (remove it from their configuration).

See http://k8s.io/docs/reference/using-api/api-concepts/#conflicts

$ kapply preview --server-side --field-manager=sean --force-conflicts ~/testdata/testdata-namespace-2
pod/pod-b serversideapplied (preview-server)
pod/pod-c serversideapplied (preview-server)
pod/pod-d serversideapplied (preview-server)
3 resource(s) applied. 0 created, 0 unchanged, 0 configured, 3 serverside applied (preview-server)
0 resource(s) pruned, 0 skipped (preview-server)
```

PR #270, opened 2020-11-02
change the apply interface to use inventory.InventoryInfo

### LINKED ISSUE #267: change the Apply interface to use inventory.Inventory instead of unstructured.Unstructured
Change the Applier.Run from
```
func (a *Applier) Run(ctx context.Context, objects []*unstructured.Unstructured,
options Options) <-chan event.Event {}
```
to
```
func (a *Applier) Run(ctx context.Context, inventory inventory.Inventory, objects []*unstructured.Unstructured,
options Options) <-chan event.Event {}
```

Liujingfang1:
Fix #267

### DIFF
--- a/cmd/destroy/cmddestroy.go
+++ b/cmd/destroy/cmddestroy.go
@@ -13,7 +13,6 @@ import (
 	"k8s.io/kubectl/pkg/util/i18n"
 	"sigs.k8s.io/cli-utils/cmd/printers"
 	"sigs.k8s.io/cli-utils/pkg/apply"
-	"sigs.k8s.io/cli-utils/pkg/inventory"
 	"sigs.k8s.io/cli-utils/pkg/provider"
 )
 
@@ -64,12 +63,7 @@ func (r *DestroyRunner) RunE(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	inventoryClient, err := r.provider.InventoryClient()
-	if err != nil {
-		return err
-	}
-
-	inv, _, err := inventory.SplitUnstructureds(inventoryClient.InvInfoFactoryFunc(), objs)
+	inv, _, err := r.provider.InventoryInfo(objs)
 	if err != nil {
 		return err
 	}
--- a/cmd/destroy/cmddestroy.go
+++ b/cmd/destroy/cmddestroy.go
@@ -64,7 +64,12 @@ func (r *DestroyRunner) RunE(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	inv, _, err := inventory.SplitUnstructureds(objs)
+	inventoryClient, err := r.provider.InventoryClient()
+	if err != nil {
+		return err
+	}
+
+	inv, _, err := inventory.SplitUnstructureds(inventoryClient.InvInfoFactoryFunc(), objs)
 	if err != nil {
 		return err
 	}
--- a/cmd/destroy/cmddestroy.go
+++ b/cmd/destroy/cmddestroy.go
@@ -13,15 +13,17 @@ import (
 	"k8s.io/kubectl/pkg/util/i18n"
 	"sigs.k8s.io/cli-utils/cmd/printers"
 	"sigs.k8s.io/cli-utils/pkg/apply"
+	"sigs.k8s.io/cli-utils/pkg/manifestreader"
 	"sigs.k8s.io/cli-utils/pkg/provider"
 )
 
 // GetDestroyRunner creates and returns the DestroyRunner which stores the cobra command.
-func GetDestroyRunner(provider provider.Provider, ioStreams genericclioptions.IOStreams) *DestroyRunner {
+func GetDestroyRunner(provider provider.Provider, loader manifestreader.ManifestLoader, ioStreams genericclioptions.IOStreams) *DestroyRunner {
 	r := &DestroyRunner{
 		Destroyer: apply.NewDestroyer(provider),
 		ioStreams: ioStreams,
 		provider:  provider,
+		loader:    loader,
 	}
 	cmd := &cobra.Command{
 		Use:                   "destroy (DIRECTORY | STDIN)",
@@ -40,7 +42,8 @@ func GetDestroyRunner(provider provider.Provider, ioStreams genericclioptions.IO
 // DestroyCommand creates the DestroyRunner, returning the cobra command associated with it.
 func DestroyCommand(f cmdutil.Factory, ioStreams genericclioptions.IOStreams) *cobra.Command {
 	provider := provider.NewProvider(f)
-	return GetDestroyRunner(provider, ioStreams).Command
+	loader := manifestreader.NewManifestLoader(f)
+	return GetDestroyRunner(provider, loader, ioStreams).Command
 }
 
 // DestroyRunner encapsulates data necessary to run the destroy command.
@@ -49,13 +52,14 @@ type DestroyRunner struct {
 	ioStreams genericclioptions.IOStreams
 	Destroyer *apply.Destroyer
 	provider  provider.Provider
+	loader    manifestreader.ManifestLoader
 
 	output string
 }
 
 func (r *DestroyRunner) RunE(cmd *cobra.Command, args []string) error {
 	// Retrieve the inventory object.
-	reader, err := r.provider.ManifestReader(cmd.InOrStdin(), args)
+	reader, err := r.loader.ManifestReader(cmd.InOrStdin(), args)
 	if err != nil {
 		return err
 	}
@@ -63,7 +67,7 @@ func (r *DestroyRunner) RunE(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	inv, _, err := r.provider.InventoryInfo(objs)
+	inv, _, err := r.loader.InventoryInfo(objs)
 	if err != nil {
 		return err
 	}
--- a/cmd/preview/cmdpreview.go
+++ b/cmd/preview/cmdpreview.go
@@ -100,7 +100,12 @@ func (r *PreviewRunner) RunE(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	inv, objs, err := inventory.SplitUnstructureds(objs)
+	inventoryClient, err := r.provider.InventoryClient()
+	if err != nil {
+		return err
+	}
+
+	inv, objs, err := inventory.SplitUnstructureds(inventoryClient.InvInfoFactoryFunc(), objs)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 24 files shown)


cmd/preview/cmdpreview.go
```diff
@@ -117,18 +121,14 @@ func (r *PreviewRunner) RunE(cmd *cobra.Command, args []string) error {
 			ForceConflicts:  false,
 			FieldManager:    common.DefaultFieldManager,
 		}
-		ch = r.Applier.Run(ctx, objs, apply.Options{
+		ch = r.Applier.Run(ctx, inventory.WrapInventoryObj(inv), objs, apply.Options{
```
seans3, line 124: I don't believe this will work. There are two versions of `WrapInventoryObj`; one for ConfigMap and one for `ResourceGroup`. We will need to set this `InventoryFactoryFunc` based on the inventory object that is detected, so we can correctly handle both `ConfigMap` and `ResourceGroup` inventory objects while we're migrating. This is currently being worked on.

seans3 (changes requested):

cmd/destroy/cmddestroy.go
```diff
@@ -68,7 +68,7 @@ func (r *DestroyRunner) RunE(cmd *cobra.Command, args []string) error {
 
 	// Run the destroyer. It will return a channel where we can receive updates
 	// to keep track of progress and any issues.
-	ch := r.Destroyer.Run(inv)
+	ch := r.Destroyer.Run(inventory.WrapInventoryObj(inv))
```
seans3, line 71: I do not believe this will work. As described elsewhere, this `InventoryFactoryFunc` will need to be set in the `Provider`.

cmd/status/cmdstatus.go
```diff
@@ -96,7 +96,7 @@ func (r *StatusRunner) runE(cmd *cobra.Command, args []string) error {
 
 	// Based on the inventory template manifest we look up the inventory
 	// from the live state using the inventory client.
-	identifiers, err := invClient.GetClusterObjs(inv)
+	identifiers, err := invClient.GetClusterObjs(inventory.WrapInventoryObj(inv))
```
seans3, line 99: same

mortent:
So the way I'm thinking about this, is that the inventory information passed into the `Apply` function should only contain the information needed to look up and create a resource for the inventory regardless of type. Only the name, namespace and id should be needed here since for anything that requires detailed knowledge about the type of the resource, we will delegate the work to the provider. But the provider obviously only know the type of the resource, so it needs to be passed in the details, which in the general case is just the same information that was passed to `Apply`.

So in the case where a ConfigMap is used to store the inventory information, the provider will know how to create new ConfigMaps, look up ConfigMaps. But the specifics for each package are passed into the Provider through the `Apply` function in the form of the inventory information struct/interface.

Liujingfang1:
This PR uses the inventory interface as 
```
type Inventory interface {
	// Load retrieves the set of object metadata from the inventory object
	Load() ([]object.ObjMetadata, error)
	// Store the set of object metadata in the inventory object
	Store(objs []object.ObjMetadata) error
	// GetObject returns the object that stores the inventory
	GetObject() (*unstructured.Unstructured, error)
}
```

@mortent Are you saying this interface should be part of the `Provider` and not used here?

Then for the argument of the Run function, it accepts an interface to get the name/namespace/id of an inventory object. Like the following?

```
type inv interface {
      Name() string
      Namespace() string
      Id() string
}
```

mortent:
I think we want to delegate everything that is specific to a particular resource type (that is being used for the inventory) to the Provider. But we want users to be able to create the `Applier` once, and then reuse it for several packages or for the same package multiple times. So anything that is specific to each package must be passed into the functionality in the provider each time. The `inv` interface contains the information that changes between each package, so it is passed into the `Applier.Apply` function and then given to the Provider implementation when needed.

cmd/preview/cmdpreview.go
Liujingfang1: Thanks for pointing this out. I updated it to use the function from the `provider.InventoryClient`. In addition to the `InventoryFactoryFunc`, also added two other functions to change `InventoryInfo` to `Unstructured` and vice versa.

cmd/destroy/cmddestroy.go
Liujingfang1: Updated the code to use the function from `provider.InventoryClient`.

cmd/status/cmdstatus.go
Liujingfang1: Updated the code to use the function from provider.InventoryClient.

Liujingfang1:
Updated the change to use the following interface in the apply interface.

```
type InventoryInfo interface {
      Name() string
      Namespace() string
      Id() string
}
```

To make it work, also added two other functions to convert `InventoryInfo` to `Unstructured` and convert `Unstructured` to `InventoryInfo`.

In the subsequent change, we'll need to change the `Reader()` interface to return an `InventoryInfo` and `[]*Unstructured`. Then in each command, we don't need to split the resources or do the conversion to `InventoryInfo`.

Liujingfang1:
Rebased the recent change. PTAL

pkg/inventory/inventory.go
```diff
@@ -106,7 +114,7 @@ func ValidateNoInventory(objs []*unstructured.Unstructured) error {
 // splitUnstructureds takes a slice of unstructured.Unstructured objects and
 // splits it into one slice that contains the inventory object templates and
 // another one that contains the remaining resources.
-func SplitUnstructureds(objs []*unstructured.Unstructured) (*unstructured.Unstructured, []*unstructured.Unstructured, error) {
+func SplitUnstructureds(factoryFunc UnstructuredToInvInfoFunc, objs []*unstructured.Unstructured) (InventoryInfo, []*unstructured.Unstructured, error) {
```
mortent, line 117: This doesn't seem right. This function should only be needed in the cobra commands in this repo. I think we should just make this work only for ConfigMaps. Any users of the library that use a different resource type for inventory can handle this for their particular use-case.

pkg/inventory/inventory-client.go
```diff
@@ -27,39 +27,47 @@ type InventoryClient interface {
...
+	Replace(inv InventoryInfo, objs []object.ObjMetadata) error
 	// DeleteInventoryObj deletes the passed inventory object from the APIServer.
-	DeleteInventoryObj(inv *unstructured.Unstructured) error
+	DeleteInventoryObj(inv InventoryInfo) error
 	// SetDryRunStrategy sets the dry run strategy on whether this we actually mutate.
 	SetDryRunStrategy(drs common.DryRunStrategy)
+	// ApplyInventoryNamespace applies the Namespace that the inventory object should be in.
 	ApplyInventoryNamespace(invNamespace *unstructured.Unstructured) error
+	// InvInfoFactoryFunc returns the factory function to create an InventoryInfo from an unstructured.
+	InvInfoFactoryFunc() UnstructuredToInvInfoFunc
```
mortent, line 46: Why is this needed? I think this would no longer be needed if we just remove the `SplitUnstructured` function completely from the apply-as-a-library code. In the new structure, the library should never need to do this.

Liujingfang1: This is needed since the inventory object is split from the resource list. The underlying object of `InventoryInfo` is the inventory object, which is not included in the resource list passed into `Apply` or `Run`. That is my previous understanding. If you mean the inventory object should be part of the resource list, then we don't need the `SplitUnstructured` function or the new function added here.

pkg/inventory/inventory.go
Liujingfang1: Since we need to split the list of resources to get the InventoryInfo, I added a function `InventoryInfo` in the provider interface for that purpose. Then the `SplitUnstructureds` is only used in the provider for ConfigMap.

pkg/inventory/inventory-client.go
Liujingfang1: Removed this since the Split function is hidden behind the `provider.InventoryInfo()`.

mortent: So I think this is an improvement. My concern still is that we are mixing up the features needed in the cobra commands (which isn't really something we expect others to reuse) from apply-as-a-library (which we do want to make it easy for others to use). I think we should consider breaking the Provider interface into two separate interfaces:
 * The Provider interface that will contain the `InventoryClient`, `Factory` and `ToRESTMapper` functions (maybe even remove the `ToRESTMapper` function since that is already available through the Factory).
 * Create a separate `ManifestLoader` (open to suggestions for better names here) that will contain the `ManifestReader` and `InventoryInfo` functions. This interface is probably temporary, but it allows us to reuse the cobra commands without polluting the interfaces for apply-as-a-library with things that aren't needed.

The `Get<Apply/Preview/Destroy>Runner` functions will accept both interfaces, but the `NewApplier` function will only need the Provider interface.
WDYT?

Liujingfang1: Agreed with breaking the provider into separate interfaces. One is for parsing/loading resources and the other is for applying. The parsing interface `ManifestLoader` is only used in the cobra commands. It may not be temporary since parsing is also needed in other client. Later we can also put that into its own library. 

`ManifestLoader` can contain `ManifestReader` and `InventoryInfo`. Or we can change the interface of `ManifestReader`, where the reader returns both `InventoryInfo` and `[]*Unstructured`.

Liujingfang1: Updated the change for splitting the interface. PTAL.

mortent:
This looks good. I think splitting the Provider interface works well. I'll let @seans3 take a look too.

cmd/destroy/cmddestroy.go
```diff
@@ -41,7 +42,8 @@ func GetDestroyRunner(provider provider.Provider, ioStreams genericclioptions.IO
 // DestroyCommand creates the DestroyRunner, returning the cobra command associated with it.
 func DestroyCommand(f cmdutil.Factory, ioStreams genericclioptions.IOStreams) *cobra.Command {
 	provider := provider.NewProvider(f)
-	return GetDestroyRunner(provider, ioStreams).Command
+	loader := manifestreader.NewManifestLoader(f)
```
seans3, line 45: Why are we creating a new object here, when the `ManifestReader` can be retrieved from the existing `Provider`. This change will necessitate other changes in layers above this that have created different `Providers` and `ManifestReaders`.

seans3 (changes requested):
Let's get together to talk about the new `ManifestLoader`, since this has several consequences for other layers and parts.

cmd/destroy/cmddestroy.go
seans3: This separation will break the `DualDelegatingProvider` in kpt.

Liujingfang1: The main reason of the separation is to make the boundary between the parsing the applying more clear. The apply library doesn't need to know the `Reader` interface. The caller of the library need to make sure the `InventoryInfo` parsed from the `Reader` work with the `provider`. More details [here](https://github.com/kubernetes-sigs/cli-utils/pull/270#discussion_r517788024).

Yes, this will break the current `DualDelegatingProvider`. That's expected since we are changing the interface. I'll update the kpt side to work with the new cli-utils library.

mortent: Yeah, this will require some adjustments for users of cli-utils, but it makes a much clearer distinction between what is needed for the apply library (the Provider interface) and what is needed in the cobra commands (ManifestLoader).

seans3 (approved):
/lgtm

PR #271, opened 2020-11-03
Moves destroyer Initialize() to immediately before Run()

seans3:
* Very small change which moves `destroyer.Initialize()` to immediately before `destroyer.Run()` in two places.
* Does NOT change any functionality.
* Groups operations more logically.


mortent:
/lgtm
We should see if we still need to have the `Initialize` function at all.

PR #273, opened 2020-11-06
add the new lifecycle annotation for deletion

Liujingfang1:
Add the lifecycle annotation for deletion
```
client.lifecycle.config.k8s.io/deletion: detach
```

At the same time, we keep the existing annotation on deletion for backward compatibility.


pkg/apply/prune/prune.go
```diff
@@ -157,6 +157,11 @@ func preventDeleteAnnotation(annotations map[string]string) bool {
 				return true
 			}
 		}
+		if annotation == common.LifecycleDeleteAnnotation {
```
mortent, line 160: It seems like we could create a data structure which would contain the different combinations or annotation and value we accept to prevent deletion instead of duplicating the nested if statements here?

Liujingfang1: Moved this check into a common function. PTAL.

mortent (approved):
/lgtm

PR #276, opened 2020-11-07
WIP: add option for continue on error

Liujingfang1:


Liujingfang1:
/assign @mortent @seans3 
Share it early to get your opinions on this. The change covers the majority part of the new apply option. I'll also add unit tests for it.

pkg/apply/event/event.go
```diff
@@ -72,7 +72,8 @@ type ResourceGroup struct {
 }
 
 type ErrorEvent struct {
-	Err error
+	Object runtime.Object
```
mortent, line 75: This isn't the right place to add this I think. We can discuss the event hierarchy, but I think we want to add errors to the existing event types for apply/prune... The idea is that the top-level error event is for fatal errors, so the client can assume that after receiving one of these error events, the channel will be closed.

pkg/apply/prune/prune.go
```diff
@@ -96,6 +100,17 @@ func (po *PruneOptions) Prune(localInv *unstructured.Unstructured, localObjs []*
...
 		mapping, err := po.mapper.RESTMapping(clusterObj.GroupKind)
 		if err != nil {
+			u := &unstructured.Unstructured{}
+			u.SetGroupVersionKind(schema.GroupVersionKind{
+				Group: clusterObj.GroupKind.Group,
+				Kind:  clusterObj.GroupKind.Kind,
+			})
+			u.SetNamespace(clusterObj.Namespace)
+			u.SetName((clusterObj.Name))
+			eventChannel <- createErrorEvent(u, err)
```
mortent, line 110: I think we should avoid creating unstructured objects that aren't actually either complete resources read from a manifest or resources read from the cluster. We do have the `object.ObjMeta` type for capturing the identifying information about a resource.

pkg/apply/prune/prune.go
```diff
@@ -96,6 +100,17 @@ func (po *PruneOptions) Prune(localInv *unstructured.Unstructured, localObjs []*
 	for _, clusterObj := range clusterObjs {
 		mapping, err := po.mapper.RESTMapping(clusterObj.GroupKind)
 		if err != nil {
+			u := &unstructured.Unstructured{}
+			u.SetGroupVersionKind(schema.GroupVersionKind{
```
mortent, line 104: We shouldn't create a GVK when we don't have the version.

pkg/apply/solver/solver.go
```diff
@@ -66,105 +67,133 @@ func (t *TaskQueueSolver) BuildTaskQueue(ro resourceObjects,
...
 	}
 
 	return tasksToQueue(tasks)
 }
 
+func (t *TaskQueueSolver) appendApplyTask(tasks []taskrunner.Task,
+	objects []*unstructured.Unstructured,
+	crds []*unstructured.Unstructured,
+	o Options) []taskrunner.Task {
+	if o.ContinueOnError {
```
mortent, line 105: I think we should look at other options before we implement this pattern. This can end up with a large number of tasks...

pkg/apply/taskrunner/runner.go
```diff
@@ -210,7 +214,16 @@ func (b *baseRunner) run(ctx context.Context, taskQueue chan Task,
 			currentTask.ClearTimeout()
 			if msg.Err != nil {
 				b.amendTimeoutError(msg.Err)
-				return msg.Err
+				if !o.continueOnError {
```
mortent, line 217: Is this something that can happen? It seems like we should either handle this within the tasks themselves or always handle it here (not sure if the latter is even possible).

pkg/apply/event/event.go
Liujingfang1: Ideally we'd like include the error in the `ApplyEvent` or `PruneEvent`. For `PruneEvent`, we can make that change since the objects to be pruned are handled one by one. It's straightforward to add an error in the returned `PruneEvent`. The 'ApplyEvent` is blocked by some limitation in the current code: The `ApplyEvent` is sent out only when the object is applied successfully. The `ApplyEvent` is sent out from the kubectl library, it seems not feasible to inject the error in it.

pkg/apply/prune/prune.go
Liujingfang1: Ack, I think adding the error in the `PruneEvent` will remove this change.

pkg/apply/solver/solver.go
Liujingfang1: I considered some other option, but they are not better than the proposed change. According to Sean, when kubectl library >= 1.18, `ContinueOnError` is the default behavior. If we adopt that and create an Apply task with multiple objects, `ContinueOnError` is supported by `AbortOnError` is not supported. If we create a bunch of Apply tasks, each applies one object, then we can support both `ContinueOnError` and `AbortOnError`. It becomes an option the library caller can choose.

Are you concerned about the performance of a large number of tasks? Where could be the bottleneck? If it's in creating the kubectl ApplyOption, we can optimize it to avoid recreating it later.

pkg/apply/taskrunner/runner.go
Liujingfang1: This change is to allow the runner continue with next task when the previous task got an error. I think it is necessary. Assume a typical call of apply-as-a-library, it includes apply tasks and prune tasks. Even if we can handle the error and continue the process within each task, we should continue with the prune tasks even some error happens in the apply tasks.

pkg/apply/event/event.go
Liujingfang1: I figured out a way of doing so. When the kubectl `ApplyOptions.Run` fails, we send an ApplyEvent with an error.

pkg/apply/prune/prune.go
Liujingfang1: Added an `ObjectMeta` field to the `PruneEvent` to solve this problem.

pkg/apply/solver/solver.go
Liujingfang1: Updated this to keep creating one apply task and avoid creating `ApplyOptions` multiple times. The difference will be within the ApplyTask. When `ContinueOnError` is true, the apply tasks calls kubectl `ApplyOptions.Run` multiple times.

seans3:
/hold

Before submitting this, we should upgrade the version of kubectl we use in go.mod.

Liujingfang1:
@seans3 I think this change is independent of upgrading the version of kubectl. No matter how the underlying kubectl behaves, the cli-utils library should provide options for ContinueOnError and not ContinueOnError.

pkg/apply/event/event.go
```diff
@@ -154,4 +159,5 @@ type DeleteEvent struct {
 	Type      DeleteEventType
 	Operation DeleteEventOperation
 	Object    runtime.Object
+	Error     error
```
mortent, line 162: I would assume the `DeleteEvent` should look identical to the `PruneEvent`. Why no `ObjectMeta` here?

pkg/apply/event/event.go
```diff
@@ -129,9 +132,11 @@ const (
 )
 
 type PruneEvent struct {
-	Type      PruneEventType
-	Operation PruneEventOperation
-	Object    runtime.Object
+	Type       PruneEventType
+	Operation  PruneEventOperation
+	Object     runtime.Object
+	ObjectMeta object.ObjMetadata
```
mortent, line 138: We usually name this `Identifier`. Not sure if it is a great name, but we should probably just be consistent.

pkg/apply/event/event.go
```diff
@@ -129,9 +132,11 @@ const (
 )
 
 type PruneEvent struct {
-	Type      PruneEventType
-	Operation PruneEventOperation
-	Object    runtime.Object
+	Type       PruneEventType
+	Operation  PruneEventOperation
+	Object     runtime.Object
```
mortent, line 137: I think we can just make this `*unstructured.Unstructured`. I think we are using a client that will always return unstructured objects, so we can probably make it easier for clients by using a concrete type here.

pkg/apply/taskrunner/runner.go
mortent: But I think the only way an error can end up on the `TaskChannel` is if one of the tasks put it there. So could we assume it is the responsibility of each task to put non-fatal errors on the `EventChannel` rather than the `TaskChannel`? That also means that if an error comes on the `TaskChannel`, it is considered fatal and we should exit.

Liujingfang1:
@seans3 I create another PR for upgrading the kubectl version. Could you please take a look https://github.com/kubernetes-sigs/cli-utils/pull/277? The apply library "ContinueOnError" behavior should be handled within cli-utils and independent of the kubectl's default behavior. With the kubectl upgrade, we still need this change.

pkg/apply/event/event.go
Liujingfang1: My bad, missed that part. Will add it.

pkg/apply/event/event.go
Liujingfang1: Ack, will update it.

pkg/apply/event/event.go
Liujingfang1: That sounds good to me.

pkg/apply/taskrunner/runner.go
Liujingfang1: Yep, the error in `TaskChannel` is from a task. I think it makes sense to not emit error to `TaskChannel` when we want continue the apply. For the errors sent to `EventChannel`, I think you mean they are sent as part of ApplyEvent/PruneEvent/DeleteEvent.

pkg/apply/event/event.go
Liujingfang1: Updated this

pkg/apply/event/event.go
Liujingfang1: Changed it to `*unstructured.Unstructured`.

pkg/apply/applier.go
```diff
@@ -233,6 +234,7 @@ func (a *Applier) Run(ctx context.Context, inventory *unstructured.Unstructured,
 			PollInterval:     options.PollInterval,
 			UseCache:         true,
 			EmitStatusEvents: options.EmitStatusEvents,
+			ContinueOnError:  options.ContinueOnError,
```
seans3, line 237: We do not want this boolean. Continue on error is the default.

seans3 (changes requested):
This is major surgery. Unfortunately, we do not have the end-to-end test coverage to allow this major change. We'll need to add significant e2e tests before considering this change.

Liujingfang1:
@seans3 Agreed that we need significant test on this. Since the CLI doesn't support the continue-on-error, we need to think about how to test it end to end. 

So far the change is only for the Apply and prune libraries, when the resources are already loaded and passed to the `Applier`. It mainly affects the applying and pruning single resources in the apply/prune loop. Please also note that the CLI doesn't have the option for `ContinueOnError`, it exists on the first error encountered. So some of the areas is not affected.

> Reading the resource manifests

This is not changed since reading resources is out side of the apply library.

> Adding the ObjMeta to the Inventory
> Merging the initial Inventory at the beginning of the apply

This is not changed. If we don't block the apply on this, the inventory object in the cluster may result in incorrect state with the resources.

> Applying the resources

This is changed, we want to continue the next resource when the previous resource got an error.
> Pruning the resources

This is changed, we want to prune the next resource when the previous one got an error.

> Polling the status of resources

This is not changed.

> Replacing the final inventory at the end of the algorithm

This shouldn't change. When the prune got an error, the replacing shouldn't on the final inventory shouldn't happen.

Liujingfang1:
closed via #286

PR #279, opened 2020-11-13
CachingClusterReader should continue on all list errors

mortent:
Currently the CachingClusterReader bails out if it gets any other errors than `NotFound` when listing a specific GVK in a namespace. This is not correct, it should just store the error and return it whenever anyone requests a resource that should have been included in the result from the list call.


Liujingfang1:
Looks good to me. @seans3 Do you want to take a look?

seans3 (approved):
/lgtm

PR #281, opened 2020-11-17
update events to include indentifier and error

Liujingfang1:
This is part of the overall change for ContinueOnError. 
For each object being applied/pruned, the event stream should return an event for it. The event includes enough information for
- GKNN
- the object itself if available
- the error encountered if any

### DIFF
--- a/pkg/apply/event/event.go
+++ b/pkg/apply/event/event.go
@@ -4,7 +4,7 @@
 package event
 
 import (
-	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	pollevent "sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
 	"sigs.k8s.io/cli-utils/pkg/object"
 )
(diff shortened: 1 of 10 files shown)


pkg/apply/event/event.go
```diff
@@ -110,6 +113,7 @@ const (
 type StatusEvent struct {
 	Type     StatusEventType
 	Resource *pollevent.ResourceStatus
+	Error    error
```
mortent, line 116: Not sure if this is needed. The `*pollevent.ResourceStatus` already includes any errors. If we wanted to include the error directly on the `StatusEvent` instead, we should consider if we want to use a different type than `*pollevent.ResourceStatus` from the polling library.

Liujingfang1: I wasn't aware of that. Then we don't need to include the error here.

Liujingfang1: fixed

mortent:
/lgtm
As a follow-up we should also remember to update the different printers so they can provide users with useful information whenever apply/prune/delete fails.

Liujingfang1:
@mortent  Created https://github.com/kubernetes-sigs/cli-utils/issues/282 for the follow up request. We can work on it after the continue-on-error change.

PR #283, opened 2020-11-17
change prune to be continue on error

Liujingfang1:
This is part of the continue-on-error change. 
- It continues to prune the next object when the current one failed. The failure is sent to a PruneEvent
- The skipped resources are also sent to some PruneEvent
- If a resource is not pruned, keep it in the inventory list

Also added some unit tests for the change.


seans3 (approved):
/lgtm

PR #284, opened 2020-11-18
Adds end-to-end test for pruning and namespaces

seans3:
* Adds end-to-end test that shows namespaces are not pruned when objects are still in them.
* Fixes regression where namespaces where incorrectly pruned when objects were still in them.

### DIFF
--- a/examples/alphaTestExamples/pruneNamespace.md
+++ b/examples/alphaTestExamples/pruneNamespace.md
@@ -1,0 +1,184 @@
+[kind]: https://github.com/kubernetes-sigs/kind
+
+# Demo: Namespaces and Prune
+
+This demo shows that namespaces will **not** be pruned if
+there are objects remaining in them. The namespace for the
+inventory object will be the default namespace, since not
+all of the app objects are in the same namespace (pod-b
+is in the default namespace). When pod-a, pod-b, and
+the test-namespace are omitted from the subsequent apply
+the test-namespace will be considered for pruning, but it
+will **not** happend because there is still one object
+in the namespace--pod-c.
+
+First define a place to work:
+
+<!-- @makeWorkplace @testE2EAgainstLatestRelease -->
+```
+DEMO_HOME=$(mktemp -d)
+```
+
+Alternatively, use
+
+> ```
+> DEMO_HOME=~/demo
+> ```
+
+## Establish the base
+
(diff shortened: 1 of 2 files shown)


examples/alphaTestExamples/pruneNamespace.md
```diff
@@ -0,0 +1,184 @@
...
+expectedOutputLine "configmap/cm-a pruned"
+expectedOutputLine "configmap/cm-b pruned"
+expectedOutputLine "configmap/cm-c unchanged"
+expectedOutputLine "2 resource(s) pruned, 1 skipped"
+
+# The test-namespace should not be pruned.
+kubectl get ns test-namespace --no-headers | wc -l > $OUTPUT/status
+expectedOutputLine "1"
+# There should be only one inventory object
+kubectl get cm --selector='cli-utils.sigs.k8s.io/inventory-id' --no-headers | wc -l > $OUTPUT/status
```
Liujingfang1, line 176: I think we also need to verify the namespace "test-namespace" is still in the inventory list.

examples/alphaTestExamples/pruneNamespace.md
```diff
@@ -0,0 +1,184 @@
...
+
+<!-- @makeWorkplace @testE2EAgainstLatestRelease -->
+```
+DEMO_HOME=$(mktemp -d)
+```
+
+Alternatively, use
+
+> ```
+> DEMO_HOME=~/hello
```
Liujingfang1, line 25:
```
DEMO_HOME=~/demo
```

seans3: Fixed.

examples/alphaTestExamples/pruneNamespace.md
seans3: No. Currently any resource that is skipped in the prune is removed from the inventory. So the `test-namespace` will not be in the inventory list. 

It sounds like we want to change this (including for resources that are prune-skipped because of the `on-remove: keep` annotation). But that will happen in a future PR.

Liujingfang1: I changed that in #283, the skipped namespace and the skipped resource due to lifecycle annotation are both added to the inventory list.

seans3: Ok--nice. Let me try to update the e2e test.

PR #286, opened 2020-11-18
change apply to continue on error

Liujingfang1:
- Update the apply task to not return error so that it can always proceed to the next apply task or the prune task.
-  Change the ApplyOptions.Run to multiple calls. Each time only one object is applied. This change is to have better control of the return error, which will be consumed by ConfigSync. So ConfigSync will know

   - if a single resource is applied successfully
   - if the apply failure is due to an unknown type

- Add a unit test for applying two resources. One of them failed, we should get the expected event.

### DIFF
--- a/pkg/apply/error/error.go
+++ b/pkg/apply/error/error.go
@@ -1,0 +1,39 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+package error
+
+type UnknwonTypeError struct {
+	err error
+}
+
+func (e *UnknwonTypeError) Error() string {
+	return e.err.Error()
+}
+
+func NewUnknwonTypeError(err error) *UnknwonTypeError {
+	return &UnknwonTypeError{err: err}
+}
+
+type ApplyRunError struct {
+	err error
+}
+
+func (e *ApplyRunError) Error() string {
+	return e.err.Error()
+}
+
+func NewApplyRunError(err error) *ApplyRunError {
+	return &ApplyRunError{err: err}
+}
+
+type InitializeApplyOptionError struct {
+	err error
+}
+
+func (e *InitializeApplyOptionError) Error() string {
+	return e.err.Error()
+}
+
+func NewInitializeApplyOptionError(err error) *InitializeApplyOptionError {
+	return &InitializeApplyOptionError{err: err}
+}
--- a/pkg/apply/error/error.go
+++ b/pkg/apply/error/error.go
@@ -2,16 +2,16 @@
 // SPDX-License-Identifier: Apache-2.0
 package error
 
-type UnknwonTypeError struct {
+type UnknownTypeError struct {
 	err error
 }
 
-func (e *UnknwonTypeError) Error() string {
+func (e *UnknownTypeError) Error() string {
 	return e.err.Error()
 }
 
-func NewUnknwonTypeError(err error) *UnknwonTypeError {
-	return &UnknwonTypeError{err: err}
+func NewUnknownTypeError(err error) *UnknownTypeError {
+	return &UnknownTypeError{err: err}
 }
 
 type ApplyRunError struct {
--- a/pkg/apply/info/info_helper.go
+++ b/pkg/apply/info/info_helper.go
@@ -16,9 +16,9 @@ import (
 // InfoHelper provides functions for interacting with Info objects.
 type InfoHelper interface {
 	// UpdateInfo sets the mapping and client for the provided Info
-	// objects. This must be called at a time when all needed resource
+	// object. This must be called at a time when all needed resource
 	// types are available in the RESTMapper.
-	UpdateInfo(infos *resource.Info) error
+	UpdateInfo(info *resource.Info) error
 
 	BuildInfo(obj *unstructured.Unstructured) (*resource.Info, error)
 }
--- a/pkg/apply/info/info_helper.go
+++ b/pkg/apply/info/info_helper.go
@@ -15,12 +15,12 @@ import (
 
 // InfoHelper provides functions for interacting with Info objects.
 type InfoHelper interface {
-	// UpdateInfos sets the mapping and client for the provided Info
+	// UpdateInfo sets the mapping and client for the provided Info
 	// objects. This must be called at a time when all needed resource
 	// types are available in the RESTMapper.
-	UpdateInfos(infos []*resource.Info) error
+	UpdateInfo(infos *resource.Info) error
 
-	BuildInfos(objs []*unstructured.Unstructured) ([]*resource.Info, error)
+	BuildInfo(obj *unstructured.Unstructured) (*resource.Info, error)
 }
 
 func NewInfoHelper(factory util.Factory) *infoHelper {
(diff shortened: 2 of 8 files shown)


pkg/apply/error/error.go
```diff
@@ -0,0 +1,39 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+package error
+
+type UnknwonTypeError struct {
```
mortent, line 5: I think there is a typo here. It should be `UnknownTypeError` right?

pkg/apply/info/info_helper.go
```diff
@@ -15,12 +15,12 @@ import (
 
 // InfoHelper provides functions for interacting with Info objects.
 type InfoHelper interface {
-	// UpdateInfos sets the mapping and client for the provided Info
+	// UpdateInfo sets the mapping and client for the provided Info
 	// objects. This must be called at a time when all needed resource
 	// types are available in the RESTMapper.
-	UpdateInfos(infos []*resource.Info) error
+	UpdateInfo(infos *resource.Info) error
```
mortent, line 21: We should change the parameter here to singular.

pkg/apply/task/apply_task.go
```diff
@@ -114,15 +108,31 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
 		ao, err := applyOptionsFactoryFunc(taskContext.EventChannel(),
 			a.ServerSideOptions, a.DryRunStrategy, a.Factory)
 		if err != nil {
-			a.sendTaskResult(taskContext, err)
+			taskContext.EventChannel() <- createApplyFailedEvent(object.ObjMetadata{},
```
mortent, line 111: I think this could be tricky for clients. Is the thinking that they should check the type of the error to determine whether the error refers to a specific resource?

pkg/apply/task/apply_task.go
```diff
@@ -187,9 +197,9 @@ func newApplyOptions(eventChannel chan event.Event, serverSideOptions common.Ser
 	}, nil
 }
 
-func (a *ApplyTask) sendTaskResult(taskContext *taskrunner.TaskContext, err error) {
+func (a *ApplyTask) sendTaskResult(taskContext *taskrunner.TaskContext) {
 	taskContext.TaskChannel() <- taskrunner.TaskResult{
-		Err: err,
+		Err: nil,
```
mortent, line 202: We should just avoid setting the `Err` property here instead of setting it to `nil`.

mortent (commented):
Some small comments. In general we should think about how we handle errors that are specific a resource vs errors where we are unable to set up the infrastructure. Should we continue on absolutely all errors and how would the client handle the different situations?

pkg/apply/task/apply_task.go
Liujingfang1: Yes, the client needs to check the type of error. 
When receiving an ApplyFailedEvent, it checks the error type inside it.
- `InitializeApplyOptionError` means the it fails to initialize the ApplyOptions, thus all resources in this task are failed to be applied.
- `UnknownTypeError` means that it fails to apply a single resource because the type is not recognized, for example, a CR whose CRD hasn't been registered.
- `ApplyRunError` means that it fails to apply a single resource.

pkg/apply/error/error.go
Liujingfang1: fixed.

pkg/apply/task/apply_task.go
Liujingfang1: fixed

pkg/apply/info/info_helper.go
Liujingfang1: fixed

pkg/apply/task/apply_task.go
Liujingfang1: > Should we continue on absolutely all errors and how would the client handle the different situations?

The client side needs to know which resources are failed in apply due to an unknown type. If we apply all resources in one `ApplyOption.Run()`, it returns all error messages as a single error, like the format `[error when ***, error when ***, error when ***]`. However, after "error when", the message is unstructured. It is not feasible for the client side to parse this error and determine if it is an unknown type or some other apply error. By applying resources individually and return different error types, the client side can rely on the error type to determine the failure type.

mortent: I agree that unknown types should be reported as an error for the particular resource. 
I'm a bit concerned about requiring users to check the type of an error in order to determine if other fields in an event are populated. For example, in our cli ux we need to match errors to resources based on the `object.ObjMetadata` information. But I don't think the client should be required to enumerate all possible errors and what they mean. So I think we should consider designing our event hierarchy so the user can check on either the type of an event or an enum in an event to determine if it pertains to a specific resource.

Liujingfang1:
@seans3  I compared the number of requests for two kapply binaries with `kapply apply -v=6 <DIR>`. Once binary is built from this PR, the other binary is built from the head of the master branch. The applied package contains a CRD, a Namespace, two Deployments and a Service. They have the same number of https:// requests.

pkg/apply/task/apply_task.go
Liujingfang1: That's a good point. The current approach assumes the consumer to use both the `Type` and `Error` in the event. This can be improved by only using the `Type`. How about change the ApplyEventType to the following
```
const (
	ApplyEventResourceUpdate ApplyEventType = iota.   # for a single resource successfully applied
	ApplyEventResourceFailedUnknownType         # for single resource failed due to unknown type
	ApplyEventResourceFailed                 # for single resource failed apply
	ApplyEventAllCompleted                   # for all resources have been applied. This is sent out at the end of the apply loop
	ApplyEventAllFailed                      # for failed to build the apply infra
)
```

seans3 (approved):
Looks good to me. Once Morten's concerns are addressed we can merge.

pkg/apply/task/apply_task.go
mortent: So I don't think we need any additional `ApplyEventResource*` event types. Whenever this event comes in, clients should assume that the `Identifier` is set, so it will be possible to identify which resource it pertains to. In order to find out if there was an error, I think we have two options. We can say that users should check if the `Error` property is different from `nil`, or we can add an additional `error` entry to `ApplyEventOperation`. I think it is reasonable that users should just check whether `Error` is nil.

I think we should just keep the `ApplyEventCompleted` event. The semantics here is that we have completed applying the resources in a particular task, but it doesn't say anything about whether we were successful. 

What remains is then how we should handle errors that can't necessarily be attached to a specific resource. I think there are three options here:
 * Just return a `ApplyEventFailed` event and leave it up to the client to figure out which resources were not applied as a result of this error. This event would then be followed by a `ApplyEventCompleted` event that signals that the task has completed.
 * Iterative over all the resources passed into the apply_task and return a `ApplyEventResourceUpdate` event for each with the error. This is somewhat compelling since the client doesn't need to worry about different "types" of apply errors.
 * Return an `ApplyEventFailed` event and include a slice of `object.ObjMetadata` that covers all the resources. This seems more complicated for the client than the previous one.
I probably lean towards the second if we think we can do it. We just need to make sure we always have information about the resources that should be applied.

A separate issue that just came to mind is how we should handle `Wait` tasks if one or more resources fail to apply. It might lead us to wait for a condition we know will never happen.

Liujingfang1: Agreed that with option2, the successful apply and failed apply events will be consistent. That makes the client side easier to consume those events. Option 3 is a little improvement compare with Option 1, but both of them need client sides to write two consuming logic.

For the `Wait` tasks, we need a way to skip waiting for the failed task. We could tell the runner not to start the wait task or exclude the failed resource from the wait task identifiers. I may need some time to figure out the best way for doing this, but it will be a different change.

Liujingfang1: Updated the change. PTAL

Also filed https://github.com/kubernetes-sigs/cli-utils/issues/290 for the new requirement of `Wait` task.

PR #289, opened 2020-11-19
add inventory policy

Liujingfang1:
This change adds the option InventoryPolicy. This policy controls the behavior of apply/prune behavior. This control is done by checking if an apply/prune operation can be taken for an object with a given InventoryInfo. 

This change covers the part that if an operation can go through. In the next change, this check will be applied in the Applier.


pkg/inventory/policy.go
```diff
@@ -0,0 +1,118 @@
...
+const owningInventoryKey = "config.kubernetes.io/owning-inventory"
+
+// inventoryIDMatchStatus represents the result of comparing the
+// id from current inventory info and the inventory-id from a live object.
+type inventoryIDMatchStatus int
+
+const (
+	Empty inventoryIDMatchStatus = iota
+	Match
+	Unmatch
```
mortent, line 73: I think `NoMatch` might be a better name here?

Liujingfang1: updated

pkg/inventory/policy.go
```diff
@@ -0,0 +1,118 @@
...
+	// The apply operation can go through for any resource in the package even if the
+	// live object has an unmatched inventory-id annotation.
+	//
+	// The prune operation can go through when
+	// - The inventory-id annotation in the live object match with that
+	//   in the package.
+	AdoptAll
+)
+
+const owningInventoryKey = "config.kubernetes.io/owning-inventory"
```
mortent, line 64: We should revisit the name of this annotation, ref https://github.com/GoogleContainerTools/kpt/issues/1257

Liujingfang1: changed it to "config.k8s.io/owning-inventory"

Liujingfang1:
Addressed the comments. It's ready for another review.

mortent (approved):
/lgtm

PR #291, opened 2020-12-01
enable InventoryPolicy in applier

Liujingfang1:
This change enables the InventoryPolicy in applier. 

For each resource in the ApplyTask, it first check if the resource exists in the cluster. It then checks if the resource can be applied with the current inventory info. If it can, apply the resource; otherwise, send an event with the proper error.

Also updated the unit tests for ApplyTask and PruneTask.

It should be merged after #289

### DIFF
--- a/pkg/apply/prune/prune.go
+++ b/pkg/apply/prune/prune.go
@@ -144,8 +144,7 @@ func (po *PruneOptions) Prune(localInv inventory.InventoryInfo, localObjs []*uns
 			continue
 		}
 		// Handle lifecycle directive preventing deletion.
-		if preventDeleteAnnotation(metadata.GetAnnotations()) {
-			klog.V(4).Infof("prune object lifecycle directive; do not prune: %s", uid)
+		if !canPrune(localInv, obj, o.InventoryPolicy, uid) {
 			eventChannel <- createPruneEvent(clusterObj, obj, event.PruneSkipped)
 			localIds = append(localIds, clusterObj)
 			continue
@@ -160,13 +159,6 @@ func (po *PruneOptions) Prune(localInv inventory.InventoryInfo, localObjs []*uns
 				localIds = append(localIds, clusterObj)
 				continue
 			}
-		}
-		if !inventory.CanPrune(localInv, obj, o.InventoryPolicy) {
-			klog.V(4).Infof("skip pruning object that doesn't belong to current inventory: %s/%s",
-				clusterObj.Namespace, clusterObj.Name)
-			eventChannel <- createPruneEvent(clusterObj, obj, event.PruneSkipped)
-			localIds = append(localIds, clusterObj)
-			continue
 		}
 		if !o.DryRunStrategy.ClientOrServerDryRun() {
 			klog.V(4).Infof("prune object delete: %s/%s", clusterObj.Namespace, clusterObj.Name)
@@ -230,4 +222,17 @@ func createPruneFailedEvent(objMeta object.ObjMetadata, err error) event.Event {
 			Error:      err,
 		},
 	}
+}
+
+func canPrune(localInv inventory.InventoryInfo, obj *unstructured.Unstructured,
+	policy inventory.InventoryPolicy, uid string) bool {
+	if !inventory.CanPrune(localInv, obj, policy) {
+		klog.V(4).Infof("skip pruning object that doesn't belong to current inventory: %s", uid)
+		return false
+	}
+	if preventDeleteAnnotation(obj.GetAnnotations()) {
+		klog.V(4).Infof("prune object lifecycle directive; do not prune: %s", uid)
+		return false
+	}
+	return true
 }
--- a/pkg/apply/prune/prune.go
+++ b/pkg/apply/prune/prune.go
@@ -161,6 +161,13 @@ func (po *PruneOptions) Prune(localInv inventory.InventoryInfo, localObjs []*uns
 				continue
 			}
 		}
+		if !inventory.CanPrune(localInv, obj, o.InventoryPolicy) {
+			klog.V(4).Infof("skip pruning object that doesn't belong to current inventory: %s/%s",
+				clusterObj.Namespace, clusterObj.Name)
+			eventChannel <- createPruneEvent(clusterObj, obj, event.PruneSkipped)
+			localIds = append(localIds, clusterObj)
+			continue
+		}
 		if !o.DryRunStrategy.ClientOrServerDryRun() {
 			klog.V(4).Infof("prune object delete: %s/%s", clusterObj.Namespace, clusterObj.Name)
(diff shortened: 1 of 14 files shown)


pkg/apply/task/apply_task.go
```diff
@@ -116,6 +127,22 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
 					object.UnstructuredToObjMeta(obj), event.Unchanged, applyerror.NewUnknownTypeError(err))
 				continue
 			}
+
+			clusterObj, err := getClusterObj(dynamic, info)
+			if err != nil {
+				if !apierrors.IsNotFound(err) {
+					fmt.Println("I'm here in line 134")
```
mortent, line 134: We probably want to remove this line.

pkg/apply/task/apply_task.go
```diff
@@ -98,7 +109,7 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
 
 		// Create a new instance of the applyOptions interface and use it
 		// to apply the objects.
-		ao, err := applyOptionsFactoryFunc(taskContext.EventChannel(),
+		ao, dynamic, err := applyOptionsFactoryFunc(taskContext.EventChannel(),
```
mortent, line 112: Maybe we should do this in two steps, so we first create the dynamic client and then pass that into `applyOptionsFactoryFunc`?

pkg/apply/task/apply_task.go
```diff
@@ -116,6 +127,22 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
 					object.UnstructuredToObjMeta(obj), event.Unchanged, applyerror.NewUnknownTypeError(err))
 				continue
 			}
+
+			clusterObj, err := getClusterObj(dynamic, info)
```
mortent, line 131: It is a bit unfortunate that we need this, but I'm not sure if there is a way to avoid it for now. The implementation of `ApplyOptions.Run` also does a get, so we end up fetching every object twice.

pkg/apply/task/apply_task.go
```diff
@@ -116,6 +127,22 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
...
+			if err != nil {
+				if !apierrors.IsNotFound(err) {
+					fmt.Println("I'm here in line 134")
+					if !inventory.CanApply(a.InvInfo, clusterObj, a.InventoryPolicy) {
+						taskContext.EventChannel() <- createApplyEvent(
+							object.UnstructuredToObjMeta(obj),
+							event.Unchanged,
+							applyerror.NewCannotApplyError(fmt.Errorf("can't apply the resource since it belongs to another inventory object")))
```
mortent, line 139: Is this error message always correct? If the policy is `InventoryPolicyMustMatch`, apply can be blocked simply because the resource already exist, even if it doesn't belong to another inventory.

pkg/apply/taskrunner/runner.go
```diff
@@ -42,6 +43,7 @@ type Options struct {
 	PollInterval     time.Duration
 	UseCache         bool
 	EmitStatusEvents bool
+	InventoryPolicy  inventory.InventoryPolicy
```
mortent, line 46: Why is the inventory policy required here?

pkg/apply/task/apply_task.go
Liujingfang1: Thanks for catching this : ).  Will remove it.

pkg/apply/task/apply_task.go
Liujingfang1: That's true. We can let `CanApply` return a message for why the resource can't be applied. Then this error will capture that message.

pkg/apply/task/apply_task.go
Liujingfang1: We could avoid it by not using the kubectl `ApplyOptions`,  but it would require coping lots of kubectl code. Let's keep this implementation. If we see any user report performance issue, we can improve it then.

pkg/apply/task/apply_task.go
Liujingfang1: The `dynamic` is created from `util.Factory`, which is passed into `applyOptionsFactoryFunc`, but `dynamic` is not the only one needed to create an `ApplyOptions`. It needs a `discovery` client as well. Both `dynamic` and `discovery` are from `util.Factory`. It's better to put the code for getting both `dynamic` and `discovery` together.

Another approach is to passed both `dynamic` and `discovery` to the ApplyTask. Then the `util.Factory` is not needed. This may need a bit refactoring on current code and test. If you prefer this approach, we can do it in a different PR.

pkg/apply/taskrunner/runner.go
Liujingfang1: Not needed. Removed it.

Liujingfang1:
This is ready for another review. @seans3 @mortent

pkg/inventory/policy_test.go
```diff
@@ -157,7 +157,7 @@ func TestCanApply(t *testing.T) {
 		},
 	}
 	for _, tc := range testcases {
-		actual := CanApply(tc.inv, tc.obj, tc.policy)
+		actual, _ := CanApply(tc.inv, tc.obj, tc.policy)
```
mortent, line 160: Shouldn't we verify that the error is nil here?

mortent (commented):
Can we add a test or two for this in `pkg/apply/applier_test.go`? Those tests are a bit verbose currently, but since we have most of the planned changes to the API in place, it would be a good time to see if we can simplify it and add better test coverage. It should help us make sure all the pieces continue to work together as we make updates.

pkg/inventory/policy_test.go
Liujingfang1: Not necessary. We can to verify the `actual` is the same as expected. The returned error is not always nil.

pkg/apply/prune/prune.go
```diff
@@ -161,6 +161,13 @@ func (po *PruneOptions) Prune(localInv inventory.InventoryInfo, localObjs []*uns
 				continue
 			}
 		}
+		if !inventory.CanPrune(localInv, obj, o.InventoryPolicy) {
```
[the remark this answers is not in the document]
Liujingfang1: Yes, we can combine them. First, we check if the current object can be pruned by the `localInv`. If it can, then we can the lifecycle annotation.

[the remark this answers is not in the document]
Liujingfang1: Combined them into one function `canPrune`.

Liujingfang1:
@mortent  I added unit tests for the ApplyTask. I think that gives us a good coverage for what happens with respect to different combinations of the inventory policy and annotation value. This PR already has a large size. We can add more tests in `pkg/apply/applier_test.go` in a follow up PR.

mortent (approved):
/lgtm

PR #293, opened 2020-12-08
update printer for events

### LINKED ISSUE #282: update printer to use the new information added in apply/prune/deleteEvent
A follow up of https://github.com/kubernetes-sigs/cli-utils/pull/281, we can update the printer to give more information based on new fields in Apply/Prune/DeleteEvent.

Liujingfang1:
To fix #282.

After we add the continue on error in the applier, the event includes a new field `Identifier`. This field will be always assigned during applying/pruning. The existing field `Object` is sometimes nil. So the printer should relay on `Identifier` instead of the `Object` for ApplyEvent/PruneEvent/DeleteEvent.

Another change is that we added a `failedCount` in the printer for apply/prune event. Here is an example of the output
```
$ kapply apply .
namespace/test-namespace created
service/helloworld-gke created
deployment.apps/helloworld-gke1 failed
deployment.apps/helloworld-gke2 created
4 resource(s) applied. 3 created, 0 unchanged, 0 configured, 1 failed
0 resource(s) pruned, 0 skipped, 0 failed
```

### DIFF
--- a/cmd/printers/events/formatter.go
+++ b/cmd/printers/events/formatter.go
@@ -32,8 +32,8 @@ type formatter struct {
 func (ef *formatter) FormatApplyEvent(ae event.ApplyEvent, as *list.ApplyStats, c list.Collector) error {
 	switch ae.Type {
 	case event.ApplyEventCompleted:
-		output := fmt.Sprintf("%d resource(s) applied. %d created, %d unchanged, %d configured",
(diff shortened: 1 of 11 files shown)


seans3 (approved):
This looks good to me. I'll let Morten review before LGTM'ing.

mortent (approved):
/lgtm

PR #295, opened 2020-12-10
add test cases for applier

Liujingfang1:
Add three test cases for applying and pruning with matched/unmatched annotation


seans3 (approved):
/lgtm

PR #297, opened 2020-12-15
ingore the failed resources in wait task

### LINKED ISSUE #290: skip wait tasks for resources that fails during apply/prune
Since the apply process is changed to continue on error. We should also update the wait tasks. For any resources that are failed during apply/prune, they should be skipped in the wait tasks. Otherwise, the tasks would wait for conditions that are not going to happen.

Liujingfang1:
fix #290

### DIFF
--- a/pkg/apply/applier_test.go
+++ b/pkg/apply/applier_test.go
@@ -265,20 +265,6 @@ func TestApplier(t *testing.T) {
 						Status: status.CurrentStatus,
 					},
 				},
-				{
-					EventType: pollevent.ResourceUpdateEvent,
-					Resource: &pollevent.ResourceStatus{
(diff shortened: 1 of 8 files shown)


mortent (approved):
Looks good to me. I'll let @seans3 take a look too.

seans3:
/hold

...regression in ResourceGroup pruning. We'll need to fix that first.

Liujingfang1:
@seans3 Is the regression from this PR?

seans3:
> Sean Sullivan Is the regression from this PR?

This PR broke pruning: https://github.com/kubernetes-sigs/cli-utils/pull/291

Let's not change anything until the pruning is fixed.

seans3 (approved):
/lgtm

PR #301, opened 2020-12-21
Improve the Applier tests

mortent:
This updates the `Applier` tests to (mostly) avoid using mocks of `Applier` components, but instead rely on the client stubs. With this change, the handling of the inventory object is part of the tests.

With the exception of changing a field to be published, it only contains changes to the test code.

It also adds additional tests.

### DIFF
--- a/pkg/apply/applier_test.go
+++ b/pkg/apply/applier_test.go
@@ -9,14 +9,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"path"
 	"reflect"
 	"regexp"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
-	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/apply/applier_test.go
```diff
@@ -303,28 +451,40 @@ func TestApplier(t *testing.T) {
 				},
 				{
 					eventType:       event.StatusType,
-					statusEventType: event.StatusEventResourceUpdate,
+					statusEventType: event.StatusEventCompleted,
 				},
 				{
-					eventType:       event.StatusType,
-					statusEventType: event.StatusEventCompleted,
+					// TODO(mortent): This doesn't seem correct. We shouldn't try
```
Liujingfang1, line 457: In this case, we didn't try to prune this resource. We skip it. Maybe the `event.PruneSkipped` is confusing to you, we can change it to a more meaningful operation such as `event.UnPrune`.

Liujingfang1 (commented):
Overall looks good to me.

pkg/apply/applier_test.go
mortent: Yeah, so I realize we don't actually end up pruning it, but it just seems like in this case the resource shouldn't even be considered for pruning. My thought is that during apply, if a resource is part of the set of resources provided to the apply-task, we should include its `uid` in the `taskContext` so the resource during pruning is treated as if it actually was applied. WDYT?

Liujingfang1: I agree that that the `uid` should be in included in the `taskContext`. This is some misbehavior introduced after adding the `canApply` in the ApplyTask. I think moving line https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/apply/task/apply_task.go#L150 ahead of the triggering of `CanApply` function can fix this.

Liujingfang1: https://github.com/kubernetes-sigs/cli-utils/pull/303 is to fix the issue.

seans3 (approved):
/lgtm

seans3 (approved):
/lgtm

PR #302, opened 2020-12-21
allow pruning object without owning-inventory annotation

Liujingfang1:
We need to allow pruning objects which don't have the owning-inventory annotations. Those objects are from a previous apply using an old version of the apply library, where the `owning-inventory` annotation wasn't added. When the package is applied by the new version of the apply library, some resources are adopted(assume this is available). Then the resources that are removed from the local package don't have the chance to get adopted. The existing code is not able to prune those resources.

/assign @mortent @seans3


pkg/inventory/policy.go
```diff
@@ -119,7 +122,7 @@ func CanPrune(inv InventoryInfo, obj *unstructured.Unstructured, policy Inventor
 	matchStatus := inventoryIDMatch(inv, obj)
 	switch matchStatus {
 	case Empty:
-		return false
+		return true
```
mortent, line 125: So should we maybe use the inventory policy here to determine what we should do?

Liujingfang1: We can use the inventory policy here. Among the three policies, the `AdoptAll` and `AdoptIfNoInventory` allows applying in this case, so we can also allow pruning for them. 

The motivation of this PR is that we need to think about how to handle the first apply after the cli is switched to the new library version with inventory policy. If the cli runs `AdoptIfNoInventory` for the first apply, then all resources can be either applied with owning-inventory annotation or be pruned. The following apply with `InventoryPolicyMustMatch` would work.

Liujingfang1: Updated this in the 2nd commit.

pkg/inventory/policy.go
```diff
@@ -119,7 +121,7 @@ func CanPrune(inv InventoryInfo, obj *unstructured.Unstructured, policy Inventor
 	matchStatus := inventoryIDMatch(inv, obj)
 	switch matchStatus {
 	case Empty:
-		return false
+		return policy == AdoptIfNoInventory || policy == AdoptAll
```
mortent, line 124: This looks good, but should we also consider the `NoMatch` case? It seems to me that if the user provided the `AdoptAll` value, then we should maybe allow pruning in this case. In general, should we treat pruning the same as adoption, so basically if a resource can be updated (and therefore adopted), should we also allow it to be pruned?

Liujingfang1: That's a good point. I'm thinking how we could encounter the case that we need to prune a resource with `AdoptAll` policy and `NoMatch`. Since `AdoptAll` won't be exposed to the cli, so we won't encounter this case through cli. It could only happen for the library consumers.

When the library is called with `AdoptAll` policy and the inventory object is updated. However, a resource is failed to be applied. So the `owning-inventory` annotation doesn't match with the inventory object. Later this resource is removed from the config. So it needs to be pruned in the re-apply. Then we encounter the case mentioned above. In this case, it makes sense to allow pruning the resource.

So I think we should allow the pruning. The resource to be pruned can be viewed as a failed apply + prune.

Liujingfang1: updated.

seans3 (approved):
/lgtm

PR #303, opened 2020-12-22
add ids in the apply task into taskContext

Liujingfang1:
The existing code doesn't add the id into the taskContext when the resource has a different owning-inventory. As a result, the resource was included in the prune task incorrectly. This change is to fix that.

/assign @mortent


mortent:
Looks good, but can we add a test?

Liujingfang1:
Updated the existing test to cover the change.

mortent (approved):
/lgtm

PR #310, opened 2020-12-31
Fix missing identifier delete event

mortent:
This addresses an issue where the identifier was not copied over from the `PruneEvent` to the `DeleteEvent`.


seans3 (approved):
/lgtm

PR #314, opened 2021-01-14
Fixes panic when id not found in collector

seans3:
* Fixes panic in `ResourceStateCollector.processStatusEvent` caused by de-referencing a field from a nil pointer.
* Fixes: https://github.com/GoogleContainerTools/kpt/issues/1329
* Adds unit tests for `collector.go` increasing unit test coverage from 16.1% to 23.7%
* Adds logging messages for debugging

### DIFF
--- a/cmd/printers/table/collector.go
+++ b/cmd/printers/table/collector.go
@@ -7,6 +7,7 @@ import (
 	"sort"
 	"sync"
 
+	"k8s.io/klog"
 	"sigs.k8s.io/cli-utils/pkg/apply/event"
 	pe "sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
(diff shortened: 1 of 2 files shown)


cmd/printers/table/collector.go
```diff
@@ -182,8 +183,17 @@ func (r *ResourceStateCollector) processEvent(ev event.Event) error {
 // update for a resource.
 func (r *ResourceStateCollector) processStatusEvent(e event.StatusEvent) {
 	if e.Type == event.StatusEventResourceUpdate {
+		klog.V(7).Infoln("processing status event")
 		resource := e.Resource
-		previous := r.resourceInfos[resource.Identifier]
+		if resource == nil {
```
mortent, line 188: So making sure it doesn't crash is good, but do we know why this happened in the first place? The idea is that the `StatusEventResourceUpdate` is about a specific resource, so why doesn't the event include the resource?

cmd/printers/table/collector.go
```diff
@@ -182,8 +183,17 @@ func (r *ResourceStateCollector) processEvent(ev event.Event) error {
...
+			klog.V(4).Infoln("status event missing Resource field; no processing")
+			return
+		}
+		previous, found := r.resourceInfos[resource.Identifier]
+		if !found {
```
mortent, line 193: Similar to above, but this seems like a situation that is a bug. If I remember correctly the `resourceInfos` map should be pre-populated with all resources which we are interested in. So if we get an event about a different resource, something must be wrong elsewhere.

cmd/printers/table/collector.go
seans3: I don't think this is happening in this bug. But we should program defensively.

cmd/printers/table/collector.go
seans3: This probably happens when a `PersistentVolume` status hangs, but I haven't been able to reproduce this yet. Please have a look at the connected issue, which points out this `previous` is nil and not found.

PR #319, opened 2021-01-26
Fix for final inventory update

seans3:
* Updates the `Prune` function to correctly store the final inventory items. These items are:
  1) the successfully applied objects
  2) the prune failures
* Fixes error where unsuccessfully applied objects were included in the final inventory.
* Adds to `Prune` unit test to validate the final inventory stored is correct.
* Small clean-up for `TaskContext` in `apply_task` to ensure only successfully applied objects are added to context.
* Many new debug logging statements.

### DIFF
--- a/pkg/apply/prune/prune.go
+++ b/pkg/apply/prune/prune.go
@@ -83,113 +83,173 @@ type Options struct {
 }
 
 // Prune deletes the set of resources which were previously applied
-// (retrieved from previous inventory objects) but omitted in
-// the current apply. Prune also delete all previous inventory
-// objects. Returns an error if there was a problem.
-func (po *PruneOptions) Prune(localInv inventory.InventoryInfo, localObjs []*unstructured.Unstructured, currentUIDs sets.String,
-	taskContext *taskrunner.TaskContext, o Options) error {
+// but omitted in the current apply. Calculates the set of objects
+// to prune by removing the currently applied objects from the union
+// set of the previously applied objects and currently applied objects
+// stored in the cluster inventory. As a final step, stores the current
+// inventory which is all the successfully applied objects and the
+// prune failures. Does not stop when encountering prune failures.
+// Returns an error for unrecoverable errors.
+//
+// Parameters:
+//   localInv - locally read inventory object
+//   localObjs - locally read, currently applied (attempted) objects
+//   currentUIDs - UIDs for successfully applied objects
+//   taskContext - task for apply/prune
+func (po *PruneOptions) Prune(localInv inventory.InventoryInfo,
+	localObjs []*unstructured.Unstructured,
+	currentUIDs sets.String,
+	taskContext *taskrunner.TaskContext,
+	o Options) error {
+	// Validate parameters
 	if localInv == nil {
 		return fmt.Errorf("the local inventory object can't be nil")
 	}
-	invNamespace := localInv.Namespace()
-	klog.V(4).Infof("prune local inventory object: %s/%s", invNamespace, localInv.Name())
 	// Get the list of Object Meta from the local objects.
 	localIds := object.UnstructuredsToObjMetas(localObjs)
 	// Create the set of namespaces for currently (locally) applied objects, including
 	// the namespace the inventory object lives in (if it's not cluster-scoped). When
 	// pruning, check this set of namespaces to ensure these namespaces are not deleted.
-	localNamespaces := mergeObjNamespaces(localObjs)
-	if invNamespace != "" {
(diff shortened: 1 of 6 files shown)


pkg/apply/prune/prune_test.go
```diff
@@ -259,35 +265,57 @@ func TestPrune(t *testing.T) {
...
 					}
-				} else if err == nil {
-					t.Fatalf("Expected error during Prune() but received none")
 				}
 			})
 		}
 	}
 }
 
+// unionObjects returns the union of sliceA and sliceB as a slice of unstructured objects.
```
Liujingfang1, line 301: If both sliceA and sliceB contains the unstructured object with the same ObjMeta, the one from sliceB is taken.

pkg/apply/prune/prune.go
```diff
@@ -83,114 +83,174 @@ type Options struct {
...
+			klog.V(5).Infof("prune object in current apply; do not prune: %s", uid)
 			continue
 		}
 		// Handle lifecycle directive preventing deletion.
 		if !canPrune(localInv, obj, o.InventoryPolicy, uid) {
-			taskContext.EventChannel() <- createPruneEvent(clusterObj, obj, event.PruneSkipped)
-			localIds = append(localIds, clusterObj)
+			klog.V(4).Infof("skip prune for lifecycle directive %s/%s", pruneObj.Namespace, pruneObj.Name)
+			taskContext.EventChannel() <- createPruneEvent(pruneObj, obj, event.PruneSkipped)
+			pruneFailures = append(pruneFailures, pruneObj)
```
Liujingfang1, line 157: Maybe add a comment here for why adding it to `pruneFailures`. It's actually not a failure, we want to use `pruneFailures` to keep it in the final inventory list.

pkg/apply/prune/prune.go
```diff
@@ -83,114 +83,174 @@ type Options struct {
...
+				if klog.V(4) {
+					klog.Errorf("prune failed for %s/%s (%s)", pruneObj.Namespace, pruneObj.Name, err)
+				}
+				taskContext.EventChannel() <- createPruneFailedEvent(pruneObj, err)
+				pruneFailures = append(pruneFailures, pruneObj)
+				taskContext.CaptureResourceFailure(pruneObj)
 				continue
 			}
+			err = namespacedClient.Delete(context.TODO(), pruneObj.Name, metav1.DeleteOptions{})
+			if err != nil {
```
Liujingfang1, line 185: The error handling here is the same as the one after `namespacedClient, err := po.namespacedClient(pruneObj)`. We can combine the getting client and deleting the object into one function, like what you have done for `getObject`. The new function can be `deleteObject`.

Liujingfang1 (commented):
Overall looks good to me.

mortent:
So overall looks good, but I want to understand how we decide whether a resource that we failed to apply should be included in the inventory or not. I want to make sure that a situation where a resource already exists in the cluster and apply (which in this case would be an update/patch) fails, we don't end up orphaning the resource. I don't think this were among the testcases.

seans3:
> So overall looks good, but I want to understand how we decide whether a resource that we failed to apply should be included in the inventory or not. I want to make sure that a situation where a resource already exists in the cluster and apply (which in this case would be an update/patch) fails, we don't end up orphaning the resource. I don't think this were among the testcases.


We had an informative discussion about this corner case, and we believe we know a way forward to make the final inventory storage step even more error tolerant. This PR is strictly better than before, so we will merge this current PR. The improvements to the algorithm will be instituted in a future PR. The following is documentation of our discussion to further improve the final inventory calculation.

The apply operation currently does a `get` before the `apply`. From these operations we believe we have the information necessary to pass to the `Step 4` final inventory (through the `TaskContext`) to improve its error tolerance. Currently, we are only storing the UID of successfully applied objects. We will enhance this data structure according to the following decision tree.
1. If we are able to successfully `get` the object, and the `apply` succeeds, the object remains in the final inventory.
1. If we are able to successfully `get` the object, but the `apply` fails, then we know it exists in the cluster, and it should therefore remain in the inventory.
2. If the `get` fails with a `Not Found` error, then we know the item is not in the cluster. This is considered an initial creation. If the `apply` succeeds for this initial creation, then it will be added to the final inventory. If the `apply` fails it will not be added to the final inventory because it is not in the cluster.
3. If the `get` fails with a different error, then we check if the item was in the previous inventory.
    a. If the item was in the previous inventory, then we assume it is still in the cluster and we keep it in the inventory
    b. If the item was not in the previous inventory, then we can confidently assume it is not in the cluster and we do not add it to the inventory.

PR #322, opened 2021-02-03
Better final inventory calculation during errors

seans3:
* More accurate final inventory set calculation when errors occur.
* Pass through the set of previous inventory objs from `applier` to `apply_task`.
* The previous inventory is used to determine if an object currently exists in the cluster. If it does, and an error occurs, we want to keep this object in the inventory.
* Updates the `apply_task.Start()` function to better determine inclusion/exclusion in final inventory set.
* Adds unit tests for `apply_task.Start()`

### DIFF
--- a/pkg/apply/task/apply_task.go
+++ b/pkg/apply/task/apply_task.go
@@ -17,7 +17,7 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/klog"
 	"k8s.io/kubectl/pkg/cmd/apply"
-	"k8s.io/kubectl/pkg/cmd/delete"
+	cmddelete "k8s.io/kubectl/pkg/cmd/delete"
 	"k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/util/slice"
 	applyerror "sigs.k8s.io/cli-utils/pkg/apply/error"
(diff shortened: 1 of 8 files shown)


pkg/apply/task/apply_task.go
```diff
@@ -144,12 +148,20 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
 			if err != nil {
 				if !apierrors.IsNotFound(err) {
 					if klog.V(4) {
-						klog.Errorf("error retrieving %s/%s from cluster--continue",
+						klog.Errorf("error (%s) retrieving %s/%s from cluster--continue",
+							err, info.Namespace, info.Name)
+					}
+					e := event.Failed
```
Liujingfang1, line 154: `e` -> `op` or `operation`

pkg/apply/task/apply_task.go
```diff
@@ -292,6 +308,38 @@ func (a *ApplyTask) filterCRsWithCRDInSet(objects []*unstructured.Unstructured)
 	return objs, objsWithCRD, nil
 }
 
+// objInCluster returns true if the passed object is in the slice of
+// previous inventory, because an object in the previous inventory
+// exists in the cluster.
+func (a *ApplyTask) objInCluster(obj object.ObjMetadata) bool {
+	for _, clusterObj := range a.PrevInventory {
```
Liujingfang1, line 315: I think you can change `a.PrevInventory` to a map, for example `map[object.ObjMetadata]struct{}`. Then this `for` loop can be dropped.

pkg/apply/task/apply_task.go
```diff
@@ -176,14 +188,21 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
 				if klog.V(4) {
 					klog.Errorf("error applying (%s/%s) %s", info.Namespace, info.Name, err)
 				}
+				// If apply failed and the object is not in the cluster, remove
+				// it from the final inventory.
+				if !a.objInCluster(id) {
+					klog.V(5).Infof("not in cluster; removing apply fail object %s/%s from inventory",
+						info.Namespace, info.Name)
+					infos = removeInfo(infos, info)
```
Liujingfang1, line 196: I think we can also optimize here by using a map for `infos`.

Liujingfang1:
/lgtm

Thanks for working on this!

PR #327, opened 2021-02-18
Fix printing with continue-on-error

mortent:
When we changed the apply algorithm to continue on errors, it also changed the way we report errors through events. Errors that aren't fatal are no longer reported with an `ErrorEvent`, but rather attached to a `Apply|Prune|DeleteEvent`. 

This PR updates the printer logic to look for errors attached to the specific events and also makes sure we exit with a non-zero exit code if there are any errors. 

This fixes the issue report in https://github.com/GoogleContainerTools/kpt/issues/1470. The output from the example will now be:
```
$ kapply apply . --server-side              
configmap/ConfigMap failed: ConfigMap "ConfigMap" is invalid: metadata.name: Invalid value: "ConfigMap": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
1 resource(s) applied. 0 created, 0 unchanged, 0 configured, 1 failed
0 resource(s) pruned, 0 skipped, 0 failed
error: 1 resources failed

$ echo $?
1
```

### DIFF
--- a/cmd/printers/events/formatter.go
+++ b/cmd/printers/events/formatter.go
@@ -46,8 +46,13 @@ func (ef *formatter) FormatApplyEvent(ae event.ApplyEvent, as *list.ApplyStats, 
 	case event.ApplyEventResourceUpdate:
 		gk := ae.Identifier.GroupKind
 		name := ae.Identifier.Name
-		ef.print("%s %s", resourceIDToString(gk, name),
-			strings.ToLower(ae.Operation.String()))
+		if ae.Error != nil {
+			ef.print("%s failed: %s", resourceIDToString(gk, name),
+				ae.Error.Error())
+		} else {
+			ef.print("%s %s", resourceIDToString(gk, name),
+				strings.ToLower(ae.Operation.String()))
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


pkg/apply/event/event.go
```diff
@@ -81,6 +81,7 @@ type ApplyEventType int
 const (
 	ApplyEventResourceUpdate ApplyEventType = iota
 	ApplyEventCompleted
+	ApplyEventFailed
```
Liujingfang1, line 84: So `ApplyEventFailed` is for a single resource. For any single resource, we expect to generate either a `ApplyEventResourceUpdate` event or a `ApplyEventFailed` event?

mortent: I realized the event hierarchy probably require some more thought, but we should try to address the asymmetry between the Prune|Delete events and the Apply event. 
I removed this from this PR, so this just addresses the printing problem. We have an existing issue to address the event hierarchy: https://github.com/GoogleContainerTools/kpt/issues/1470

Liujingfang1: Yes, there is some inconsistency for Apply|Prune|Delete events. Let's address that in a different PR.

PR #336, opened 2021-03-15
capture apply/prune duration for single object in events

Liujingfang1:
This change adds a new field `Duration` to ApplyEvent, PruneEvent and DeleteEvent.

The `Duration` is used to capture the time taken by applying/pruning/deleting a single object. 

@seans3  @mortent

### DIFF
--- a/pkg/apply/duration/duration.go
+++ b/pkg/apply/duration/duration.go
@@ -1,0 +1,20 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package duration
+
+import "time"
+
+var startTime time.Time
+
+// SetStartTime sets the start time to now.
+func SetStartTime(t time.Time) {
+	startTime = t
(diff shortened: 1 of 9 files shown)


pkg/apply/duration/duration.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package duration
+
+import "time"
+
+var startTime time.Time
```
ash2k, line 8: This would make it impossible to use this package (and all the packages that depend on it) concurrently. Why not just use a local variable?

ash2k (changes requested):

karlkfi:
> The Duration is used to capture the time taken by applying/pruning/deleting a single object.

Is this supposed to be the time spent applying or the time spent reconciling?

Was this to be exposed by Config Sync?

Is it still needed?

PR #338, opened 2021-03-22
Don't emit Prune event for NoMatchError

Liujingfang1:
The existing prune has a bug when handling the following case:
1. The inventory list contains a CR
2. The CRD is deleted from the cluster. Consequently the CR is also deleted from the cluster.
3. Then remove the configuration for the CR from the local directory
4. Apply the directory, since the CR is in the previously applied set and not in the current apply set. The CR should be pruned.
5. When getting the CR from the cluster, an error is get since no type is not matched since the CRD is deleted.
6. It emits a PruneEvent with this error.

Since the CR doesn't exist in step 5, we actually don't need to trigger the deletion for it. Emitting the PruneEvent with an error doesn't make sense either. It only needs to remove the GKNN of the CR from the inventory list, which is already done by the current code.

This PR checks if the error from step 5 is a NoMatchError. It it is, continue to the next object.

### DIFF
--- a/pkg/apply/prune/prune.go
+++ b/pkg/apply/prune/prune.go
@@ -130,7 +130,7 @@ func (po *PruneOptions) Prune(localInv inventory.InventoryInfo,
 		obj, err := po.getObject(pruneObj)
 		if err != nil {
 			// Object not found in cluster, so no need to delete it; skip to next object.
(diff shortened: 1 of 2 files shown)


mortent (approved):
/lgtm

PR #339, opened 2021-03-27
Improve the event hierarchy

mortent:
This PR is a proposal for how we can change the event hierarchy to make it more consistent, and to provide better support for more advanced use-cases where we might have multiple groups of resources that must be applied and pruned separately (typically because there are dependencies between them so one must exist and be running before the next is applied, similar for prune).

The main changes are:
 * Instead of having variants of each event type specify when a task has completed, it introduces an `ActionGroupEvent` which corresponds to the task queue. These will be sent before a new task is started and when a task is completed. It includes information about the set of resources handled by the task and allows us to name each task. The list of ActionGroups/tasks are provided to clients at the beginning with the `InitEvent`.
 * `Apply`, `Prune`, and `Delete` events no longer has information pertaining to the full set of resources in a task. These events always pertain to just a single resource. I think we have concluded that information about whether a set of resources were all successfully applied/pruned isn't very useful. It is also pretty trivial for a client to keep track of that if needed.
 * `Status` events also always pertain to a single resource. 
 * `Wait` tasks currently doesn't have a set of events pertaining to the task, although `Status` events are somewhat tied to the wait tasks. But unlike the `Apply`, `Prune` and `Delete` events which will only be emitted when the corresponding task is running, `Status` events can come at any point.

The updated types for Events are available at https://github.com/mortent/cli-utils/blob/155c4e6911bff5bbabe950fb4902c236baf58d48/pkg/apply/event/event.go

Some known issues that I haven't fully figured out yet.
 * This design doesn't provide a way for a client to know whether a `Wait` task completed successfully or not. While clients can probably figure this out, it is a bit awkward. Creating a specific `Wait` events for this would be strange, since it would be kinda like the `ActionGroup` events. An alternative might be to include an additional property on the `ActionGroup` events that would signal success/failure.


pkg/apply/event/event.go
```diff
@@ -64,9 +67,12 @@ type ResourceAction int
 const (
 	ApplyAction ResourceAction = iota
 	PruneAction
+	DeleteAction
+	WaitAction
 )
 
-type ResourceGroup struct {
+type ActionGroup struct {
+	Name        string
```
Liujingfang1, line 75: What is the intent of adding the `Name` field. Based on the code from other file, it is assigned to be the TaskName. This is kind of duplicate with the values for `Action`.

pkg/apply/event/event.go
```diff
@@ -75,92 +81,73 @@ type ErrorEvent struct {
...
+	Action    ResourceAction
+	Type      ActionGroupEventType
+}
+
 //go:generate stringer -type=ApplyEventOperation
 type ApplyEventOperation int
 
 const (
-	ServersideApplied ApplyEventOperation = iota
+	NoneApply ApplyEventOperation = iota
```
Liujingfang1, line 102: The `Failed` operation is removed. How does the event consumer know the apply is failed? Is it by checking the `Error` field in the ApplyEvent?

pkg/apply/event/event.go
```diff
@@ -75,92 +81,73 @@ type ErrorEvent struct {
...
 
-//go:generate stringer -type=DeleteEventType
-type DeleteEventType int
-
-const (
-	DeleteEventResourceUpdate DeleteEventType = iota
-	DeleteEventCompleted
-	DeleteEventFailed
-)
-
```
Liujingfang1, line 151: Just to make sure my understanding is correct. The `DeleteEventType` is removed. Similar for ApplyEventType and PruneEventType. That means for consumers, they can't rely on the DeleteEventFailed to check if failure happens. They need to check the Error field inside each event.

pkg/apply/event/event.go
mortent: So the idea is that we want some way to distinguish different tasks of the same type. The most relevant example for the current functionality is when we create a separate Apply task for CRDs, but eventually we want to support additional ordering features which could lead to multiple tasks of the same type.
In this PR, the tasks just get named with the name of the type and then a number. So if we have two apply tasks, they will be named `apply-1` and `apply-2`. I have been looking at whether this is a better way to name them that reflects their purpose, but I haven't found a good solution yet.

pkg/apply/event/event.go
mortent: Yeah, the idea is to avoid having two places where users can check for errors/failures. It is not clear what it means if an ApplyEvent has operation `Failed` but no error, or the other way around. Do we have examples where having a separate `operation` would be useful?

pkg/apply/event/event.go
mortent: Yeah, the idea here is that we no longer have any events that pertains to all resources in the task. So every event of ApplyEventType/PruneEventType/DeleteEventType will belong to a single resource, i.e. the `Identifier` field should always have a value. This also means that there is only one way to check for failures, and that is to check the `Error` field.

pkg/apply/event/event.go
Liujingfang1: I'm not aware of such examples.

cmd/printers/events/formatter.go
```diff
@@ -29,78 +27,57 @@ type formatter struct {
...
-				ae.Error.Error())
-		} else {
-			ef.print("%s %s", resourceIDToString(gk, name),
-				strings.ToLower(ae.Operation.String()))
-		}
+func (ef *formatter) FormatApplyEvent(ae event.ApplyEvent) error {
+	gk := ae.Identifier.GroupKind
+	name := ae.Identifier.Name
+	if ae.Error != nil {
+		ef.print("%s failed: %s", resourceIDToString(gk, name),
```
Liujingfang1, line 34: Maybe `"%s apply failed: %s"`?

pkg/apply/event/event.go
```diff
@@ -37,6 +38,8 @@ type Event struct {
 	// ErrorEvent contains information about any errors encountered.
 	ErrorEvent ErrorEvent
 
+	ActionGroupEvent ActionGroupEvent
```
Liujingfang1, line 41: Could you add some comments for `ActionGroupEvent`?

pkg/apply/event/event.go
Liujingfang1, line 102: Maybe change `NonApply` to `ApplyUnspecified`. `NonApply` seems to have the similar meaning as `Unchanged`.

pkg/apply/event/event.go
```diff
@@ -75,92 +81,73 @@ type ErrorEvent struct {
...
-	PruneEventCompleted
-	PruneEventFailed
-)
-
 //go:generate stringer -type=PruneEventOperation
 type PruneEventOperation int
 
 const (
-	Pruned PruneEventOperation = iota
+	NonePrune PruneEventOperation = iota
```
Liujingfang1, line 127: similar to `NonApply`, how about changing `NonePrune` to `PruneUnspecified` or `Unspecified`.

pkg/apply/event/event.go
```diff
@@ -75,92 +81,73 @@ type ErrorEvent struct {
...
-	DeleteEventCompleted
-	DeleteEventFailed
-)
-
 //go:generate stringer -type=DeleteEventOperation
 type DeleteEventOperation int
 
 const (
-	Deleted DeleteEventOperation = iota
+	NoneDelete DeleteEventOperation = iota
```
Liujingfang1, line 143: Ditto

Liujingfang1 (commented):
Overall the change looks good to me. Can you add some documentation that explains how the resolver create multiple action groups; how each action group is named and how the wait tasks work. It's better to explain that with some examples. With that, it is easier for the library users and first-time developers. The documentation can be added in a separate PR.

pkg/apply/event/event.go
mortent: Unfortunately I can't use the same name (like `Unspecified`) for the the different types. So they need to have a name that includes Apply/Prune/Destroy. I have renamed them to `(Apply|Prune|Destroy)Unspecified`

mortent:
> Overall the change looks good to me. Can you add some documentation that explains how the resolver create multiple action groups; how each action group is named and how the wait tasks work. It's better to explain that with some examples. With that, it is easier for the library users and first-time developers. The documentation can be added in a separate PR.

Yeah, this whole effort desperately needs better documentation..

seans3 (approved):
/lgtm

PR #346, opened 2021-05-09
[WIP] Add depends-on annotation and apply sort order

seans3:
* Using `depends-on` annotation, sort apply objects. Insert `WaitTask` between `ApplyTask`s to ensure a successful apply before running the next apply.
* Without the `depends-on`annotation, the functionality does **not** change.
* Adds an object graph and unit tests, where `ObjMetadata` are the vertices, and directed edges show an object depends on another object.
* Adds a topological `Sort()` to the object graph to generate an ordering of object apply sets.
* Adds implicit object ordering for CRD's and namespaces.


seans3:
Completed this functionality in another PR: https://github.com/kubernetes-sigs/cli-utils/pull/379

PR #348, opened 2021-05-12
Use rand from apimachinery to avoid seeding rng every time

ash2k:
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/rand/rand.go

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -5,8 +5,8 @@ package common
 
 import (
 	"fmt"
-	"math/rand"
 
+	"k8s.io/apimachinery/pkg/util/rand"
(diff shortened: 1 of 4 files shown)


ash2k:
I'm now part of this org too, let's see if this works

/ok-to-test

ash2k:
Friendly ping @mortent

ash2k:
/assign seans3

@seans3 Could you have a look? I'd like to make more (and more interesting) improvements to the library as I'd like to start using it.

mortent (approved):
Looks good to me. Thanks for contributing.
/lgtm
/approve

PR #349, opened 2021-05-14
Fix flaky applier test

mortent:
The test `apply resource with existing object belonging to different inventory` were expecting the `StatusEventResourceUpdate` events to come during the wait task. But since applying the resource failed, we actually don't wait for those events. Thus, we can't predict where they will be in the final event slice.


seans3 (approved):
/lgtm

PR #351, opened 2021-05-19
Don't cache nil mapper

ash2k:
If there was an error getting the mapper, it will not be retried and a nil, nil will be returned, which is incorrect.


mortent (approved):
/lgtm
/approve

PR #352, opened 2021-05-19
More reusable BaseListPrinter

ash2k:
Make `BaseListPrinter` more reusable by hiding IO streams from it. Formatter owns the output, so it can fully encapsulate it. That way a formatter can print to e.g. a logger, not to a raw stream and that would not require ignoring the passed IO streams.

I'd like to make a formatter, that prints to a logger. But I'd still like to use `BaseListPrinter` to call it. By making this change my custom formatter factory would not need to ignore IO streams.


mortent (approved):
/lgtm
/approve

mortent (approved):
/lgtm

PR #353, opened 2021-05-19
Provide the DryRunStrategy through the destroyer options

mortent:
Providing the DryRunStrategy by setting it as a property on the Destroyer is awkward and makes it hard to use an interface to represent the Deployer. This makes the DryRunStrategy part of the DestroyOptions using the same pattern as the Applier.

### DIFF
--- a/pkg/apply/destroyer.go
+++ b/pkg/apply/destroyer.go
@@ -33,14 +33,18 @@ func NewDestroyer(provider provider.Provider) *Destroyer {
 // Destroyer performs the step of grabbing all the previous inventory objects and
 // prune them. This also deletes all the previous inventory objects
 type Destroyer struct {
-	provider       provider.Provider
-	PruneOptions   *prune.PruneOptions
-	invClient      inventory.InventoryClient
-	DryRunStrategy common.DryRunStrategy
+	provider     provider.Provider
+	PruneOptions *prune.PruneOptions
+	invClient    inventory.InventoryClient
 }
 
 type DestroyerOption struct {
+	// InventoryPolicy defines the inventory policy of apply.
 	InventoryPolicy inventory.InventoryPolicy
+
+	// DryRunStrategy defines whether changes should actually be performed,
+	// or if it is just talk and no action.
+	DryRunStrategy common.DryRunStrategy
 }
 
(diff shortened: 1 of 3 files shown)


pkg/apply/destroyer.go
```diff
@@ -68,7 +72,9 @@ func (d *Destroyer) Run(inv inventory.InventoryInfo, option *DestroyerOption) <-
 
 	go func() {
 		defer close(ch)
-		d.invClient.SetDryRunStrategy(d.DryRunStrategy)
+		// TODO: We should see if we can avoid having the dry-run strategy
+		// as a field on the inventory client.
```
ash2k, line 76: I was looking at that just yesterday! It's very unfortunate that the client is mutable and hence it's not safe for concurrent use. And because of that the `Applier` is also not thread safe. I wanted to ask - would it make sense to refactor the constructor+`Initialize()` into a more traditional pattern of factory+immutable object i.e. remove `Initialize()` and remove mutability. Then the "dry run" strategy can be passed to the client on construction and the client is immutable. Client construction is done in Applier as needed, without any mutation. That is what I came up with yesterday, maybe there is a better way. WDYT?

I'm happy to make the refactor and to update `kpt` too to acomodate the breaking changes.

ash2k: Found it https://github.com/GoogleContainerTools/kpt/issues/1462

ash2k: Yeah, it'd be better as a parameter to client's methods than a constructor parameter.

mortent: Yeah, we want the Applier and the Destroyer to be stateless and immutable so users can create it once and then use it multiple times. This also includes the InventoryClient. I'm also hoping we eventually can simplify the interface of the InventoryClient as it currently has 7-8 functions that seems somewhat redundant.
The `Initialize` functions in both the Applier and the Destroyer are remains from when we used the kubectl libraries in a different way, and I agree we want to avoid those if possible. 

If you want to refactor this, that would be great! The kpt team will take care of any updates needed in kpt when we pull in the changes, so no need to worry about that unless you want to contribute.

seans3 (approved):
/lgtm

PR #354, opened 2021-05-20
Reduce allocations in set operations

ash2k:
/kind cleanup


pkg/object/objmetadata.go
```diff
@@ -253,17 +253,17 @@ func calcHash(objs []string) (uint32, error) {
 // do not exist in "b" (A - B).
 func SetDiff(setA []ObjMetadata, setB []ObjMetadata) []ObjMetadata {
 	// Create a map of the elements of A
-	m := map[string]ObjMetadata{}
+	m := make(map[ObjMetadata]struct{}, len(setA))
 	for _, a := range setA {
-		m[a.String()] = a
```
ash2k, line 258: No need to turn structs into strings as long as all members as hashable.

mortent (approved):
/lgtm
/approve

PR #355, opened 2021-05-21
Remove Applier.Initialize()

ash2k:
WDYT? If this approach is fine, I'll refactor other objects in a similar way (in separate PRs).

/assign @mortent 
/kind cleanup

### DIFF
--- a/cmd/apply/cmdapply.go
+++ b/cmd/apply/cmdapply.go
@@ -20,11 +20,11 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/inventory"
 	"sigs.k8s.io/cli-utils/pkg/manifestreader"
 	"sigs.k8s.io/cli-utils/pkg/provider"
+	"sigs.k8s.io/cli-utils/pkg/util/factory"
 )
 
 func GetApplyRunner(provider provider.Provider, loader manifestreader.ManifestLoader, ioStreams genericclioptions.IOStreams) *ApplyRunner {
 	r := &ApplyRunner{
-		Applier:   apply.NewApplier(provider),
 		ioStreams: ioStreams,
 		provider:  provider,
 		loader:    loader,
@@ -73,7 +73,6 @@ type ApplyRunner struct {
 	Command    *cobra.Command
 	PreProcess func(info inventory.InventoryInfo, strategy common.DryRunStrategy) (inventory.InventoryPolicy, error)
 	ioStreams  genericclioptions.IOStreams
-	Applier    *apply.Applier
 	provider   provider.Provider
 	loader     manifestreader.ManifestLoader
 
@@ -132,13 +131,19 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
 			return err
 		}
 	}
+	f := r.provider.Factory()
+	statusPoller, err := factory.NewStatusPoller(f)
+	if err != nil {
+		return err
+	}
 
 	// Run the applier. It will return a channel where we can receive updates
 	// to keep track of progress and any issues.
-	if err := r.Applier.Initialize(); err != nil {
+	a, err := apply.NewApplier(r.provider, statusPoller)
+	if err != nil {
 		return err
 	}
-	ch := r.Applier.Run(context.Background(), inv, objs, apply.Options{
+	ch := a.Run(context.Background(), inv, objs, apply.Options{
 		ServerSideOptions: r.serverSideOptions,
 		PollInterval:      r.period,
 		ReconcileTimeout:  r.reconcileTimeout,
--- a/cmd/preview/cmdpreview.go
+++ b/cmd/preview/cmdpreview.go
@@ -12,7 +12,6 @@ import (
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/util/i18n"
-
 	"sigs.k8s.io/cli-utils/cmd/flagutils"
 	"sigs.k8s.io/cli-utils/cmd/printers"
 	"sigs.k8s.io/cli-utils/pkg/apply"
@@ -21,6 +20,7 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/inventory"
 	"sigs.k8s.io/cli-utils/pkg/manifestreader"
 	"sigs.k8s.io/cli-utils/pkg/provider"
+	"sigs.k8s.io/cli-utils/pkg/util/factory"
 )
 
 var (
@@ -31,7 +31,6 @@ var (
 // GetPreviewRunner creates and returns the PreviewRunner which stores the cobra command.
 func GetPreviewRunner(provider provider.Provider, loader manifestreader.ManifestLoader, ioStreams genericclioptions.IOStreams) *PreviewRunner {
 	r := &PreviewRunner{
-		Applier:   apply.NewApplier(provider),
 		Destroyer: apply.NewDestroyer(provider),
 		ioStreams: ioStreams,
 		provider:  provider,
@@ -75,7 +74,6 @@ type PreviewRunner struct {
 	Command    *cobra.Command
 	PreProcess func(info inventory.InventoryInfo, strategy common.DryRunStrategy) (inventory.InventoryPolicy, error)
 	ioStreams  genericclioptions.IOStreams
-	Applier    *apply.Applier
 	Destroyer  *apply.Destroyer
 	provider   provider.Provider
 	loader     manifestreader.ManifestLoader
@@ -123,7 +121,15 @@ func (r *PreviewRunner) RunE(cmd *cobra.Command, args []string) error {
 	// if destroy flag is set in preview, transmit it to destroyer DryRunStrategy flag
 	// and pivot execution to destroy with dry-run
 	if !previewDestroy {
-		err = r.Applier.Initialize()
+		_, err = common.DemandOneDirectory(args)
+		if err != nil {
+			return err
+		}
+		statusPoller, err := factory.NewStatusPoller(r.provider.Factory())
+		if err != nil {
+			return err
+		}
+		a, err := apply.NewApplier(r.provider, statusPoller)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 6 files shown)


cmd/apply/cmdapply.go
```diff
@@ -133,13 +134,25 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
 
 	// Run the applier. It will return a channel where we can receive updates
 	// to keep track of progress and any issues.
-	if err := r.Applier.Initialize(); err != nil {
-		return err
-	}
-	ch := r.Applier.Run(context.Background(), inv, objs, apply.Options{
+	a := apply.NewApplier(f, statusPoller, invClient, info.NewInfoHelper(f), pruneOpts)
```
mortent, line 154: My concern about this approach is that it just pulls the functionality that was previously in `Initialize` and puts it in the consumer. This means the api for the `NewApplier` function requires several objects that seems unnecessary. Fixing some of this (like the `PruneOptions`) requires a larger change, so I think we should initially focus on making the API for the Applier as clean as possible so we can later refactor the internals without breaking users. 
How about we let the `NewApplier` function return an error as the second return value and we can just do all the setup inside the function?

ash2k: Thanks for the review!

> This means the api for the NewApplier function requires several objects that seems unnecessary.

But they are necessary - it needs them? I think it's good to accept direct dependencies and leave out the construction code:

- As you can see, tests become cleaner (no/fewer private field reassignment)
- It becomes possible to inject custom implementations (which may be crucial for downstream consumers if not now, but in the future).
- That also makes it possible to share instances of those dependencies.

FWIW I, as a consumer of this library, would really prefer to have the simplest possible constructor that takes all the direct dependencies and I don't mind having to create the dependencies myself. It's clear how to do it!

However, I see your point. I thought about it too while working on this.

- One other option is to have a helper construction function in the `cmd` package. Leaving it out from the `applier` package keeps the list of imports shorter - not depending on the implementation packages is a good thing.
- Yet another option is to use the builder pattern to construct the Applier. Builder would use default implementations for omitted dependencies. Ideally the builder should live in a separate package to keep the dependency graph clean, but maybe that's not worth it.

> I think we should initially focus on making the API for the Applier as clean as possible so we can later refactor the internals without breaking users.

I agree, we just need to agree on what "clean" means =) Happy to iterate on the PR.

ash2k: @mortent friendly ping :)

mortent: So I see your point. I think the challenge here is that we currently don't have the right abstractions, so the dependencies here are a bit random. In particular the InfoHelper and the PruneOptions are awkward. 

The Provider interface was meant to provide the tools the Applier needs to talk to the cluster, in practice this means everything related to talking the cluster (which is the `util.Factory` interface) and how to interact with the inventory object(s) in the cluster (through the `InventoryClient`). I can see that allowing users to provide a separate implementation of the StatusPoller could be useful, but I don't think that even uses an interface right now.
So would improve things we had an interface that accepted the Provider interface and the StatusPoller (by creating an interface for this)?

ash2k: > In particular the InfoHelper and the PruneOptions are awkward.

I fully agree.

> I can see that allowing users to provide a separate implementation of the StatusPoller could be useful, but I don't think that even uses an interface right now.

[It does](https://github.com/kubernetes-sigs/cli-utils/blob/1c8e8871efb7082a83d88250157373d4a29effd2/pkg/apply/poller/poller.go#L23-L25). I'm particularly interested in either using a custom implementation or ideally somehow sharing the existing caching implementation across multiple `Applier` instances. In my program I have 0-N (depends on user-supplied config) `Applier`s and I definitely don't want more than one informer for the same GVK. In addition to that, when an applier is no longer needed I'd like to shut down caching informer(s) that only it used and that are not needed by other appliers (some sort of "reference counting" approach would work here). If I don't do this, my long running (unlike e.g. `kpt`) program may eventually run out of memory (or at least would be wasting memory, network, and CPU resources). Currently there is no way for me to do this but I'll keep sending PRs to get there, if that's ok.

> The Provider interface was meant to provide the tools the Applier needs to talk to the cluster

Honestly I don't see much value in it vs just providing direct dependencies. [Currently](https://github.com/kubernetes-sigs/cli-utils/blob/1c8e8871efb7082a83d88250157373d4a29effd2/pkg/provider/provider.go#L26-L45) it creates a client on each invocation. That is unnecessary and can be replaced with `return p.client`. Then the `Provider` is just a holder of pre-created "tools". But then it doesn't need to be an interface because it does not encapsulate any behavior, it just contains things for grouping purposes. It's not wrong or bad, just saying it's not necessary to have an interface. In my personal opinion it's just easier to provide all dependencies directly to each consumer. It's a bit more verbose but there is no "wrapping" into another holder object and it's clearer what needs what.

> So would improve things we had an interface that accepted the Provider interface and the StatusPoller (by creating an interface for this)?

Are you suggesting something like this:

```go
func NewApplier(provider provider.Provider, statusPoller poller.Poller) *Applier 
```

I would add `PruneOptions` there but make an interface for it. The reasoning is that it encapsulates some behavior, it's not just a bunch of configuration options.

I'm suggesting two options:

```go
type provider.Provider interface {
	Factory() util.Factory
	InventoryClient() inventory.InventoryClient     // removed "error" because client is pre-constructed
}

func NewApplier(provider provider.Provider, statusPoller poller.Poller, pruner prune.Pruner) *Applier 
```

OR

```go
func NewApplier(factory util.Factory, invClient inventory.InventoryClient, statusPoller poller.Poller, pruner prune.Pruner) *Applier 
```

Note that both options above don't return `error` from `NewApplier()` because it's just a matter of assigning fields. In the second option we can probably eliminate the concept of `Provider` completely - as I said above, it does not encapsulate any behavior if `InventoryClient()` method is refactored to just return a field, which makes sense as there is no need to construct the client each time.

Both options use:

```go
type prune.Pruner interface {
    Prune(localInv inventory.InventoryInfo,
	localObjs []*unstructured.Unstructured,
	currentUIDs sets.String,
	taskContext *taskrunner.TaskContext,
	o prune.Options) error 
}
```

Later, something can be done to `info.InfoHelper`, I'm currently not sure what.

WDYT?

ash2k: @mortent I've reworked this PR to use `func NewApplier(provider provider.Provider, statusPoller poller.Poller) (*Applier, error)` signature. I'm interested to hear your thoughts on what I wrote above. If this PR looks good, please feel free to merge it as it would be cleaner to make next refactorings (whatever they are) in separate PRs.

mortent: > [It does](https://github.com/kubernetes-sigs/cli-utils/blob/1c8e8871efb7082a83d88250157373d4a29effd2/pkg/apply/poller/poller.go#L23-L25). I'm particularly interested in either using a custom implementation or ideally somehow sharing the existing caching implementation across multiple `Applier` instances. In my program I have 0-N (depends on user-supplied config) `Applier`s and I definitely don't want more than one informer for the same GVK. In addition to that, when an applier is no longer needed I'd like to shut down caching informer(s) that only it used and that are not needed by other appliers (some sort of "reference counting" approach would work here). If I don't do this, my long running (unlike e.g. `kpt`) program may eventually run out of memory (or at least would be wasting memory, network, and CPU resources). Currently there is no way for me to do this but I'll keep sending PRs to get there, if that's ok.

Yeah, please do. The plan was always to make the status poller more flexible, both in terms of letting users extend the resource types for which the library knows how to compute status and implementing different solutions for fetching status from the cluster. We have discussed using informers or watches instead of polling, but we haven't had time to do it yet.

> Honestly I don't see much value in it vs just providing direct dependencies. [Currently](https://github.com/kubernetes-sigs/cli-utils/blob/1c8e8871efb7082a83d88250157373d4a29effd2/pkg/provider/provider.go#L26-L45) it creates a client on each invocation. That is unnecessary and can be replaced with `return p.client`. Then the `Provider` is just a holder of pre-created "tools". But then it doesn't need to be an interface because it does not encapsulate any behavior, it just contains things for grouping purposes. It's not wrong or bad, just saying it's not necessary to have an interface. In my personal opinion it's just easier to provide all dependencies directly to each consumer. It's a bit more verbose but there is no "wrapping" into another holder object and it's clearer what needs what.

Yeah, I'm convinced. It doesn't really do much right now and the fact that it wraps the `util.Factory` is a bit confusing.

> 
> > So would improve things we had an interface that accepted the Provider interface and the StatusPoller (by creating an interface for this)?
> 
> Are you suggesting something like this:
> 
> ```go
> func NewApplier(provider provider.Provider, statusPoller poller.Poller) *Applier 
> ```

Yeah, something like that was what I had in mind, but I'm probably ok with getting rid of the Provider interface. I just want @seans3 to provide input on this too.

> 
> I would add `PruneOptions` there but make an interface for it. The reasoning is that it encapsulates some behavior, it's not just a bunch of configuration options.

Is there a particular aspect of the pruner you want to customize, or is it more to keep the option open? I agree with the direction, just curious about the use-case.

> 
> I'm suggesting two options:
> 
> ```go
> type provider.Provider interface {
> 	Factory() util.Factory
> 	InventoryClient() inventory.InventoryClient     // removed "error" because client is pre-constructed
> }
> 
> func NewApplier(provider provider.Provider, statusPoller poller.Poller, pruner prune.Pruner) *Applier 
> ```
> 
> OR
> 
> ```go
> func NewApplier(factory util.Factory, invClient inventory.InventoryClient, statusPoller poller.Poller, pruner prune.Pruner) *Applier 
> ```
> 
> Note that both options above don't return `error` from `NewApplier()` because it's just a matter of assigning fields. In the second option we can probably eliminate the concept of `Provider` completely - as I said above, it does not encapsulate any behavior if `InventoryClient()` method is refactored to just return a field, which makes sense as there is no need to construct the client each time.
> 
> Both options use:
> 
> ```go
> type prune.Pruner interface {
>     Prune(localInv inventory.InventoryInfo,
> 	localObjs []*unstructured.Unstructured,
> 	currentUIDs sets.String,
> 	taskContext *taskrunner.TaskContext,
> 	o prune.Options) error 
> }
> ```

Yeah, I think the second of the two options (without the Provider interface) looks good.

> 
> Later, something can be done to `info.InfoHelper`, I'm currently not sure what.

The Infohelper is mostly just a result of having to turn `*unstructured.Unstructured` into `*resource.Info` since we are using kubectl libraries to actually do the apply. I hope we can get rid of it completely.

ash2k: @mortent Thanks for the comment. I actually wanted to clarify what I wrote above, because I changed my opinion a bit while I was reworking this PR to the current state, but forgot to leave a comment. Most of what I said above is still what I think, but:

- `info.InfoHelper` should probably be moved into an `internal` package to make it clear it's not for public consumption and is a re-usable piece of implementation. No need to accept it as a constructor parameter.
- `prune.PruneOptions` - same as above. It's probably best to keep it in `internal` until there is a valid use case for customization. That would allow for more flexibility to refactor it without worrying of breaking consumers. No need to accept it as a constructor parameter.

I'm happy to refactor things further, including the removal of the `Provider` interface.

seans3 (approved):
/lgtm

PR #356, opened 2021-05-21
Use a specific error type when encountering unknown types

mortent:
The error message to users when we encounter a resource type that can't be found in the cluster or among the applied CRDs is not very helpful. This introduces a specific type for this error and a more helpful error message.


seans3 (approved):
/lgtm

PR #358, opened 2021-05-24
Move inventory operations into tasks

seans3:
* Moves inventory operations into their own tasks.
* Moves initial inventory `Merge` from `applier` into `InvAddTask`, which is the first task in the queue.
* Moves final inventory `Replace` from `prune` into `InvSetTask`, which is the last task in the queue.
* Adds unit tests for these two new tasks.
* Stores prune failures in task context for final inventory calculation.


mortent:
So it seems like the Inventory tasks must always be the first and last tasks in the task queue. With this being the case, I'm not convinced that it wouldn't be easier to keep it as part of the framework instead.
I couldn't see from this PR how we would handle situations where we might not want to run all tasks in the queue, an example being that we want to bail out if a wait task times out. Is the thinking that we would then jump to the end of the task queue to make sure we always update the inventory?

seans3:
> So it seems like the Inventory tasks must always be the first and last tasks in the task queue. With this being the case, I'm not convinced that it wouldn't be easier to keep it as part of the framework instead.

I think the advantage of using tasks instead of making it part of the framework is that it's more flexible, since we might end up having multiple `InvAddTask`s or multiple `InvSetTask`s.

> I couldn't see from this PR how we would handle situations where we might not want to run all tasks in the queue, an example being that we want to bail out if a wait task times out. Is the thinking that we would then jump to the end of the task queue to make sure we always update the inventory?

Yes--the thinking is to always run the `InvSetTask`. But I wanted to keep the functionality the same, so I didn't implement that yet. Maybe there will be parameters for "continue-on-error" within the task to tell the framework how to run.

pkg/apply/task/inv_add_task.go
```diff
@@ -0,0 +1,81 @@
...
+// into the cluster. The InvAddTask should add/merge inventory references
+// before the actual object is applied.
+type InvAddTask struct {
+	InvClient inventory.InventoryClient
+	InvInfo   inventory.InventoryInfo
+	Objects   []*unstructured.Unstructured
+}
+
+func (i *InvAddTask) Name() string {
+	return "inventory-add"
```
mortent, line 26: We might want to have unique names here in cases where we have more than one instance in the task queue. The idea was that clients can use the name to identify tasks/groups for the `ActionGroup` events.

pkg/apply/event/event.go
```diff
@@ -72,6 +72,7 @@ const (
 	PruneAction
 	DeleteAction
 	WaitAction
+	InventoryAction
```
mortent, line 75: Does any of the cli code (in particular the printing code) make assumptions about which type of events they should expect? We probably should avoid them failing on unknown event types, but I'm not sure how it is done today.

pkg/apply/applier_test.go
```diff
@@ -689,17 +689,6 @@ var obj2 = &unstructured.Unstructured{
 	},
 }
```
mortent, line 691: I would expect the events coming from the applier to change. Should we make sure this is covered in the applier tests?

pkg/apply/solver/solver_test.go
```diff
@@ -38,10 +38,12 @@ func TestTaskQueueSolver_BuildTaskQueue(t *testing.T) {
 			objs:    []*unstructured.Unstructured{},
 			options: Options{},
 			expectedTasks: []taskrunner.Task{
+				&task.InvAddTask{},
```
mortent, line 41: Should we consider these as any other task, so they will show up in the event stream? Or should we somehow consider these an implementation detail and hide them?

pkg/apply/event/event.go
seans3: Currently, it just does nothing because the switch statement for inventory event types do not have any code:

https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/print/list/base.go#L129

We can add extra verbosity inventory printing, but I don't think we need them to start.

pkg/apply/solver/solver_test.go
seans3: Added inventory events.

pkg/apply/event/event.go
```diff
@@ -154,3 +165,14 @@ type DeleteEvent struct {
 	Object     *unstructured.Unstructured
 	Error      error
 }
+
+type InventoryAddEvent struct {
```
mortent, line 169: So what happens if the `InventoryAddEvent` fails? It seems like we should just bail out since it isn't safe to continue. So I'm uncertain what the use is for the `Error` property here. In general for the events, if there is a fatal error, we return an event of type `ErrorEvent`.

I'm thinking we should not add these events unless we have a use-case for it. We can always add it later if it needed.

pkg/apply/task/inv_add_task.go
```diff
@@ -0,0 +1,105 @@
...
+			taskContext.EventChannel() <- createInvAddFailedEvent(err)
+			taskContext.TaskChannel() <- taskrunner.TaskResult{Err: err}
+			return
+		}
+		// Ensures the namespace exists before applying the inventory object into it.
+		if invNamespace := inventoryNamespaceInSet(i.InvInfo, i.Objects); invNamespace != nil {
+			klog.V(4).Infof("applying inventory namespace %s", invNamespace.GetName())
+			if err := i.InvClient.ApplyInventoryNamespace(invNamespace); err != nil {
+				taskContext.EventChannel() <- createInvAddFailedEvent(err)
+				taskContext.TaskChannel() <- taskrunner.TaskResult{Err: err}
```
mortent, line 52: This seems redundant. We should only return a single event for the error. In this case, the last line should be sufficient.

pkg/apply/task/inv_add_task.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (i *InvAddTask) Identifiers() []object.ObjMetadata {
+	return object.UnstructuredsToObjMetas(i.Objects)
+}
+
+// Start updates the inventory by merging the locally applied objects
+// into the current inventory.
+func (i *InvAddTask) Start(taskContext *taskrunner.TaskContext) {
+	go func() {
+		if err := inventory.ValidateNoInventory(i.Objects); err != nil {
```
mortent, line 42: Where do we check for `dry-run` here?

seans3: dry-run is set on the `InventoryClient`: https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/apply/applier.go#L221

pkg/apply/event/event.go
seans3: Removed inventory events and tests.

pkg/apply/task/inv_add_task.go
seans3: Removed event channel redundancy.

PR #359, opened 2021-05-25
Expose ManifestReader helper functions to users

mortent:
These are useful for users who wants to write their own implementation of the ManifestReader.


seans3 (approved):
/lgtm

PR #362, opened 2021-05-28
Simplify: remove unneeded meta.Accessor() usage

ash2k:


pkg/inventory/inventory.go
```diff
@@ -73,16 +72,7 @@ func IsInventoryObject(obj *unstructured.Unstructured) bool {
 // for the passed inventory object. Returns error if the passed object is nil or
 // is not a inventory object.
 func retrieveInventoryLabel(obj *unstructured.Unstructured) (string, error) {
-	if obj == nil {
```
ash2k, line 76: Cannot be `nil` here.

pkg/inventory/inventory.go
```diff
@@ -134,16 +124,12 @@ func SplitUnstructureds(objs []*unstructured.Unstructured) (*unstructured.Unstru
 // an error if name stored in the object differs from the name in
 // the Info struct.
 func addSuffixToName(obj *unstructured.Unstructured, suffix string) error {
-	if obj == nil {
```
ash2k, line 137: Cannot be `nil` here.

pkg/inventory/inventory.go
```diff
@@ -163,15 +149,7 @@ func addSuffixToName(obj *unstructured.Unstructured, suffix string) error {
 // This fixes a problem where inventory object names collide if
 // they are created in the same namespace.
 func fixLegacyInventoryName(obj *unstructured.Unstructured) error {
-	if obj == nil {
```
ash2k, line 166: Cannot be `nil` here.

pkg/inventory/inventory.go
seans3: Up to you if you want to remove these checks. There is disagreement here, but many consider defensively validating all parameters good practice. These validation checks make the contract between the caller and callee more concrete.

seans3 (approved):
This is cleaner--thanks.

/lgtm

pkg/inventory/inventory.go
ash2k: I think in Go the default assumption is that parameters are not `nil` (for pointers). I don't think you can often find such checks in code, at least not in the Kubernetes codebase. Also, this is a private function and arguably the check is not needed at all. This check is like an assertion rather than input validation. But the code will crash immediately anyway if a `nil` is passed, so no need for an extra assertion. That was my reasoning :)

PR #366, opened 2021-06-07
Refactor destroyer to use tasks

seans3:
* Refactors `destroyer` to delete objects using tasks, which will allow us to order the deletions.
* Refactors `TaskQueueSolver` to `TaskQueueBuilder` now using the `Builder` pattern for reuse in the `destroyer`.
* Adds task to delete inventory object (and unit test).
* Adds `EventFactory` interface to abstract prune/delete event generation.
* Adds e2e mdrip tests to validate destroy behavior.

### DIFF
--- a/pkg/apply/applier.go
+++ b/pkg/apply/applier.go
@@ -190,7 +190,7 @@ func (r *ResourceObjects) AllIds() []object.ObjMetadata {
 // on progress and any errors are reported back on the event channel.
 // Cancelling the operation or setting timeout on how long to Wait
 // for it complete can be done with the passed in context.
-// Note: There sn't currently any way to interrupt the operation
+// Note: There isn't currently any way to interrupt the operation
 // before all the given resources have been applied to the cluster. Any
 // cancellation or timeout will only affect how long we Wait for the
 // resources to become current.
@@ -201,31 +201,29 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
 	a.invClient.SetDryRunStrategy(options.DryRunStrategy) // client shared with prune, so sets dry-run for prune too.
 	go func() {
 		defer close(eventChannel)
-
 		// This provides us with a slice of all the objects that will be
 		// applied to the cluster. This takes care of ordering resources
 		// and handling the inventory object.
-		resourceObjects, err := a.prepareObjects(invInfo, objects)
+		resourceObjs, err := a.prepareObjects(invInfo, objects)
 		if err != nil {
 			handleError(eventChannel, err)
 			return
 		}
-
 		mapper, err := a.provider.Factory().ToRESTMapper()
 		if err != nil {
 			handleError(eventChannel, err)
 			return
 		}
-
 		// Fetch the queue (channel) of tasks that should be executed.
 		klog.V(4).Infoln("applier building task queue...")
-		taskQueue := (&solver.TaskQueueSolver{
+		taskBuilder := &solver.TaskQueueBuilder{
 			PruneOptions: a.PruneOptions,
 			Factory:      a.provider.Factory(),
 			InfoHelper:   a.infoHelper,
 			Mapper:       mapper,
 			InvClient:    a.invClient,
-		}).BuildTaskQueue(resourceObjects, solver.Options{
+		}
+		opts := solver.Options{
 			ServerSideOptions:      options.ServerSideOptions,
 			ReconcileTimeout:       options.ReconcileTimeout,
 			Prune:                  !options.NoPrune,
@@ -233,8 +231,14 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
 			PrunePropagationPolicy: options.PrunePropagationPolicy,
 			PruneTimeout:           options.PruneTimeout,
 			InventoryPolicy:        options.InventoryPolicy,
-		})
-
+		}
+		// Build the task queue by appending tasks in the proper order.
+		taskQueue := taskBuilder.
+			AppendInvAddTask(resourceObjs).
+			AppendApplyWaitTasks(resourceObjs, opts).
+			AppendPruneWaitTasks(resourceObjs, opts).
+			AppendInvSetTask(resourceObjs).
+			Build()
 		// Send event to inform the caller about the resources that
 		// will be applied/pruned.
 		eventChannel <- event.Event{
@@ -243,10 +247,9 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
 				ActionGroups: taskQueue.ToActionGroups(),
 			},
 		}
-
 		// Create a new TaskStatusRunner to execute the taskQueue.
 		klog.V(4).Infoln("applier building TaskStatusRunner...")
-		runner := taskrunner.NewTaskStatusRunner(resourceObjects.AllIds(), a.StatusPoller)
+		runner := taskrunner.NewTaskStatusRunner(resourceObjs.AllIds(), a.StatusPoller)
 		klog.V(4).Infoln("applier running TaskStatusRunner...")
 		err = runner.Run(ctx, taskQueue.ToChannel(), eventChannel, taskrunner.Options{
 			PollInterval:     options.PollInterval,
@@ -303,7 +306,7 @@ type Options struct {
 // have not been provided.
 func setDefaults(o *Options) {
 	if o.PollInterval == time.Duration(0) {
-		o.PollInterval = 2 * time.Second
+		o.PollInterval = poller.DefaultPollInterval
 	}
 	if o.PrunePropagationPolicy == metav1.DeletionPropagation("") {
 		o.PrunePropagationPolicy = metav1.DeletePropagationBackground
--- a/pkg/apply/destroyer.go
+++ b/pkg/apply/destroyer.go
@@ -4,17 +4,21 @@
 package apply
 
 import (
-	"fmt"
+	"context"
+	"time"
 
 	"github.com/go-errors/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/cli-utils/pkg/apply/event"
+	"sigs.k8s.io/cli-utils/pkg/apply/poller"
 	"sigs.k8s.io/cli-utils/pkg/apply/prune"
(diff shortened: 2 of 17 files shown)


pkg/apply/applier.go
```diff
@@ -201,40 +201,44 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
...
-		})
-
+		}
+		// Build the task queue by appending tasks in the proper order.
+		taskQueue := taskBuilder.
+			AppendInvAddTask(resourceObjs).
+			BuildApplyTaskQueue(resourceObjs, opts).
+			BuildPruneTaskQueue(resourceObjs, opts).
+			AppendInvSetTask(resourceObjs).
+			Build()
```
mortent, line 241: I find the pattern here with several `Build*` functions a bit confusing.

pkg/apply/destroyer.go
```diff
@@ -4,19 +4,25 @@
...
 	"sigs.k8s.io/cli-utils/pkg/apply/prune"
+	"sigs.k8s.io/cli-utils/pkg/apply/solver"
 	"sigs.k8s.io/cli-utils/pkg/apply/taskrunner"
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/cli-utils/pkg/inventory"
 	"sigs.k8s.io/cli-utils/pkg/provider"
+	"sigs.k8s.io/cli-utils/pkg/util/factory"
 )
 
+const defaultPollInterval = 2 * time.Second
```
mortent, line 24: I'm pretty sure there is already a default value for this defined somewhere. The value should probably be the same for destroy/apply, so we should try to do this in one place.

pkg/apply/destroyer.go
```diff
@@ -68,107 +80,62 @@ func (d *Destroyer) Initialize() error {
...
-		ch <- event.Event{
-			Type: event.ActionGroupType,
-			ActionGroupEvent: event.ActionGroupEvent{
-				GroupName: "destroyer",
-				Type:      event.Finished,
-				Action:    event.DeleteAction,
+		resourceObjs := &ResourceObjects{LocalInv: inv, PruneIds: pruneIds}
+		klog.V(4).Infoln("destroyer building task queue...")
+		taskBuilder := &solver.TaskQueueBuilder{
+			PruneOptions: d.PruneOptions,
```
mortent, line 101: It seems strange here that `PruneOptions` are provided to the `TaskBuilder`. Why aren't this just part of the `solver.Options`?

pkg/apply/destroyer.go
```diff
@@ -68,107 +80,62 @@ func (d *Destroyer) Initialize() error {
...
+			Mapper:       mapper,
+			InvClient:    d.invClient,
+		}
+		opts := solver.Options{
+			Prune:                  true,
+			PruneTimeout:           time.Duration(0),
+			DryRunStrategy:         option.DryRunStrategy,
+			PrunePropagationPolicy: metav1.DeletePropagationBackground,
+			InventoryPolicy:        option.InventoryPolicy,
+			ReconcileTimeout:       time.Duration(0),
```
mortent, line 112: The `ReconcileTimeout` shouldn't be used here I would assume, so I think it would be better if we just didn't set it at all.

pkg/apply/destroyer.go
```diff
@@ -68,107 +80,62 @@ func (d *Destroyer) Initialize() error {
...
+			PruneOptions: d.PruneOptions,
+			Factory:      d.provider.Factory(),
+			Mapper:       mapper,
+			InvClient:    d.invClient,
+		}
+		opts := solver.Options{
+			Prune:                  true,
+			PruneTimeout:           time.Duration(0),
+			DryRunStrategy:         option.DryRunStrategy,
+			PrunePropagationPolicy: metav1.DeletePropagationBackground,
```
mortent, line 110: Both the `PruneTimeout` and the `PrunePropagationPolicy` should be configurable. We should at least add a TODO here.

pkg/apply/destroyer.go
```diff
@@ -68,107 +80,62 @@ func (d *Destroyer) Initialize() error {
...
-				}
-			}
+		// Create a new TaskStatusRunner to execute the taskQueue.
+		klog.V(4).Infoln("destroyer building TaskStatusRunner...")
+		runner := taskrunner.NewTaskStatusRunner(resourceObjs.AllIds(), d.StatusPoller)
+		klog.V(4).Infoln("destroyer running TaskStatusRunner...")
+		err = runner.Run(context.Background(), taskQueue.ToChannel(), eventChannel, taskrunner.Options{
+			UseCache:         true,
+			PollInterval:     defaultPollInterval,
+			EmitStatusEvents: false,
```
mortent, line 134: We should just set this to `true`. Longer-term we should remove this property completely and leave it up to the caller to filter out status events if they are not needed.

pkg/apply/destroyer.go
```diff
@@ -68,107 +80,62 @@ func (d *Destroyer) Initialize() error {
...
-					},
-				}
-			}
+		// Create a new TaskStatusRunner to execute the taskQueue.
+		klog.V(4).Infoln("destroyer building TaskStatusRunner...")
+		runner := taskrunner.NewTaskStatusRunner(resourceObjs.AllIds(), d.StatusPoller)
+		klog.V(4).Infoln("destroyer running TaskStatusRunner...")
+		err = runner.Run(context.Background(), taskQueue.ToChannel(), eventChannel, taskrunner.Options{
+			UseCache:         true,
+			PollInterval:     defaultPollInterval,
```
mortent, line 133: We should make this configurable just like with the `apply` command.

pkg/apply/prune/prune.go
```diff
@@ -150,7 +150,11 @@ func (po *PruneOptions) Prune(localInv inventory.InventoryInfo,
 		// Handle lifecycle directive preventing deletion.
 		if !canPrune(localInv, obj, o.InventoryPolicy, uid) {
 			klog.V(4).Infof("skip prune for lifecycle directive %s/%s", pruneObj.Namespace, pruneObj.Name)
-			taskContext.EventChannel() <- createPruneEvent(pruneObj, obj, event.PruneSkipped)
+			e := createPruneEvent(pruneObj, obj, event.PruneSkipped)
+			if po.Destroy {
+				e = createDeleteEvent(pruneObj, obj, event.DeleteSkipped)
```
mortent, line 155: This feels a little bit awkward, but we can see if there are ways to improve this later.

pkg/apply/prune/prune.go
```diff
@@ -248,6 +260,18 @@ func createPruneEvent(id object.ObjMetadata, obj *unstructured.Unstructured, op
 	}
 }
 
+// createDeleteEvent is a helper function to package a delete event.
+func createDeleteEvent(id object.ObjMetadata, obj *unstructured.Unstructured, op event.DeleteEventOperation) event.Event {
```
mortent, line 264: How we do we create Delete events with errors?

pkg/apply/solver/solver.go
```diff
@@ -86,127 +97,148 @@ type resourceObjects interface {
...
+// Build returns the queue of tasks that have been created.
+func (t *TaskQueueBuilder) Build() *TaskQueue {
+	return &TaskQueue{
+		tasks: t.tasks,
+	}
+}
+
+// AppendInvAddTask appends an inventory add task to the task queue.
+// Returns a pointer to the Builder to chain function calls.
+func (t *TaskQueueBuilder) AppendInvAddTask(ro resourceObjects) *TaskQueueBuilder {
```
mortent, line 109: Do we need to expose this to callers? It seemed like this is something that can be handled inside the `Solver/TaskQueueBuilder`.

pkg/apply/destroyer.go
seans3: `TaskQueueSolver` was renamed to `TaskQueueBuilder`, but I did not change this. `PruneOptions` was a parameter to `TaskQueueSolver`. If we want to change this, let's do it in another smaller PR.

pkg/apply/destroyer.go
seans3: Added a TODO for this.

pkg/apply/prune/prune.go
seans3: It is a bit awkward, but it allows us to get rid of even more awkwardness with the channel transformers. Adding a TODO to improve this in the future.

pkg/apply/solver/solver.go
seans3: We can add a layer on top of the builder, but I think this is a reasonable step to build the task queue in a more dynamic way. I think we can firm up the API when we get to "depends-on".

pkg/apply/prune/prune.go
seans3: In `destroy`, we did not have specific delete event errors before; only generic error events. And the event channel prune/destroy transformer did not handle error events. So this functionality is the same as before.

pkg/apply/destroyer.go
mortent: We might have to fix this sooner rather than later. I don't think depends-on for prune/delete will work unless the timeout and propagation policy is set correctly.

pkg/apply/prune/prune.go
mortent: But I think we should fix this now. As far as I can tell, the code on lines [184](https://github.com/kubernetes-sigs/cli-utils/pull/366/files#diff-9d41d28d25555dfdd294c704d1577e1d7ecac3fd06fe94fb9e4df5a1c1409e2fR184) and [193](https://github.com/kubernetes-sigs/cli-utils/pull/366/files#diff-9d41d28d25555dfdd294c704d1577e1d7ecac3fd06fe94fb9e4df5a1c1409e2fR193) will lead to Prune events being created during destroy.

pkg/apply/solver/solver.go
mortent: Agree

pkg/apply/prune/prune.go
seans3: I created an `EventFactory` interface to abstract the prune/delete event generation.

pkg/apply/prune/prune.go
seans3: Fixed. Also created `EventFactory` interface to clean up event generation.

pkg/apply/destroyer.go
seans3: Fixed. Any cli which imports this library will have to surface these two parameters.

mortent (approved):
/lgtm

PR #368, opened 2021-06-10
Prune refactor to take objects to delete

seans3:
* Updates `prune` function to take the objects to delete instead of calculating the prune set. This will allow us to order the objects to delete.
* `Applier.prepareObjects` now returns the objects to apply and prune. The previous return encapsulation `ResourceObjects` has been removed in favor of these returned apply objects and prune objects.
* Prune unit tests have been re-written.

### DIFF
--- a/pkg/apply/applier.go
+++ b/pkg/apply/applier.go
@@ -7,11 +7,13 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/go-errors/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/cli-utils/pkg/apply/event"
 	"sigs.k8s.io/cli-utils/pkg/apply/info"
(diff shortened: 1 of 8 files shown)


pkg/apply/prune/prune_test.go
```diff
@@ -183,155 +153,487 @@ var preventDelete = &unstructured.Unstructured{
...
+			}()
+			if err != nil {
+				t.Fatalf("Unexpected error during Prune(): %#v", err)
+			}
+			var actualEvents []event.Event
+			for e := range eventChannel {
+				actualEvents = append(actualEvents, e)
+			}
+			// Validate the expected/actual error events
+			if len(tc.expectedEvents) != len(actualEvents) {
```
mortent, line 499: Can we reuse the validation logic for events?
https://github.com/kubernetes-sigs/cli-utils/blob/1c8e8871efb7082a83d88250157373d4a29effd2/pkg/testutil/events.go#L53

We use this code in the e2e tests and the applier tests.

pkg/apply/applier.go
```diff
@@ -83,107 +85,42 @@ func (a *Applier) Initialize() error {
...
 	}
 	if err := inventory.ValidateNoInventory(localObjs); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
-
 	// If the inventory uses the Name strategy and an inventory ID is provided,
 	// verify that the existing inventory object (if there is one) has an ID
 	// label that matches.
+	// TODO(seans): This inventory id validation should happen in destroy and status.
```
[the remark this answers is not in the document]
seans3: Acknowledged.

pkg/apply/prune/prune_test.go
seans3: Updated prune unit tests to use the `testutil.VerifyEvents` function

mortent (approved):
/lgtm

PR #369, opened 2021-06-15
Refactor prune/delete using ValidationFilter interface

seans3:
* Creates `ValidationFilter` interface to abstract operations for prune (deletion) permission. This interface decouples prune/delete dependencies including on `inventory` and `InventoryPolicy`.
* Creates several filters which implement this `ValidationFilter` interface, including `PreventDeleteFilter`, `InventoryPolicyFilter`, `CurrentUIDFilter`, and `LocalNamespacesFilter`.
* Integrates these new filters into the prune and delete functionality.
* Adds unit tests for all the new filters.
* Changes `TaskContext.AllResourceUIDs()` function name to `TaskContext.AppliedResourceUIDs()` because it is more accurate.

### DIFF
--- a/pkg/apply/prune/prune.go
+++ b/pkg/apply/prune/prune.go
@@ -13,16 +13,15 @@ package prune
 
 import (
 	"context"
-	"fmt"
 	"sort"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/cmd/util"
+	"sigs.k8s.io/cli-utils/pkg/apply/filter"
 	"sigs.k8s.io/cli-utils/pkg/apply/taskrunner"
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/cli-utils/pkg/inventory"
@@ -39,8 +38,6 @@ type PruneOptions struct {
 	// True if we are destroying, which deletes the inventory object
 	// as well (possibly) the inventory namespace.
 	Destroy bool
-	// TODO(seans): Replace this with Filter interface to generalize prune skipping.
-	LocalNamespaces sets.String
 }
 
 // NewPruneOptions returns a struct (PruneOptions) encapsulating the necessary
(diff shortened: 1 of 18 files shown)


pkg/apply/task/prune_task.go
```diff
@@ -50,12 +49,16 @@ func (p *PruneTask) Identifiers() []object.ObjMetadata {
 // to signal to the taskrunner that the task has completed (or failed).
 func (p *PruneTask) Start(taskContext *taskrunner.TaskContext) {
 	go func() {
-		currentUIDs := taskContext.AllResourceUIDs()
-		err := p.PruneOptions.Prune(p.InventoryObject, p.Objects,
-			currentUIDs, taskContext, prune.Options{
+		// Create filter to prevent deletion of currently applied
+		// objects. Must be done here to wait for applied UIDs.
+		uidFilter := filter.CurrentUIDFilter{
+			CurrentUIDs: taskContext.AllResourceUIDs(),
```
mortent, line 55: Nit: We should consider renaming `AllResourceUIDs`. As far as I can tell, it actually only returns the UIDs of the resources that have been successfully applied.

pkg/apply/prune/prune.go
```diff
@@ -77,55 +73,57 @@ type Options struct {
...
+				if klog.V(5).Enabled() {
+					klog.Errorf("error during %s, (%s): %s", filter.Name(), pruneID, err)
+				}
+				taskContext.EventChannel() <- eventFactory.CreateFailedEvent(pruneID, err)
+				taskContext.CapturePruneFailure(pruneID)
+				break
+			}
+			if filtered {
+				klog.V(4).Infof("prune filtered by %s: %s", filter.Name(), pruneID)
+				taskContext.EventChannel() <- eventFactory.CreateSkippedEvent(pruneObj)
```
mortent, line 118: It seems like we need some way of including the reason why a resource was skipped here. It would be pretty useful to know if it is due to a lifecycle directive (and in that case, which one), because it is a namespace in use by one or more applied objects, or something different.

mortent (commented):
Just one comment. I like the direction.

pkg/apply/task/prune_task.go
seans3: Changed `AllResourceUIDs` to `AppliedResourceUIDs`.

pkg/apply/prune/prune.go
seans3: We will add a `Reason` function in the `ValidationFilter` interface to replace the `Name` function. Also, we will update the `CreateSkippedEvent` function signature to take the reason as a string. We should probably also update the `PruneEvent` struct to add a `Reason` field, but we should probably do that in another PR. Please let me know if this direction is what you had in mind.

seans3: Since this can be a large change, and since this PR is already large, we should probably put this refactor work off to another PR.

mortent (approved):
/lgtm

PR #371, opened 2021-06-23
Re-organizes pkg/object, adds annotation helpers

seans3:
* Re-organizes `pkg/object` package by moving functions out of `objmetadata.go` and into `infos.go` and new `unstructured.go`.
* Adds helper functions for annotations in `annotations.go`, including upcoming `depends-on` annotation.
* Moves `IsCRD` helper function from `solver.go` to `object` package since it is re-used elsewhere.
* Adds unit tests. Test coverage is now up to 74.7%.

### DIFF
--- a/pkg/object/annotations_test.go
+++ b/pkg/object/annotations_test.go
@@ -1,0 +1,199 @@
+// Copyright 2021 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package object
+
+import (
+	"testing"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var (
+	clusterScopedObj = ObjMetadata{Name: "cluster-obj",
+		GroupKind: schema.GroupKind{Group: "test-group", Kind: "test-kind"}}
+	namespacedObj = ObjMetadata{Namespace: "test-namespace", Name: "namespaced-obj",
+		GroupKind: schema.GroupKind{Group: "test-group", Kind: "test-kind"}}
+)
+
+var u1 = &unstructured.Unstructured{
+	Object: map[string]interface{}{
+		"apiVersion": "v1",
+		"kind":       "ConfigMap",
+		"metadata": map[string]interface{}{
+			"name":      "unused",
+			"namespace": "unused",
+			"annotations": map[string]interface{}{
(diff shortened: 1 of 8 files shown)


pkg/object/annotations_test.go
```diff
@@ -0,0 +1,199 @@
...
+)
+
+var (
+	clusterScopedObj = ObjMetadata{Name: "cluster-obj",
+		GroupKind: schema.GroupKind{Group: "test-group", Kind: "test-kind"}}
+	namespacedObj = ObjMetadata{Namespace: "test-namespace", Name: "namespaced-obj",
+		GroupKind: schema.GroupKind{Group: "test-group", Kind: "test-kind"}}
+)
+
+var u1 = &unstructured.Unstructured{
```
mortent, line 21: Nit: I think this might be more readable if we just put the KRM manifests as strings and then parse it into Unstructured. We do this in the applier tests.

pkg/object/infos.go
```diff
@@ -4,15 +4,45 @@
...
+	"fmt"
+
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/cli-runtime/pkg/resource"
 	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 )
 
+// InfosToObjMetas returns object metadata (ObjMetadata) for the
+// passed objects (infos); returns an error if one occurs.
+func InfosToObjMetas(infos []*resource.Info) ([]ObjMetadata, error) {
```
mortent, line 16: Probably something for another PR, but we should see if we can reduce the number of times we end up converting between infos, unstructured, and ObjMetadata...

seans3: We should probably keep objects in `Unstructured` format for most of the time, since `Info` and `ObjMetadata` can easily be created from this foundational format.

seans3: ...but you're right. That's probably best in another PR.

mortent: Yeah, what gets a bit difficult is that converting from Unstructured to ObjMetadata without checks (like scope to know whether cluster-scoped or namespace-scoped) isn't necessarily safe. I'm working on a fix that would make this check immediately in the Applier.

pkg/object/annotations_test.go
seans3: Sounds good. I think this is better handled by creating a testutil package with a bunch of standard KRM resources for all tests. That should be another PR.

PR #372, opened 2021-06-24
Remove Provider interface

ash2k:
Follow up for conversation here https://github.com/kubernetes-sigs/cli-utils/pull/355#discussion_r638164688.

I couldn't just remove the `Provider` interface without introducing an indirection to allow injecting a custom inventory client and at the same time deferring its construction (because top level functions don't return an error). So I introduced the `InventoryClientFactory`. WDYT?


pkg/inventory/inventory-client-factory.go
```diff
@@ -0,0 +1,21 @@
...
+
+package inventory
+
+import cmdutil "k8s.io/kubectl/pkg/cmd/util"
+
+var (
+	_ InventoryClientFactory = ClusterInventoryClientFactory{}
+)
+
+type InventoryClientFactory interface {
```
seans3, line 12: Let's add comments to describe this factory interface.

pkg/inventory/inventory-client-factory.go
```diff
@@ -0,0 +1,21 @@
...
+	NewInventoryClient(factory cmdutil.Factory) (InventoryClient, error)
+}
+
+type ClusterInventoryClientFactory struct {
```
seans3, line 16: Let's add some comments to describe this concrete factory. I think we should change the name from `Cluster` to `ConfigMap`, since the default inventory object is a `ConfigMap`. But we can debate that for later.

seans3 (changes requested):
Looks good. Adding the factory seems fine to me. Some minor nits about adding comments.

seans3 (approved):
/lgtm

PR #373, opened 2021-06-24
Adds Reason to ValidationFilter and skipped events

seans3:
* Adds reason string in return of `Filter` interface function to provide an explanation for prune/delete filtering.
* Implements reason return string for each `ValidationFilter` and unit tests.
* Adds `Reason` field to `PruneEvent` and `DeleteEvent` to store prune/delete skipping explanation.
* Exports `InventoryIDMatch()` function for better prune/delete reason string in the `InventoryPolicyFilter`.
* Does **NOT** surface the `Reason` in printing yet.

### DIFF
--- a/pkg/apply/filter/filter.go
+++ b/pkg/apply/filter/filter.go
@@ -14,6 +14,8 @@ import (
 type ValidationFilter interface {
 	// Name returns a filter name (usually for logging).
 	Name() string
-	// Filter returns true if validation fails or an error.
-	Filter(obj *unstructured.Unstructured) (bool, error)
+	// Filter returns true if validation fails. If true a
+	// reason string is included in the return. If an error happens
+	// during filtering it is returned.
+	Filter(obj *unstructured.Unstructured) (bool, string, error)
 }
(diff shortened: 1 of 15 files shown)


pkg/apply/filter/filter.go
```diff
@@ -16,4 +16,6 @@ type ValidationFilter interface {
 	Name() string
 	// Filter returns true if validation fails or an error.
 	Filter(obj *unstructured.Unstructured) (bool, error)
+	// Reason returns the string explanation if filtered.
+	Reason() string
```
mortent, line 20: So I find this a bit strange. We are using a single instance of a filter to handle all resources, but this adds state specific to a single resource. This seems like a solution that can easily lead to bugs.

seans3: You're right. It is not re-entrant. I will change the signature of the `Filter()` function to additionally return the reason string.

pkg/apply/prune/prune.go
```diff
@@ -97,10 +97,11 @@ func (po *PruneOptions) Prune(pruneObjs []*unstructured.Unstructured,
 		klog.V(5).Infof("attempting prune: %s", pruneID)
 		// Check filters to see if we're prevented from pruning/deleting object.
 		var filtered bool
+		var reason string
```
mortent, line 100: Nit: Why is `reason` defined outside of the loop? It seems like it is only used inside.

mortent (approved):
/lgtm

PR #374, opened 2021-06-25
Log statement at start of tasks

seans3:
* Simple log statement at the beginning of inventory tasks and wait task.


mortent (approved):
/lgtm

PR #375, opened 2021-06-26
Adds object dependency sorting and tests

seans3:
* Adds `SortObjs` and `ReverseSortObjs` functions to sort object dependencies into ordered sets of objects. Unit test coverage is 97.2% for `pkg/object/graph`.
* Object dependencies include:
  1) Explicit `depends-on` annotation
  2) Implicit `Namespace` and objects within the namespace
  3) Implicit `CRD` and custom resources defined by the CRD
* Refactors testutils to include some common object-related functions for testing objects.

### DIFF
--- a/pkg/apply/applier_test.go
+++ b/pkg/apply/applier_test.go
@@ -69,49 +69,6 @@ spec:
 	}
 )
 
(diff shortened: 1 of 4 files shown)


mortent (approved):
/lgtm

PR #376, opened 2021-06-28
Destroyer calculates emit status events

seans3:
* `Destroyer` now dynamically chooses whether to emit status events on the event channel (previously hard-coded). The first pass of this choice is to only emit status events when the `--delete-timeout` is set.
* Adds single log statement to `applier.go`


mortent (approved):
/lgtm

PR #377, opened 2021-06-29
Inventory not found is not error during delete

seans3:
* The delete inventor task now skips the `Not Found` error. If the inventory object is not found, then it has already been deleted.
* Adds unit tests.

### DIFF
--- a/pkg/apply/task/delete_inv_task_test.go
+++ b/pkg/apply/task/delete_inv_task_test.go
@@ -6,6 +6,8 @@ package task
 import (
 	"testing"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/apply/task/delete_inv_task_test.go
```diff
@@ -29,4 +31,21 @@ func TestDeleteInvTask(t *testing.T) {
 	if result.Err != nil {
 		t.Errorf("unexpected error running DeleteInvTask: %s", result.Err)
 	}
+	// Hard-code error from fake inventory client, and make sure
```
mortent, line 34: Can we set this up as table-driven tests?

seans3: Done

mortent (approved):
/lgtm

PR #380, opened 2021-07-01
Drop github.com/go-errors/errors and github.com/pkg/errors dependencies

ash2k:
Kubernetes is getting rid of the second one too https://github.com/kubernetes/kubernetes/issues/103043. The first one is still a transitive dep, only coming from Kustomize. Can probably be removed there too.

Should probably drop `gotest.tools` too and leave only `github.com/stretchr/testify` for assertions.

### DIFF
--- a/cmd/status/cmdstatus.go
+++ b/cmd/status/cmdstatus.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/go-errors/errors"
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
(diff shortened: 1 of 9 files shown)


cmd/status/cmdstatus.go
```diff
@@ -123,7 +122,7 @@ func (r *StatusRunner) runE(cmd *cobra.Command, args []string) error {
 		ErrOut: cmd.ErrOrStderr(),
 	})
 	if err != nil {
-		return errors.WrapPrefix(err, "error creating printer", 1)
+		return fmt.Errorf("error creating printer: %v", err)
```
mortent, line 125: We should probably use `%w` instead of `%v` here so we have the option of unwrapping the error. Same goes for the other changes.

ash2k: Then I suppose it needs to be done consistently across the whole codebase? There are places where errors are not wrapped at all and places that use `%v`. Are these particular spots I'm touching in this PR special in some way that unwrapping must be possible? Just trying to understand.

mortent: I think we should just use `%w` to wrap the error whenever we use this pattern. I'm sure it is not used consistently across the codebase, but it seems like a good practice.

ash2k: @mortent I agree. Done!

mortent (approved):
/lgtm
/approve

PR #381, opened 2021-07-02
Avoid referencing cmd line flags in error strings

mortent:
Create a custom error type rather than using an error string that references command line flags. Any users of the library can chose to handle the error in any way they see fit.
Ref: https://github.com/kubernetes-sigs/cli-utils/pull/378#discussion_r661142159


seans3 (approved):
/lgtm

PR #383, opened 2021-07-05
Add option for custom status readers

### LINKED ISSUE #382: Allow to inject custom StatusReaders into a StatusPoller
Currently, StatusPoller does not have a way to inject custom StatusReaders when running `Poll`. This is problematic for downstream projects making use of CRDs and having the desire to supply custom readers to kstatus. One of the issues this causes is described in https://github.com/fluxcd/kustomize-controller/issues/387.

makkes:
This commit adds the `CustomStatusReadersFactoryFunc` option to the
StatusPoller's `Poll` method options so that consumers can provide
custom status readers for certain GroupKinds (esp. CRDs).

It also changes the `NewGenericStatusReader` func so that it accepts a
statusFunc parameter so that consumers only have to implement that
single func and use the genericStatusReader for all the boilerplate
stuff (like fetching an object from the API).

refs #382

### DIFF
--- a/pkg/kstatus/polling/polling.go
+++ b/pkg/kstatus/polling/polling.go
@@ -68,6 +68,10 @@ type Options struct {
 	// then each resource will be fetched when needed with GET calls.
 	UseCache bool
 
+	// CustomStatusReadersFactoryFunc, when called, provides the StatusPoller with a map of custom
+	// StatusReaders for the given GroupKinds. These will be used along with the StatusReaders shipped with this
+	// library. However, it can also be used to override these with custom StatusReaders if the returned map
+	// has a key for the given GroupKinds, e.g. apps/deployments.
 	CustomStatusReadersFactoryFunc func(engine.ClusterReader, meta.RESTMapper) map[schema.GroupKind]engine.StatusReader
 }
 
--- a/pkg/kstatus/polling/polling.go
+++ b/pkg/kstatus/polling/polling.go
@@ -14,6 +14,7 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/engine"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/statusreaders"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/cli-utils/pkg/object"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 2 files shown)


pkg/kstatus/polling/polling.go
```diff
@@ -56,6 +67,8 @@ type Options struct {
 	// all needed resources before each polling cycle. If this is set to false,
 	// then each resource will be fetched when needed with GET calls.
 	UseCache bool
+
+	CustomStatusReadersFactoryFunc func(engine.ClusterReader, meta.RESTMapper) map[schema.GroupKind]engine.StatusReader
```
mortent, line 71: Can we add a comment to this field explaining what it does?

pkg/kstatus/polling/polling.go
```diff
@@ -38,10 +39,20 @@ type StatusPoller struct {
 // back on the event channel returned. The statusPollerRunner can be cancelled at any time by cancelling the
 // context passed in.
 func (s *StatusPoller) Poll(ctx context.Context, identifiers []object.ObjMetadata, options Options) <-chan event.Event {
+	statusReaderFactory := createStatusReaders
```
mortent, line 42: I wish there was a simpler way do to this, but there doesn't seem to be an obvious way do to it. It is a bit unfortunate that that the `StatusReadersFactoryFunc` is a function instead of an interface.

pkg/kstatus/polling/statusreaders/generic.go
```diff
@@ -14,14 +14,16 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/object"
 )
 
-func NewGenericStatusReader(reader engine.ClusterReader, mapper meta.RESTMapper) engine.StatusReader {
+type StatusFunc func(u *unstructured.Unstructured) (*status.Result, error)
```
mortent, line 17: Add a comment

mortent (commented):
Supporting custom status readers has been the plan since the beginning, we just haven't had time to do it. This is a good change. I just have a few small comments.

pkg/kstatus/polling/polling.go
makkes: I agree it's a little involved but I wanted to introduce this change as lightweight as possible. An interface would be great as a next step, though.

pkg/kstatus/polling/polling.go
makkes: done

pkg/kstatus/polling/statusreaders/generic.go
makkes: done

mortent (approved):
/lgtm
/approve

PR #384, opened 2021-07-06
Printers for status returns error

mortent:
This updates the `Collector` in the polling package to return any fatal errors and not only update them in the collector state. It also updates the `Printer` interface used by the `status` command to also return the error rather than delegating handling of the error completely to the printer. This allows for centralized error handling regardless of output format, for example by setting the return code based on type of error.

### DIFF
--- a/cmd/status/printers/event/event_printer.go
+++ b/cmd/status/printers/event/event_printer.go
@@ -32,7 +32,7 @@ func NewEventPrinter(ioStreams genericclioptions.IOStreams) *eventPrinter {
 // every event and is responsible for stopping the poller when appropriate.
 // This function will block.
 func (ep *eventPrinter) Print(ch <-chan pollevent.Event, identifiers []object.ObjMetadata,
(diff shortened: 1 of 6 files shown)


cmd/status/printers/event/event_printer.go
```diff
@@ -44,8 +44,12 @@ func (ep *eventPrinter) Print(ch <-chan pollevent.Event, identifiers []object.Ob
 			cancelFunc(statusCollector, e)
 		}),
 	)
-	// Block until the done channel is closed.
-	<-done
+	// Listen to the channel until it is closed.
+	var err error
+	for msg := range done {
+		err = msg.Err
+	}
```
ash2k, line 51: Is there a chance that it can return multiple items? If yes, do we want to return the first one? If no, everything looks good to me.

mortent: No, it can only return one. An error event means the status poller is closing down.

ash2k (approved):
/lgtm

PR #385, opened 2021-07-09
Order by group kind

ash2k:
Taking group into account ensures unrelated resources with the same kind are ordered after the important ones.


mortent (approved):
/lgtm
/approve

PR #387, opened 2021-07-13
Avoid dryRun as state in inventory client

mortent:
This updates the inventory client interface to accept `DryRunStrategy` as a parameter to the functions instead of relying it to be set state on the implementation. This means an additional parameter that must be passed around, but I prefer that to keeping it as state. I think we can further simplify part of the `InventoryClient` by pulling out functionality that doesn't depend on the type of the inventory object.

This also removes the `UpdateLabels` function from the interface. As far as I can tell, it is not in use.

### DIFF
--- a/cmd/status/cmdstatus.go
+++ b/cmd/status/cmdstatus.go
@@ -98,7 +98,7 @@ func (r *StatusRunner) runE(cmd *cobra.Command, args []string) error {
 
 	// Based on the inventory template manifest we look up the inventory
 	// from the live state using the inventory client.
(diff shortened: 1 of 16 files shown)


ash2k (approved):
Ideally `GetPruneObjs()` and `GetClusterObjs()` methods shouldn't need the new parameter (as they should be read only), but I understand why it's needed at the moment. This is definitely a step forward and the signatures can be cleaned up later, when inventory object merging is removed from the read methods.
/lgtm

PR #388, opened 2021-07-27
Moves apply failure inventory calculation into inventory task

seans3:
* Move calculation of final inventory into an inventory task.
* Move the storage of previous inventory from apply task to inventory task, thereby removing an inventory dependency from the apply task.


mortent (approved):
/lgtm

PR #389, opened 2021-07-29
Pass the RESTMapper explicitly into infoHelper

justinsb:
This allows for better control, for example allowing us to invalidate
the discovery cache before proceeding.


seans3 (approved):
/lgtm

PR #391, opened 2021-08-08
Remove some strings.TrimSpace() calls

ash2k:
Names and labels cannot have any whitespace. If they do, it's an error that should be reported and fixed. Was discussed here https://github.com/kubernetes-sigs/cli-utils/pull/378#discussion_r661146618.

/assign @mortent


seans3 (approved):
/lgtm

PR #392, opened 2021-08-09
No inventory object is not an error

seans3:
* Changes `inventory.SplitUnstructureds` so that a missing inventory object is not an error. This is part of the work to separate inventory from apply, since these two operations are orthogonal. It should be possible to apply including **depends-on** ordering without an inventory object. If an application importing `Applier` wants to enforce an inventory object, the code will have to live in the application (directly before `Applier.Run`).


pkg/inventory/inventory.go
```diff
@@ -96,10 +96,13 @@ func ValidateNoInventory(objs []*unstructured.Unstructured) error {
...
-// splitUnstructureds takes a slice of unstructured.Unstructured objects and
-// splits it into one slice that contains the inventory object templates and
-// another one that contains the remaining resources.
-func SplitUnstructureds(objs []*unstructured.Unstructured) (*unstructured.Unstructured, []*unstructured.Unstructured, error) {
+// SplitUnstructureds takes of slice of Unstructured objects, and returns
+// the one inventory object (if it exists), and the slice of the remaining
+// objects. Returns nil as the inventory object (no error) if there is
+// not an inventory object in the passed slice. Returns an error if there
+// are multiple inventory objects in the passed slice.
+func SplitUnstructureds(objs []*unstructured.Unstructured) (
```
ash2k, line 104: This is a breaking change for the API. I think it'd be beneficial to make it a compilation failure so that anyone who updates the dependency will have to explicitly handle the situation. Otherwise it compiles but then doesn't work the way it used to.

pkg/inventory/inventory.go
```diff
@@ -110,7 +113,7 @@ func SplitUnstructureds(objs []*unstructured.Unstructured) (*unstructured.Unstru
 		}
 	}
 	if len(invs) == 0 {
-		return nil, resources, NoInventoryObjError{}
```
ash2k, line 113: I think `NoInventoryObjError` can be deleted as it's only used in one more place (and a test), which should be cleaned up too.

ash2k (commented):
Code changes look good, but I've left some suggestions.

mortent:
Looks good to me. Agree with @ash2k that if there is a reasonable way to make this change a compile time failure, that would be useful.

pkg/inventory/inventory.go
```diff
@@ -110,13 +109,13 @@ func SplitUnstructureds(objs []*unstructured.Unstructured) (*unstructured.Unstru
...
+		return nil, resources, nil
 	} else if len(invs) > 1 {
 		return nil, resources, MultipleInventoryObjError{
 			InventoryObjectTemplates: invs,
 		}
 	}
-	return invs[0], resources, nil
+	return WrapInventoryInfoObj(invs[0]), resources, nil
```
ash2k, line 118: This change makes this function not usable for those who want to use a custom inventory object. Is this intentional?

PR #394, opened 2021-08-17
Prune/Delete should use the specified propagation policy

mortent:
PruneOptions doesn't properly propagate the provided PropagationPolicy to the namespacedClient, which means that providing a PropagationPolicy doesn't have any effect. This addresses this issue and adds a test.


seans3 (approved):
Looks good. I can't tell why e2e tests are failing.

mortent:
@seans3 The destroyer didn't set default values on the passed in options, which caused the e2e tests to fail. This is fixed now.

seans3 (approved):
/lgtm

PR #396, opened 2021-08-23
Expose cyclic dependency error

seans3:
* Expose fatal cyclic dependency error during "depends-on" calculation.

Error looks like
```
error: cyclic dependency
        {test-cyclic-error-ns pod-a Pod} -> {test-cyclic-error-ns pod-b Pod}
        {test-cyclic-error-ns pod-b Pod} -> {test-cyclic-error-ns pod-c Pod}
        {test-cyclic-error-ns pod-c Pod} -> {test-cyclic-error-ns pod-a Pod}
```

### DIFF
--- a/pkg/apply/solver/solver.go
+++ b/pkg/apply/solver/solver.go
@@ -54,6 +54,7 @@ type TaskQueueBuilder struct {
 	waitCounter      int
 	pruneCounter     int
 	tasks            []taskrunner.Task
+	err              error
 }
 
 type TaskQueue struct {
@@ -92,10 +93,15 @@ type Options struct {
 }
 
 // Build returns the queue of tasks that have been created.
-func (t *TaskQueueBuilder) Build() *TaskQueue {
+// TODO(seans): Now that we're reporting errors, we probably
+// want to move away from the Builder patter for the TaskBuilder.
+func (t *TaskQueueBuilder) Build() (*TaskQueue, error) {
+	if t.err != nil {
+		return nil, t.err
+	}
 	return &TaskQueue{
 		tasks: t.tasks,
-	}
+	}, nil
 }
 
 // AppendInvAddTask appends an inventory add task to the task queue.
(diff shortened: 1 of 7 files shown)


pkg/apply/solver/solver.go
```diff
@@ -232,7 +239,10 @@ func (t *TaskQueueBuilder) AppendPruneWaitTasks(pruneObjs []*unstructured.Unstru
 	if o.Prune {
 		// Use the "depends-on" annotation to create a graph, ands sort the
 		// objects to prune into sets using a (reverse) topological sort.
-		pruneSets := graph.ReverseSortObjs(pruneObjs)
+		pruneSets, err := graph.ReverseSortObjs(pruneObjs)
+		if err != nil {
+			t.err = err
```
mortent, line 244: This feels a bit awkward and probably suggests that we are using the builder pattern in an unusual way here. Normally you would expect that the set/configure functions doesn't do any work, but just configures the builder and then all the actual work will be done in the `Build` function. 
With the current approach, it seems like we might want to check the `err` property on the builder on every call, so we can just exit immediately.

pkg/object/graph/graph.go
```diff
@@ -141,3 +145,18 @@ func (g *Graph) Sort() ([][]object.ObjMetadata, error) {
...
+// The cycle makes it impossible to topological sort.
+type CyclicDependencyError struct {
+	Edges []Edge
+}
+
+func (cde CyclicDependencyError) Error() string {
+	var errorBuf bytes.Buffer
+	errorBuf.WriteString("cyclic dependency")
+	for _, edge := range cde.Edges {
+		errorBuf.WriteString(fmt.Sprintf("\n\t%s -> %s", edge.From, edge.To))
```
mortent, line 159: Do we have any other examples where we print the ObjMetadata for resources? The example in the PR description seems to suggest a format that I don't think we have used elsewhere.

pkg/apply/solver/solver.go
seans3: I agree it is a bit awkward. But I'd prefer to update the `Builder` pattern in another PR if that's our path. Otherwise, this PR will be a lot larger. I'll add a TODO.

pkg/object/graph/graph.go
seans3: I have updated the printing to the more standard: <NAMESPACE>/<NAME>. It now looks like
```
error: cyclic dependency
	test-cyclic-error-ns/pod-b -> test-cyclic-error-ns/pod-c
	test-cyclic-error-ns/pod-c -> test-cyclic-error-ns/pod-a
	test-cyclic-error-ns/pod-a -> test-cyclic-error-ns/pod-b
```

mortent (approved):
/lgtm

PR #399, opened 2021-08-25
Moves inventory policy check for apply into filter

seans3:
* Moves inventory policy check during apply task into a `ValidationFilter`.
* Adds unit tests to `InventoryPolicyApplyFilter`.
* Removes the `DynamicClient` from the `applyOptionsFactoryFunc`.
* Continues effort to separate inventory functionality from apply, since we want to be able to apply without inventory.

### DIFF
--- a/pkg/apply/filter/inventory-policy-apply-filter.go
+++ b/pkg/apply/filter/inventory-policy-apply-filter.go
@@ -1,0 +1,73 @@
+// Copyright 2021 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package filter
+
+import (
+	"context"
+	"fmt"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/dynamic"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
+	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+// InventoryPolicyApplyFilter implements ValidationFilter interface to determine
+// if an object should be applied based on the cluster object's inventory id,
+// the id for the inventory object, and the inventory policy.
+type InventoryPolicyApplyFilter struct {
+	Client    dynamic.Interface
+	Mapper    meta.RESTMapper
+	Inv       inventory.InventoryInfo
+	InvPolicy inventory.InventoryPolicy
+}
+
+// Name returns a filter identifier for logging.
+func (ipaf InventoryPolicyApplyFilter) Name() string {
+	return "InventoryPolicyApplyFilter"
+}
+
+// Filter returns true if the passed object should be filtered (NOT applied) and
+// a filter reason string; false otherwise. Returns an error if one occurred
+// during the filter calculation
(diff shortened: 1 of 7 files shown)


pkg/apply/filter/inventory-policy-apply-filter.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Filter returns true if the passed object should be filtered (NOT applied) and
+// a filter reason string; false otherwise. Returns an error if one occurred
+// during the filter calculation
+func (ipaf InventoryPolicyApplyFilter) Filter(obj *unstructured.Unstructured) (bool, string, error) {
+	if obj == nil {
+		return true, "missing object", nil
+	}
+	// Object must be retrieved from the cluster to get the inventory id.
+	clusterObj, err := ipaf.getObject(object.UnstructuredToObjMetaOrDie(obj))
```
mortent, line 42: Does this add an additional `Get` for every resource, or does it just move the existing `Get` from inside the apply task to a filter? 
Regardless, I'm wondering if we should look at ways where the object from the cluster are provided to the filter, instead of it being fetched by the filter itself.  This approach doesn't seem to scale well with additional filters.

pkg/apply/task/apply_task.go
```diff
@@ -90,7 +83,7 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
 		klog.V(2).Infof("apply task starting (%d objects)", len(objects))
 		// Create a new instance of the applyOptions interface and use it
 		// to apply the objects.
-		ao, dynamic, err := applyOptionsFactoryFunc(taskContext.EventChannel(),
+		ao, _, err := applyOptionsFactoryFunc(taskContext.EventChannel(),
```
mortent, line 86: Do we still need to return a dynamic client from `applyOptionsFactoryFunc`?

pkg/apply/task/apply_task.go
```diff
@@ -115,29 +108,29 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
...
-						event.Unchanged, clusterObj)
+				if filtered {
+					klog.V(4).Infof("apply filtered by %s because (%s): %s", filter.Name(), reason, id)
+					taskContext.EventChannel() <- createApplyEvent(id, event.Unchanged, obj)
+					taskContext.CaptureResourceFailure(id)
+					break
 				}
-				taskContext.CaptureResourceFailure(id)
+			}
+			if filtered || filterErr != nil {
```
mortent, line 133: I think it might be clearer to just do continue when these situations happen instead of using break and then this check. It confused me a bit at first.

pkg/apply/filter/inventory-policy-apply-filter.go
seans3: It moves the existing `Get` to the filter. This pattern is necessary since we want to pass in the local file to filters. For example, if we implement the `local-config` filtering in a `ValidationFilter` we have to pass the local file--not the file fetched from the cluster. It is not even guaranteed that there is a cluster object (it won't exist for `local-config` filtered files). The only way this works is using local files.

pkg/apply/task/apply_task.go
seans3: No--removed.

pkg/apply/task/apply_task.go
seans3: Since there are two `for` loops, a single `continue` does not work. The `break` is to stop the filter loop. The `continue` is to go to the next object.

mortent (approved):
/lgtm

PR #400, opened 2021-08-28
Add apply-time-mutation feature

karlkfi:
New feature: apply-time-mutation

Mutation algorithm:
1. Detect `config.kubernetes.io/apply-time-mutation` annotation
2. Add dependency wait edges to wait until referenced source resource is reconciled (using existing depends-on order & wait)
3. Before applying the annotated target resource, loop through and apply substitutions specified in the annotation. Substitute the token value in the target field of the target resource with the latest value from the source field of the source resource.

Mutation details:
- Mutator interface mirrors Validator interface for injection, granular testing, and extensibility.
- ApplyTimeMutator implements Mutator interface
- ApplyTimeMutator uses JSONPath expression language for both getting and setting values.

Expression Language:
- JSONPath chosen, primarily for the element of least surprise.
  Kubernetes is using JSONPath elsewhere, but the existing implementations do not support mutation, just retrieval.
- Add `sigs.k8s.io/cli-utils/pkg/jsonpath` library with Get/Set on `map[string]interface{}`.
- Used for specifying target and source fields.
- Wraps http://github.com/spyzhov/ajson
- [Alternatives Considered](https://docs.google.com/document/d/1SChsQ9-l13Czpq4YGq7HA2JorZjL5HQ-27CJzwX77yE/edit)

Resource Caching:
- Add ResourceCache to avoid extra GETs
- Cache is single-threaded and only used by the mutator.

Other included changes:
- Move some testutil files to fix dependency cycles when using in new packages
- Add testutil.NestedField that supports array indexing to help test ApplyTimeMutator
- Added a few MainTest files to allow logging with klog while testing.


pkg/apply/mutator/apply-time-mutator.go
```diff
@@ -0,0 +1,216 @@
...
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/cli-utils/pkg/apply/cache"
+	"sigs.k8s.io/cli-utils/pkg/kyq"
+	"sigs.k8s.io/cli-utils/pkg/object"
+	"sigs.k8s.io/cli-utils/pkg/object/mutation"
+)
+
+// ApplyTimeMutator mutates a resource by injecting values specified by the
+// apply-time-mutation annotation.
+// The optional ResourceCache will be used to speed up source reosurce lookups,
```
karlkfi, line 26: reosurce typo

pkg/apply/cache/resource_caching_poller_test.go
```diff
@@ -0,0 +1,226 @@
...
+				}
+				if !cmp.Equal(receivedObj, expectedObj) {
+					t.Errorf("ResourceCachingPoller.Poll unexpected resource body:\nexpected: %#v\nreceived: %#v", expectedObj, receivedObj)
+				}
+			}
+		})
+	}
+}
+
+// isEventEqual returns true if th eevent is shallowly equal
```
karlkfi, line 214: eevent typo

pkg/apply/mutator/mutator.go
```diff
@@ -0,0 +1,20 @@
...
+
+package mutator
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// Interface decouples apply-time-mutation
+// from the concrete structs used for applying.
+type Interface interface {
```
thackem, line 12: nit: consider a different name, like "Mutator"

pkg/apply/mutator/apply-time-mutator.go
```diff
@@ -0,0 +1,216 @@
...
+			return mutated, reason, errors.Wrapf(err, "failed to read field from source resource: %v", sourceRef)
+		}
+
+		var newValue interface{}
+		if sub.Token == "" {
+			// token not specified, replace the entire target value with the source value
+			newValue = sourceValue
+		} else {
+			// token specified, substitute token for source field value in target field value
+			targetValueString, ok := targetValue.(string)
```
thackem, line 89: wdyt about doing some type assertions for templated destinations,?
they should probably all be strings... or just let it fall through

pkg/apply/mutator/apply-time-mutator.go
```diff
@@ -0,0 +1,216 @@
...
+
+// Name returns a mutator identifier for logging.
+func (atm *ApplyTimeMutator) Name() string {
+	return "ApplyTimeMutator"
+}
+
+// Mutate parses the apply-time-mutation annotation and loops through the
+// substitutions, applying each of them to the supplied target object.
+// Returns true with a reason, if mutation was performed.
+func (atm *ApplyTimeMutator) Mutate(obj *unstructured.Unstructured) (bool, string, error) {
```
thackem, line 43: is there a test of this?

pkg/logutil/logutil.go
```diff
@@ -0,0 +1,34 @@
...
+func WrapGoLogging() {
+	logging.SetBackend(klogGoLoggingBackend{})
+}
+
+type klogGoLoggingBackend struct{}
+
+func (k klogGoLoggingBackend) Log(level logging.Level, calldepth int, rec *logging.Record) error {
+	switch level {
+	case logging.CRITICAL:
+		klog.ErrorDepth(calldepth+1, rec.Message())
```
karlkfi, line 21: Anyone know how to set both verbosity level and call depth?
It's not clear how some of these go-logger levels should be mapped to klog levels.

go.mod
```diff
@@ -3,12 +3,16 @@ module sigs.k8s.io/cli-utils
...
+	github.com/google/go-cmp v0.5.6
 	github.com/google/uuid v1.2.0
+	github.com/mikefarah/yq/v4 v4.12.1
 	github.com/onsi/ginkgo v1.16.2
 	github.com/onsi/gomega v1.12.0
+	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.1.3
 	github.com/stretchr/testify v1.7.0
-	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
+	gopkg.in/op/go-logging.v1 v1.0.0-20160211212156-b2cb9fa56473
```
ash2k, line 14: Please don't add any logging dependencies to this library. As a consumer, I'd like the decision of what logger implementation to use to be left to me.

go.mod
```diff
@@ -3,12 +3,16 @@ module sigs.k8s.io/cli-utils
 go 1.16
 
 require (
+	github.com/google/go-cmp v0.5.6
 	github.com/google/uuid v1.2.0
+	github.com/mikefarah/yq/v4 v4.12.1
 	github.com/onsi/ginkgo v1.16.2
 	github.com/onsi/gomega v1.12.0
+	github.com/pkg/errors v0.9.1
```
ash2k, line 11: Please don't add a dependency on this package as it's not really necessary. I've removed it recently in https://github.com/kubernetes-sigs/cli-utils/pull/380

pkg/apply/mutator/mutator.go
karlkfi: Golint complains about stutter on exported functions and types:
- https://github.com/golang/lint/blob/master/lint.go#L483
- https://github.com/golang/lint/blob/master/lint.go#L938

EX: sort.Interface instead of sort.Sort: https://pkg.go.dev/sort#Interface

I think this used to be in Effective Go, but now that I'm looking for it, I can't find it in the latest copy.

It was in older versions tho:
- https://www.rose-hulman.edu/class/csse/csse403/Fall2010/SlidePDFs/EffectiveGo.pdf
- https://www.mit.edu/afs.new/sipb/project/golang/doc/effective_go.html#names

pkg/apply/mutator/apply-time-mutator.go
karlkfi: The impl here errors if the token is specified and the target field isn't a string.

Having token be optional allows for maps, lists, ints, floats, and bools to be set wholesale, without string substitution.

pkg/apply/mutator/apply-time-mutator.go
karlkfi: not yet. it's on the TODO list, but I wanted to make sure I was on the righ track with impl and wasn't gonna have to refactor and write all new tests.

go.mod
karlkfi: go-logging is used by the yq impl that's imported.

I added a logutil package to allow configuring go-logging to use klog as a backend.

go.mod
karlkfi: Gotcha. Will remove.

go.mod
ash2k: Oh, I see.

https://github.com/op/go-logging/blob/b2cb9fa56473e98db8caba80237377e83fe44db5/backend.go#L7-L28 the package field that `logging.SetBackend()` sets is not protected by a mutex so this method is only safe to call from `init()` or `main()` before any package has a chance to start any goroutines. Otherwise there is a risk of a data race.

I suggest to leave it out and let consumers do something about it if they want to. But whatever they do, it should be in `main()` or in an `init()`.

For tests (if we really need to set the backend for them?), it should be possible to set a backend in [`TestMain()`](https://pkg.go.dev/testing#hdr-Main) - that should be safe. But I also suggest to leave it out if possible to avoid making this logging library a direct dependency. Having it as a transitive dependency is unavoidable if we need to use this library, but it doesn't have to be a direct one. WDYT?

go.mod
ash2k: Thanks!

go.mod
karlkfi: Forgive my ignorance, but if the defaultBackend is only set once by SetBackend and not accessed beforehand, why does it need a mutex? What's the race?

> For tests (if we really need to set the backend for them?), it should be possible to set a backend in TestMain()

Yes, I could add a TestMain for the test packages that execute yq code. It's only run once per package, not per test, but that should be fine as long as we're not testing the logs themselves.

> Having it as a transitive dependency is unavoidable if we need to use this library, but it doesn't have to be a direct one. WDYT?

If we decide to accept this yq dependency, then my assumption is that every client that uses cli-utils will need to wrap go-logging with klog, otherwise the logging defaults to trace level on stderr, with a default format, and yq logs extensively. This assumes all cl-utils clients use the apply package and want to support apply-time-mutation, but I suspect that's a 100% overlap.

I also expect almost all cli-utils clients use klog for logging, because it's the Kubernetes standard logger. But I'm willing to be contradicted on that point. I don't know of any full list of usages.

I'm not trying to be stickler, here. I totally understand not wanting to add a new log dependency. I want to do the right thing. But if on of cli-utils's dependencies starts logging to another logger, I would expect that clients will want to be warned in release notes and provided with a workaround.

I'm willing to consider moving the logutil to its own package tho, and depending on that for tests and recommending it to clients. But making a new OSS repo means I'll need to go through a corporate process for approval. So I wanna make sure it's the best option first.

ash2k: > but if the defaultBackend is only set once by SetBackend and not accessed beforehand, why does it need a mutex? What's the race?

We (from the point of view of this library) don't know if anything else in the user's program is using the same logging library. There may be read accesses to that field from other goroutines. To make it safe, there must be memory ordering guarantees between the write and any subsequent reads. To have such guarantees, the library user must ensure the write happened before those goroutines are started. The only way to achieve this reliably is to perform the write in `main()` or `init()` - before any goroutines are started. Does it explain my concern? I'm not saying we need to wrap it with a mutex - only the logging library author could do that, but they chose not to do that (which is fine, but it puts the above constraints on consumers).

I now see why you've done what you've done, it makes sense to me. I suggest to export the `klogGoLoggingBackend` type and remove the `WrapGoLogging()` function. I don't think providing it is very valuable, I think having the `logging.SetBackend()` call in the user's program would be better because it's more explicit. Perhaps we could add a note somewhere that it must be called.

> I'm willing to consider moving the logutil to its own package tho

(I think you mean a separate repo/module) I think it's not worth it. Having it as a separate package (not a repo) is fine in my opinion.

ash2k:
From the description:

> [Set]NestedField in github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/unstructured/helpers.go#L208 (no array indexing)

I'm curios to learn more about this option. Maybe we can improve those helpers instead of adding the yq dependency?

p.s. This feature is something I've built before in [Smith](https://github.com/atlassian/smith). See [the docs](https://github.com/atlassian/smith/blob/master/docs/design/field-references.md). I didn't need a jq/yq library, just used unstructured and helpers (I'm sure it's less sophisticated, but it worked ok).

mortent (commented):
I like the direction and structure here.

pkg/apply/mutator/apply_time_mutator.go
```diff
@@ -0,0 +1,216 @@
...
+
+	// lookup resource mapping
+	mapping, err := atm.Mapper.RESTMapping(sourceGvk.GroupKind(), versions...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to map resource reference to valid type (%v): %w", ref, err)
+	}
+
+	// lookup resource from cluster
+	namespacedClient := atm.Client.Resource(mapping.Resource).Namespace(ref.Namespace)
+	obj, err := namespacedClient.Get(context.TODO(), ref.Name, metav1.GetOptions{})
```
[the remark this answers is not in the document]
karlkfi: Aye. I should make Context part of the Mutator API. Tho I see the Filter API also needs the same treatment, as does the Prune task.

pkg/apply/task/apply_task.go
```diff
@@ -132,6 +134,26 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
 			if filtered || filterErr != nil {
 				continue
 			}
+			// Execute mutators, in case the resource needs mutation.
+			for _, mutator := range a.Mutators {
```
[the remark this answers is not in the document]
karlkfi: There's no filter tests either. I did my best to try to make the changes minimal here so I could test the feature independently, but you're right that higher level testing would be good too.

karlkfi:
> I'm curios to learn more about these options. Maybe we can improve (one of) them instead of adding the yq dependency?
> 
> p.s. This feature is something I've built before in [Smith](https://github.com/atlassian/smith). See [the docs](https://github.com/atlassian/smith/blob/master/docs/design/field-references.md). I didn't need a jq/yq library, just used unstructured and helpers (I'm sure it's less sophisticated, but it worked ok).

@ash2k I spent most of the week trying to fork and modify jsonpath impls to add setter capability and failed miserably. It may be possible, but it's beyond my ability to do quickly. 

This apply-time-mutation is a big blocker for using infrastructure operators, like Config Connector, which need a little hand-holding to orchestrate ordering. And it should work well as a fallback for orchestrating multiple operators that haven't been designed to integrate. So I didn't want to block this on writing a new jsonpath impl.

My initial implementation here did use [Set]NestedFields from apimachinery, but I realized that while it worked for basic use cases, it wouldn't work for anything related to Pod containers, because it can't index arrays. So I needed something more powerful.

yq was my 2nd choice, because everyone loves jq, and the new yq v4 uses a very similar (tho poorly documented) dialect.

That said, I also found https://github.com/spyzhov/ajson yesterday and it does support jsonpath setters. But it's less popular and less well maintained than yq. Also, JSONPath has as many variations as implementations. On the plus side, it's syntax is meticulously tested (https://cburgmer.github.io/json-path-comparison/). 

I'm looking through your field references doc and I don't see any path syntax. Do you have time to jump on a Meet/Zoom and explain what you mean?

go.mod
karlkfi: Exporting the backend struct seems reasonable. Removing the WrapGoLogging function won't remove the go-logging dependency, but I can move `logging.SetBackend(klogGoLoggingBackend{})` to the struct comment and call ti directly from TestMain, since it's a one-liner.

karlkfi: Ok. I've done a bit more than that and made KlogGoLoggingBackend more flexible with configurable log and verbosity mappings. Added comments to explain why and how to use.

I've also validated it works as expected by using it in the kyq tests with TestMain.

```
go test sigs.k8s.io/cli-utils/pkg/kyq -v -args -v=5
```

pkg/apply/mutator/apply_time_mutator.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	targetID := object.UnstructuredToObjMetaOrDie(obj)
+	klog.V(5).Infof("target resource %s/%s: %#v", targetID.Namespace, targetID.Name, obj)
+
+	for _, sub := range subs {
+		sourceRef := sub.SourceRef
+
+		// lookup source resource from cache or cluster
+		sourceObj, err := atm.getObject(sourceRef)
```
thackem, line 65: Maybe I missed it, but I don't see a check for object status in getObject or here, is this dependent fully on the dependsOn implementation to assert readiness prior to fetching? will that make it more challenging in the future to move to supporting mutation references to objects outside the apply set?

karlkfi: The depends-on impl added a `graph.SortObjs` that handles ordering of apply stages with wait tasks in between. I extended the SortObjs impl to inject the source resources from apply-time-mutation into the graph:
https://github.com/karlkfi/cli-utils/blob/karl-apply-time-mutation/pkg/object/graph/sort.go#L33

So by the time the mutator is executed in the apply task, the waits should have already happened.

There might be some edge cases with resources not in the inventory. So I'll need to write some tests that exercise both pieces together. But I'm not sure where to put them yet in cli-utils.

pkg/apply/cache/resource_caching_poller_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+	"github.com/google/go-cmp/cmp"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/cli-utils/pkg/apply/cache/testutil"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/polling"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
+	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+var o1 = &unstructured.Unstructured{
```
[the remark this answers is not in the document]
karlkfi: will do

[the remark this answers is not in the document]
karlkfi: done

pkg/logutil/logutil.go
karlkfi: figured it out and updated the lib to support configurable mapping

pkg/apply/mutator/apply-time-mutator.go
karlkfi: added mutate tests:
https://github.com/karlkfi/cli-utils/blob/karl-apply-time-mutation/pkg/apply/mutator/apply_time_mutator_test.go

and kyq tests:
https://github.com/karlkfi/cli-utils/blob/karl-apply-time-mutation/pkg/kyq/kyq_test.go

pkg/apply/mutator/apply_time_mutator.go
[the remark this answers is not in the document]
karlkfi: I routed it through Mutate, Filter, and Prune, but I stopped short of changing the Task API or adding new cmd timeouts. Those probably need to be done seperately.

pkg/apply/cache/resource_caching_poller.go
```diff
@@ -0,0 +1,63 @@
...
+	// Assume the wrapped Poller will close the output channel when the context is done.
+	// This avoids losing any events sent after the context is done.
+
+	go func() {
+		// close the output channel on exit to signal completion to the caller
+		defer close(outCh)
+
+		for {
+			// Process events from the inCh and send to the outCh
+			event, ok := <-inCh
```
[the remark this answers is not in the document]
karlkfi: done

karlkfi:
Alright, this is ready for serious review.

I've added tests for ApplyTimeMutaiton that I'm satisfied with.

I haven't added any Mutate tests to the ApplyTask, but there aren't any for Filter either. So we'll probably need to backfill those later. It might require abstracting/refactoring the ApplyTask some more.

The biggest open question is whether we're satisfied with `yq` as the path selector expression language. 

If `yq` is too big of a lift, I can try experimenting with github.com/spyzhov/ajson as an alternative, but it's a one person project with no commits in the last year.

There's also an option to support JSON Path later, in addition to yq, requiring JSON Path expressions to start with `$.`. This would at least allow some overlap, if we decide to deprecate yq later.

pkg/apply/cache/resource_cache.go
```diff
@@ -0,0 +1,24 @@
...
+
+package cache
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+// ResourceCache stores unstructured resource objects in memory
+type ResourceCache interface {
```
ash2k, line 12: Are implementations supposed to be thread safe? I can see that it's being written to from one goroutine and read from another one. The `ResourceCacheMap` implementation is not thread safe and there appears to be no synchronization around it.

go.mod
ash2k: 👍

ash2k:
@karlkfi 

> I'm looking through your field references doc and I don't see any path syntax. Do you have time to jump on a Meet/Zoom and explain what you mean?

It's very basic, only nested element support IIRC. It's not good enough as a generic solution. I was just sharing it to show that there is prior art there. Speaking of which...

> This apply-time-mutation is a big blocker for using infrastructure operators, like Config Connector, which need a little hand-holding to orchestrate ordering. And it should work well as a fallback for orchestrating multiple operators that haven't been designed to integrate. So I didn't want to block this on writing a new jsonpath impl.

I totally see the need for this functionality (that's why I built Smith). This is something many people in the community need and many will use, so I wonder if there should be a wider and more visible discussion about it? I found these issues:

- https://github.com/kubernetes/kubernetes/issues/34363
- https://github.com/kubernetes/kubernetes/issues/1768
- https://github.com/kubernetes/community/pull/176

The feature seem to be too significant to me to just discuss what syntax to pick here in the comments with a tiny group of people, without anyone from potentially interested SIGs - arch, apps, apimachinery, ... Maybe it has been discussed somewhere already (I'm not participating in any meetings, so could have missed it)?

pkg/apply/cache/resource_cache.go
karlkfi: TIL the builtin Go map isn't thread safe. I'll go add some locks.

pkg/apply/mutator/apply_time_mutator.go
```diff
@@ -67,19 +79,27 @@ func (atm *ApplyTimeMutator) Mutate(obj *unstructured.Unstructured) (bool, strin
 			return mutated, reason, fmt.Errorf("failed to get source resource (%v): %w", sourceRef, err)
 		}
 
-		sourceID := object.UnstructuredToObjMetaOrDie(sourceObj)
-		klog.V(5).Infof("source resource %s/%s: %#v", sourceID.Namespace, sourceID.Name, sourceObj)
+		if klog.V(4).Enabled() {
+			yamlBytes, err := yaml.Marshal(sourceObj)
```
justinsb, line 83: I often define a function called `DebugJSON` or something like that that formats the object or returns a string representation of the error message.  You normally want to avoid changing the code flow based on logging, so you don't want to return.

pkg/apply/cache/resource_cache.go
karlkfi: Fixed.

pkg/apply/mutator/apply_time_mutator.go
karlkfi: Good catch.
Fixed to log the error.

go.sum
```diff
@@ -132,6 +132,8 @@ github.com/eapache/queue v1.1.0/go.mod h1:6eCeP0CKFpHLu8blIFXhExK/dRa7WDZfr6jVFP
 github.com/edsrzf/mmap-go v1.0.0/go.mod h1:YO35OhQPt3KJa3ryjFM5Bs14WD66h8eGKpfaBNrHW5M=
 github.com/elazarl/goproxy v0.0.0-20180725130230-947c36da3153 h1:yUdfgN0XgIJw7foRItutHYUIhlcKzcSf5vDpdhQAKTc=
 github.com/elazarl/goproxy v0.0.0-20180725130230-947c36da3153/go.mod h1:/Zj4wYkgs4iZTTu3o/KG3Itv/qCCa8VVMlb3i9OVuzc=
+github.com/elliotchance/orderedmap v1.4.0 h1:wZtfeEONCbx6in1CZyE6bELEt/vFayMvsxqI5SgsR+A=
```
seans3, line 135: Our initial effort should not contain any of these dependencies. Please remove them.

pkg/apply/applier.go
```diff
@@ -211,7 +233,7 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
 		// Create a new TaskStatusRunner to execute the taskQueue.
 		klog.V(4).Infoln("applier building TaskStatusRunner...")
 		allIds := object.UnstructuredsToObjMetasOrDie(append(applyObjs, pruneObjs...))
-		runner := taskrunner.NewTaskStatusRunner(allIds, a.statusPoller)
+		runner := taskrunner.NewTaskStatusRunner(allIds, cachingPoller)
```
seans3, line 236: This caching poller is not necessary--please revert to the previous status poller. If you want to advocate for a different poller, please do that in another PR.

pkg/apply/solver/solver.go
```diff
@@ -157,12 +158,13 @@ func (t *TaskQueueBuilder) AppendDeleteInvTask(inv inventory.InventoryInfo, dryR
 // AppendInvAddTask appends a task to the task queue to apply the passed objects
 // to the cluster. Returns a pointer to the Builder to chain function calls.
 func (t *TaskQueueBuilder) AppendApplyTask(applyObjs []*unstructured.Unstructured,
-	applyFilters []filter.ValidationFilter, o Options) *TaskQueueBuilder {
+	applyFilters []filter.ValidationFilter, applyMutators []mutator.Interface, o Options) *TaskQueueBuilder {
```
seans3, line 161: Yes. This is a productive approach.

pkg/apply/solver/solver_test.go
```diff
@@ -378,7 +384,7 @@ func TestTaskQueueBuilder_AppendApplyWaitTasks(t *testing.T) {
 					actWaitTask := toWaitTask(t, actualTask)
 					assert.Equal(t, len(expTsk.Ids), len(actWaitTask.Ids))
 					// Order is NOT important for ids stored within task.
-					if !object.SetEquals(expTsk.Ids, actWaitTask.Ids) {
+					if !object.IsSetEqual(expTsk.Ids, actWaitTask.Ids) {
```
seans3, line 387: Remove this change. If you want to advocate for a name change, please do it in another PR.

pkg/apply/solver/solver_test.go
```diff
@@ -644,7 +650,7 @@ func TestTaskQueueBuilder_AppendPruneWaitTasks(t *testing.T) {
 				case *taskrunner.WaitTask:
 					actWaitTask := toWaitTask(t, actualTask)
 					assert.Equal(t, len(expTsk.Ids), len(actWaitTask.Ids))
-					if !object.SetEquals(expTsk.Ids, actWaitTask.Ids) {
+					if !object.IsSetEqual(expTsk.Ids, actWaitTask.Ids) {
```
seans3, line 653: Remove this change. If you want to advocate for a name change, please do it in another PR.

pkg/apply/task/apply_task.go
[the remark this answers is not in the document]
seans3: Nit. Maybe pull this into an `ApplyTask` function, since this is already very long.

pkg/apply/task/apply_task_test.go
```diff
@@ -114,7 +114,7 @@ func TestApplyTask_BasicAppliedObjects(t *testing.T) {
 			require.NoError(t, err)
 
 			actual := taskContext.AppliedResources()
-			if !object.SetEquals(expected, actual) {
+			if !object.IsSetEqual(expected, actual) {
```
seans3, line 117: Remove this change. If you want to advocate for a name change, please do it in another PR.

pkg/apply/task/inv_add_task_test.go
```diff
@@ -122,7 +122,7 @@ func TestInvAddTask(t *testing.T) {
 			applyIds, err := object.UnstructuredsToObjMetas(tc.applyObjs)
 			require.NoError(t, err)
 
-			if !object.SetEquals(applyIds, task.Identifiers()) {
+			if !object.IsSetEqual(applyIds, task.Identifiers()) {
```
seans3, line 125: Remove this change. If you want to advocate for a name change, please do it in another PR.

pkg/apply/task/inv_set_task_test.go
```diff
@@ -134,7 +134,7 @@ func TestInvSetTask(t *testing.T) {
 				t.Errorf("unexpected error running InvAddTask: %s", result.Err)
 			}
 			actual, _ := client.GetClusterObjs(nil, common.DryRunNone)
-			if !object.SetEquals(tc.expectedObjs, actual) {
+			if !object.IsSetEqual(tc.expectedObjs, actual) {
```
seans3, line 137: Remove this change. If you want to advocate for a name change, please do it in another PR.

pkg/inventory/inventory-client.go
```diff
@@ -121,7 +121,7 @@ func (cic *ClusterInventoryClient) Merge(localInv InventoryInfo, objs []object.O
 		if err != nil {
 			return pruneIds, err
 		}
-		if object.SetEquals(objs, clusterObjs) {
+		if object.IsSetEqual(objs, clusterObjs) {
```
seans3, line 124: Remove this change. If you want to advocate for a name change, please do it in another PR.

pkg/inventory/inventory-client.go
```diff
@@ -160,7 +160,7 @@ func (cic *ClusterInventoryClient) Replace(localInv InventoryInfo, objs []object
 	if err != nil {
 		return err
 	}
-	if object.SetEquals(objs, clusterObjs) {
+	if object.IsSetEqual(objs, clusterObjs) {
```
seans3, line 163: Remove this change. If you want to advocate for a name change, please do it in another PR.

pkg/inventory/inventory-client_test.go
```diff
@@ -82,7 +82,7 @@ func TestGetClusterInventoryInfo(t *testing.T) {
 				if err != nil {
 					t.Fatalf("unexpected error received: %s", err)
 				}
-				if !object.SetEquals(tc.localObjs, clusterObjs) {
+				if !object.IsSetEqual(tc.localObjs, clusterObjs) {
```
seans3, line 85: Remove this change. If you want to advocate for a name change, please do it in another PR.

pkg/inventory/inventory-client_test.go
```diff
@@ -180,7 +180,7 @@ func TestMerge(t *testing.T) {
 				if !tc.isError && err != nil {
 					t.Fatalf("unexpected error: %s", err)
 				}
-				if !object.SetEquals(tc.pruneObjs, pruneObjs) {
+				if !object.IsSetEqual(tc.pruneObjs, pruneObjs) {
```
seans3, line 183: Remove this change. If you want to advocate for a name change, please do it in another PR.

pkg/inventory/inventory-client_test.go
```diff
@@ -340,7 +340,7 @@ func TestReplace(t *testing.T) {
 			if err != nil {
 				t.Fatalf("unexpected error received: %s", err)
 			}
-			if !object.SetEquals(tc.localObjs, actualObjs) {
+			if !object.IsSetEqual(tc.localObjs, actualObjs) {
```
seans3, line 343: Remove this change. If you want to advocate for a name change, please do it in another PR.

pkg/inventory/inventory-client_test.go
```diff
@@ -397,7 +397,7 @@ func TestGetClusterObjs(t *testing.T) {
 			if !tc.isError && err != nil {
 				t.Fatalf("unexpected error received: %s", err)
 			}
-			if !object.SetEquals(tc.clusterObjs, clusterObjs) {
+			if !object.IsSetEqual(tc.clusterObjs, clusterObjs) {
```
seans3, line 400: Remove this change. If you want to advocate for a name change, please do it in another PR.

pkg/inventory/inventory-client_test.go
```diff
@@ -571,7 +571,7 @@ func TestMergeInventoryObjs(t *testing.T) {
 				}
 				wrapped := WrapInventoryObj(retained)
 				mergedObjs, _ := wrapped.Load()
-				if !object.SetEquals(tc.expected, mergedObjs) {
+				if !object.IsSetEqual(tc.expected, mergedObjs) {
```
seans3, line 574: Remove this change. If you want to advocate for a name change, please do it in another PR.

pkg/kstatus/polling/engine/engine_test.go
```diff
@@ -16,10 +16,9 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
-	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/testutil"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/cli-utils/pkg/object"
-	fakemapper "sigs.k8s.io/cli-utils/pkg/testutil"
+	"sigs.k8s.io/cli-utils/pkg/testutil"
```
seans3, line 21: Please make this unrelated change in another PR.

pkg/kstatus/polling/engine/engine_test.go
```diff
@@ -103,7 +102,7 @@ func TestStatusPollerRunner(t *testing.T) {
 
 			identifiers := tc.identifiers
 
-			fakeMapper := fakemapper.NewFakeRESTMapper(
+			fakeMapper := testutil.NewFakeRESTMapper(
```
seans3, line 105: Please make this unrelated change in another PR.

pkg/kstatus/polling/engine/engine_test.go
```diff
@@ -187,7 +186,7 @@ func TestNewStatusPollerRunnerIdentifierValidation(t *testing.T) {
 	}
 
 	engine := PollerEngine{
-		Mapper: fakemapper.NewFakeRESTMapper(
+		Mapper: testutil.NewFakeRESTMapper(
```
seans3, line 189: Please make this unrelated change in another PR.

pkg/kstatus/polling/statusreaders/common_test.go
```diff
@@ -15,9 +15,8 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/testutil"
 	"sigs.k8s.io/cli-utils/pkg/object"
-	fakemapper "sigs.k8s.io/cli-utils/pkg/testutil"
+	"sigs.k8s.io/cli-utils/pkg/testutil"
```
seans3, line 19: Please make this unrelated change in another PR.

pkg/kstatus/polling/statusreaders/common_test.go
```diff
@@ -74,7 +73,7 @@ func TestLookupResource(t *testing.T) {
 			fakeReader := &fakeClusterReader{
 				getErr: tc.readerErr,
 			}
-			fakeMapper := fakemapper.NewFakeRESTMapper(deploymentGVK)
+			fakeMapper := testutil.NewFakeRESTMapper(deploymentGVK)
```
seans3, line 76: Please make this unrelated change in another PR.

pkg/kstatus/polling/statusreaders/common_test.go
```diff
@@ -199,7 +198,7 @@ spec:
 				},
 				listErr: tc.listErr,
 			}
-			fakeMapper := fakemapper.NewFakeRESTMapper(rsGVK)
+			fakeMapper := testutil.NewFakeRESTMapper(rsGVK)
```
seans3, line 201: Please make this unrelated change in another PR.

pkg/kstatus/polling/statusreaders/generic_test.go
```diff
@@ -11,10 +11,9 @@ import (
 	"github.com/stretchr/testify/assert"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/testutil"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/cli-utils/pkg/object"
-	fakemapper "sigs.k8s.io/cli-utils/pkg/testutil"
+	"sigs.k8s.io/cli-utils/pkg/testutil"
```
seans3, line 16: Please make this unrelated change in another PR.

pkg/kstatus/polling/statusreaders/generic_test.go
```diff
@@ -60,7 +59,7 @@ func TestGenericStatusReader(t *testing.T) {
 	for tn, tc := range testCases {
 		t.Run(tn, func(t *testing.T) {
 			fakeReader := testutil.NewNoopClusterReader()
-			fakeMapper := fakemapper.NewFakeRESTMapper()
+			fakeMapper := testutil.NewFakeRESTMapper()
```
seans3, line 62: Please make this unrelated change in another PR.

pkg/kstatus/polling/statusreaders/pod_controller_test.go
```diff
@@ -12,10 +12,9 @@ import (
 	v1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
-	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/testutil"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/cli-utils/pkg/object"
-	fakemapper "sigs.k8s.io/cli-utils/pkg/testutil"
+	"sigs.k8s.io/cli-utils/pkg/testutil"
```
seans3, line 17: Please make this unrelated change in another PR.

pkg/kstatus/polling/statusreaders/pod_controller_test.go
```diff
@@ -79,7 +78,7 @@ func TestPodControllerStatusReader(t *testing.T) {
 	for tn, tc := range testCases {
 		t.Run(tn, func(t *testing.T) {
 			fakeReader := testutil.NewNoopClusterReader()
-			fakeMapper := fakemapper.NewFakeRESTMapper()
+			fakeMapper := testutil.NewFakeRESTMapper()
```
seans3, line 81: Please make this unrelated change in another PR.

pkg/kstatus/polling/statusreaders/testing.go
```diff
@@ -12,8 +12,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/engine"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
-	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/testutil"
 	"sigs.k8s.io/cli-utils/pkg/object"
+	"sigs.k8s.io/cli-utils/pkg/testutil"
```
seans3, line 16: Please make this unrelated change in another PR.

pkg/kstatus/polling/testutil/fake_poller.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2021 The Kubernetes Authors.
```
seans3, line 1: This is unnecessary--please remove this file.

pkg/kyq/kyq.go
```diff
@@ -0,0 +1,176 @@
+// Copyright 2021 The Kubernetes Authors.
```
seans3, line 1: The `yq` package is unnecessary. Please remove.

pkg/logutil/logutil.go
```diff
@@ -0,0 +1,70 @@
+// Copyright 2021 The Kubernetes Authors.
```
seans3, line 1: This file is unnecessary--please remove it.

pkg/object/graph/graph.go
```diff
@@ -162,3 +162,34 @@ func (cde CyclicDependencyError) Error() string {
 	}
 	return errorBuf.String()
 }
+
+// EdgeSet is a list wrapper that allows set equality comparison.
```
seans3, line 166: The EdgeSet code is unnecessary. If you want to advocate for this minor change, please do it in another PR.

pkg/object/graph/graph_test.go
```diff
@@ -116,7 +116,7 @@ func TestObjectGraphSort(t *testing.T) {
 				}
 				for i, actualSet := range actual {
 					expectedSet := tc.expected[i]
-					if !object.SetEquals(expectedSet, actualSet) {
+					if !object.IsSetEqual(expectedSet, actualSet) {
```
seans3, line 119: Remove from this PR.

pkg/object/mutation/annotation.go
```diff
@@ -0,0 +1,69 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	Annotation = "config.kubernetes.io/apply-time-mutation"
+)
+
+func HasAnnotation(u *unstructured.Unstructured) bool {
```
seans3, line 19: All these functions appear to duplicate `dependson` code.

pkg/object/objmetadata.go
```diff
@@ -261,3 +269,9 @@ func SetEquals(setA []ObjMetadata, setB []ObjMetadata) bool {
 	}
 	return true
 }
+
+// IsSetEqual returns true if the slice of objects in setA equals
+// the slice of objects in setB.
+func IsSetEqual(setA []ObjMetadata, setB []ObjMetadata) bool {
```
seans3, line 275: Please remove this from this PR.

pkg/object/objmetadata_test.go
```diff
@@ -386,7 +386,7 @@ func TestSetDiff(t *testing.T) {
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
 			actual := SetDiff(tc.setA, tc.setB)
-			if !SetEquals(tc.expected, actual) {
+			if !IsSetEqual(tc.expected, actual) {
```
seans3, line 389: Please remove this. If you want this minor name change, please advocate for it in another PR.

pkg/object/objmetadata_test.go
```diff
@@ -434,7 +434,7 @@ func TestUnion(t *testing.T) {
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
 			actual := Union(tc.setA, tc.setB)
-			if !SetEquals(tc.expected, actual) {
+			if !IsSetEqual(tc.expected, actual) {
```
seans3, line 437: Please remove this. If you want this minor name change, please advocate for it in another PR.

pkg/object/objmetadata_test.go
```diff
@@ -481,7 +481,7 @@ func TestSetEquals(t *testing.T) {
 
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
-			actual := SetEquals(tc.setA, tc.setB)
+			actual := IsSetEqual(tc.setA, tc.setB)
```
seans3, line 484: Please remove this. If you want this minor name change, please advocate for it in another PR.

seans3 (changes requested):
Thanks for taking the initiative on this functionality. This PR is amazing progress. I think the high level approach looks great. There appears to be a significant amount of unnecessary code. Since this PR is already 64 files, we should be ruthless in pulling out unnecessary complexity from the initial implementation.

The following should be removed:
- The `ResourceCache` and the new poller. We already store the applied objects in the `TaskContext` which is in the same execution goroutine. We do not need a multi-threaded cache to start. We need to simply store the applied `Unstructured` object in the `TaskContext` instead of just the ID `ObjMetadata`.
- The `yq` library and all its dependencies and the new logging code. As I mentioned in person previously, we should use the `Unstructured` `Nested` functions to initially create the MVP of this functionality. If we need a full library in the future, we can add it then, but our initial implementation will be the simplest that works.
- Any changes unrelated to this functionality should be moved to separate PR's. Specifically, the unrelated naming changes should be removed from this PR.

To simplify this PR even more, we should identify functionality that we can merge immediately that is useful.
- The context changes are generally useful can be merged immediately
- The `dependson` file move can be merged immediately in a separate PR.

Again, thanks for taking the initiative on this important functionality.

karlkfi:
> [Remove]  The ResourceCache and the new poller. We already store the applied objects in the TaskContext which is in the same execution goroutine. ... We need to simply store the applied Unstructured object in the TaskContext instead of just the ID ObjMetadata.

I can move the cache to a subsequent PR. But without the cache, the mutator will need to do the getObject from the server every time. The Task Context does not currently store objects, just the apply status. The event includes the object but it wasn't being persisted. 

I originally put the cache in the context, as we had discussed, but I needed to pull it out in order to inject it into the Mutator for multiple reasons:
1. I followed the Filter pattern, which doesn't have access to the TaskContext. So I didn't give the Mutator access either.
2. I wanted to make every GET update the cache, but I couldn't easily wrap getObject because every different client has their own way to construct the client and there was no easy way to inject it. 
3. The Poller is the thing doing most of the GETs, but the poller doesn't have access to the TaskContext either.
4.  I originally added code to the `baserunner.run` func to update the cache, but decide that the runner was already too complex and I wouldn't be able to easy test the change. So I moved it into a Poller wrapper instead, which I think is much cleaner and easier to test.

If you want to replace the status cache in the TaskContext with the ResourceCache, we can do that, with some added methods to extract the status, but I was nervous to make too many reverse incompatible API changes, because it's unclear what the client API surface is or what will break clients.

> We do not need a multi-threaded cache to start. 

@ash2k asked me to make it multi-threaded, because it's used by multiple goroutines, and the Go map is not thread-safe by itself.

> [Remove] The yq library and all its dependencies and the new logging code. As I mentioned in person previously, we should use the Unstructured Nested functions to initially create the MVP of this functionality. If we need a full library in the future, we can add it then, but our initial implementation will be the simplest that works.

The Unstructured Nested functions do not satisfy the requirements of being able to update arrays. They can't index arrays and they don't have array selectors. I've documented the options and pros and cons in go/jsonpath-vs-yq.

 If we add Nested now, it's not clear how would we add support for jsonpath or yq afterwards without breaking users. The syntaxes are similar but not identical.

I'm not married to yq being the best solution, but I'm pretty convinced that the Nested syntax is not viable.

> [Remove] Retrieving objects from the cluster that are referenced in depends-on. 

I can't remove getObject AND the cache.

> Our first implementation only addresses objects which are applied at the same time.

We originally agreed to do this, because it was easier. But as you can see, I implemented the more advanced case already. So difficulty of implementation is not a blocker. I believe this behavior is better for the user. For what reason should we revert to inferior behavior? 

> Any changes unrelated to this functionality should be moved to separate PR's. Specifically, the unrelated naming changes should be removed from this PR. To simplify this PR even more, we should identify functionality that we can merge immediately that is useful.

Sure. I'm totally fine with chopping this up into multiple PRs. I put it all in one initially because making the coding changes is easier than getting consensus and review.

seans3:
> > [Remove]  The ResourceCache and the new poller. We already store the applied objects in the TaskContext which is in the same execution goroutine. ... We need to simply store the applied Unstructured object in the TaskContext instead of just the ID ObjMetadata.
> 
> I can move the cache to a subsequent PR. But without the cache, the mutator will need to do the getObject from the server every time. The Task Context does not currently store objects, just the apply status. The event includes the object but it wasn't being persisted.

Correct. The `TaskContext` does not currently store objects. But that is easily updated. I believe we would only be changing at most 10 lines of code (plus unit tests). And since retrieving objects from the cluster is currently out of scope, we can obviate the risk and complexity of a multi-threaded object cache with a very small amount of single-threaded code.

Your `Mutator` interface would have to be updated to include this `TaskContext` containing all the objects which have been previously applied. An easier path would probably be to drop the `Mutator` interface from this PR, which would simplify the PR even more. We could always add it back in when we have another mutate example to generalize from. But I'll leave that up to you.

> 
> I originally put the cache in the context, as we had discussed, but I needed to pull it out in order to inject it into the Mutator for multiple reasons:
> 
> 1. I followed the Filter pattern, which doesn't have access to the TaskContext. So I didn't give the Mutator access either.
> 2. I wanted to make every GET update the cache, but I couldn't easily wrap getObject because every different client has their own way to construct the client and there was no easy way to inject it.

Retrieving objects from the cluster is beyond the scope of this PR, so fortunately it does not need to be a constraint.

> 3. The Poller is the thing doing most of the GETs, but the poller doesn't have access to the TaskContext either.
> 4. I originally added code to the `baserunner.run` func to update the cache, but decide that the runner was already too complex and I wouldn't be able to easy test the change. So I moved it into a Poller wrapper instead, which I think is much cleaner and easier to test.

Once the `TaskContext` stores the previously applied objects, there is no need to touch the `Poller`.

> 
> If you want to replace the status cache in the TaskContext with the ResourceCache, we can do that, with some added methods to extract the status, but I was nervous to make too many reverse incompatible API changes, because it's unclear what the client API surface is or what will break clients.

Fortunately, the client API does not include the `TaskContext`, so updating it is not a problem.

> 
> > We do not need a multi-threaded cache to start.
> 
> Mikhail Mazurskiy asked me to make it multi-threaded, because it's used by multiple goroutines, and the Go map is not thread-safe by itself.
> 
> > [Remove] The yq library and all its dependencies and the new logging code. As I mentioned in person previously, we should use the Unstructured Nested functions to initially create the MVP of this functionality. If we need a full library in the future, we can add it then, but our initial implementation will be the simplest that works.

I see that there has been quite a bit research and effort into this `jsonpath` like functionality. And I think my initial dismissal of this approach was probably premature and glib.

The downsides I see of the `yq` instead of first attempting the `Nested` functions in `Unstructured` are
- About a dozen extra transitive dependencies that we would be pushing onto all the binaries which include the library. This might not be a huge burden.
- We have to create and maintain our own logging backend to translate `yq` logging into the current `klog`. This also might not be a huge deal.

If the other `cli-utils` stakeholders are not deterred by these downsides, I will get behind this approach.

> 
> The Unstructured Nested functions do not satisfy the requirements of being able to update arrays. They can't index arrays and they don't have array selectors. I've documented the options and pros and cons in go/jsonpath-vs-yq.
> 
> If we add Nested now, it's not clear how would we add support for jsonpath or yq afterwards without breaking users. The syntaxes are similar but not identical.
> 
> I'm not married to yq being the best solution, but I'm pretty convinced that the Nested syntax is not viable.

BTW--the `Nested` functions include `NestedSlice` and `SetNestedSlice` which address arrays. 

> 
> > [Remove] Retrieving objects from the cluster that are referenced in depends-on.
> 
> I can't remove getObject AND the cache.
> 

Fortunately, we can easily remove both.

> > Our first implementation only addresses objects which are applied at the same time.
> 
> We originally agreed to do this, because it was easier. But as you can see, I implemented the more advanced case already. So difficulty of implementation is not a blocker. I believe this behavior is better for the user. For what reason should we revert to inferior behavior?
>

It's not difficulty of implementation as much as the risk of adding unneeded complexity. We still haven't determined how we are going to deal with `depends-on` functionality for resources outside the current apply. So we'll address that in the future.
 
> > Any changes unrelated to this functionality should be moved to separate PR's. Specifically, the unrelated naming changes should be removed from this PR. To simplify this PR even more, we should identify functionality that we can merge immediately that is useful.
> 
> Sure. I'm totally fine with chopping this up into multiple PRs. I put it all in one initially because making the coding changes is easier than getting consensus and review.

I totally appreciate that this large PR was created so we could see the algorithm in total and comment on it that way. Again, thanks for all the work. I'm ready to start merging the pieces that are ready as soon as you send them. Please let me know if you'd like to pair-program again to accelerate that process.

karlkfi:
@seans3 Dropping both the ResourceCache and the Mutator interface is going to make this really hard to test. Both the baserunner and the apply task are already too complicated to get good task coverage or add tests. I didn't make the abstractions because I like abstraction. I did it to make the implementations testable with fakes and unit test. I don' t think removing them is simpler or easier or less work. Why do you want them removed?

karlkfi:
Re: putting ResourceCache in TaskContext:

The existing things in TaskContext:
- `appliedResources map[object.ObjMetadata]applyInfo`
- `failedResources map[object.ObjMetadata]struct{}`
- `pruneFailures map[object.ObjMetadata]struct{}`

All of these get updated at different times: after apply success, after apply failure, after prune failure.

I don't know how to combine those with a ResourceCache that gets updated every time there's a successful GET from the poller.

Any new GET, would over-write the cached result from the last apply/prune.

The only way I can think to stuff them all into the same cache would be to add annotations/labels and some way to filter cache queries on annotation/label, and also carry them forward after new GETs with some kind of merge.

Is there some way you envisioned to cache them all together? Would the conservation of memory be worth in increase in complexty?

karlkfi:
> BTW--the Nested functions include NestedSlice and SetNestedSlice which address arrays

Needing to know that the result is an array would increase the API complexity. 

Also, neither of those methods support navigating through an array, only an array as the result.

Even if we added a "resultType: list" field to the substitution API, it still wouldn't support things like map fields inside an array, which is the most common use of arrays in the Kubernetes and KCC resources. The simplest example is containers in pods/replicasets/deployments. Updating container env vars based on other resource values is a high value use case. Another case is reading from Service and Ingress resources, which have multiple layers of arrays and maps.

pkg/object/mutation/annotation.go
karlkfi: I refactored the dependson code to only return the value for the dependson annotation, rather than being generic and accepting the annotation key string. This way both ba be operated on similarly as if they had a common interface.

pkg/object/mutation/types.go
```diff
@@ -0,0 +1,151 @@
...
+	// +optional
+	Token string `json:"token,omitempty"`
+}
+
+// ResourceReference is a reference to a KRM resource by name and kind.
+// One of APIVersion or Group is required.
+// Group is generally preferred, to avoid needing to update the version in lock
+// step with the referenced resource.
+// If neither is provided, the empty group is used.
+type ResourceReference struct {
```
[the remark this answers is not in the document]
karlkfi: The significant difference is supporting apiVersion OR apiGroup.

I wanted user flexibility. Sometimes you have one, sometimes you have the other. This avoids the user needing to parse and extract the group from the version. This would be valuable for cases like kpt-set that has one and not the other, avoiding needing to make a custom function to parse it.

But it does have a side effect that specifying the version is more restrictive. So if you think it's important I could get on board with only supporting apiGroup.

FWIW, I put some discussion of the inputs and alternatives and prior art in http://go/apply-time-mutation-in-kpt under the "KRM Resource References" header.

karlkfi:
Pulled out dependson refactoring to https://github.com/kubernetes-sigs/cli-utils/pull/402

karlkfi:
Pulled out Context & Timeouts to https://github.com/kubernetes-sigs/cli-utils/pull/404

karlkfi:
I've documented the pros & cons of JSONPath and yq in a public doc:
https://docs.google.com/document/d/1SChsQ9-l13Czpq4YGq7HA2JorZjL5HQ-27CJzwX77yE/edit#

pkg/kstatus/polling/engine/engine_test.go
karlkfi: This is actually related because I had to move the NewFakeRESTMapper in order to avoid a dependency cycle when using it outside of kstatus. But It could be moved just to make this PR smaller.

pkg/kyq/kyq.go
karlkfi: Discussion of jsonpath vs yq is now pubilc:
https://docs.google.com/document/d/1SChsQ9-l13Czpq4YGq7HA2JorZjL5HQ-27CJzwX77yE/edit#

The reason we can't just use the JSONPath lib in client-go is that it doesn't support updates, only retrieval.

pkg/logutil/logutil.go
karlkfi: This logutil is necessary if we decide to keep yq.

If we alternatively decide to import ajson (for it's JSONPath), it uses the built-in go logger, which might need a different logutil wrapper.

pkg/object/graph/depends_test.go
```diff
@@ -9,6 +9,9 @@ import (
 	"github.com/stretchr/testify/assert"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/cli-utils/pkg/object"
+	depsutil "sigs.k8s.io/cli-utils/pkg/object/dependson/testutil"
```
karlkfi, line 12: This refactor helps mitigate an import cycle and mirrors the pattern used for the dependson and applyt-time-mutation mutators (with an unfortunate naming overlap).

pkg/object/mutation/annotation.go
karlkfi: "appear to duplicate" is because they're using the same conventional interface for consistency with different implementations.

pkg/object/mutation/types.go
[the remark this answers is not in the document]
karlkfi: ObjMetadata is exclusively used internally. ResourceReference is an external facing API.

Should they really be made the same thing?

Should the API not optionally accept APIVersion?

I'm thinking about renaming "group" to "apiGroup" to avoid confusion and be consistent with the [Cluster]RoleBinding and PersistentVolumeClaim APIs (tho others use "group"). If I do that, should that affect ObjMetadata too?

pkg/apply/task/apply_task.go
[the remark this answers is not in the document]
karlkfi: done

pkg/apply/mutator/apply_time_mutator.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	if mutated {
+		klog.V(4).Infof("mutated target resource (%v):\n%s", targetRef, yamlStringer{obj})
+	}
+
+	return mutated, reason, nil
+}
+
+// getObject returns a cached resource, if cached and cache exists, otherwise
```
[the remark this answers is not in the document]
karlkfi: based on our face to face discussion, we decided to keep the getObject here.

pkg/object/objmetadata.go
```diff
@@ -240,19 +245,22 @@ func Union(setA []ObjMetadata, setB []ObjMetadata) []ObjMetadata {
 	return union
 }
 
-// SetEquals returns true if the slice of objects in setA equals
-// the slice of objects in setB.
-func SetEquals(setA []ObjMetadata, setB []ObjMetadata) bool {
-	mapA := make(map[ObjMetadata]struct{}, len(setA))
-	for _, a := range setA {
-		mapA[a] = struct{}{}
+// ObjMetadataSet is a list wrapper that allows set equality comparison.
```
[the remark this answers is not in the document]
karlkfi: done. It's now DependencySet, as the interface from depends-on, and only used by that feature. I left the set comparison for other tests.

cmd/apply/cmdapply.go
```diff
@@ -147,7 +147,11 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	ch := a.Run(context.Background(), inv, objs, apply.Options{
+
+	// TODO: add apply command timeout
+	ctx := context.TODO()
```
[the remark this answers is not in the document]
karlkfi: done.

pkg/apply/cache/resource_caching_poller_test.go
```diff
@@ -0,0 +1,206 @@
+// Copyright 2019 The Kubernetes Authors.
```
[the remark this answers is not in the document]
karlkfi: done

pkg/apply/cache/resource_caching_poller.go
```diff
@@ -0,0 +1,57 @@
+// Copyright 2021 The Kubernetes Authors.
```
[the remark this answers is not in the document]
karlkfi: done

pkg/apply/cache/resource_cache_map.go
```diff
@@ -0,0 +1,104 @@
+// Copyright 2020 The Kubernetes Authors.
```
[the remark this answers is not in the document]
karlkfi: I refactored this to not use a mutex and only be used by the mutator. Is that ok?

pkg/kstatus/polling/testutil/fake_poller.go
karlkfi: done

pkg/object/graph/graph.go
karlkfi: removed. I thought this would be a nice clean way to more efficiently compare edges using sorting, but it was just a nice to have. I've removed it to reduce the size of this PR.

pkg/kstatus/polling/engine/engine_test.go
karlkfi: reverted

pkg/kstatus/polling/statusreaders/testing.go
karlkfi: this should be outdated... the commit is gone. not sure what happened to github that it thinks this is current...

karlkfi:
I've gone back and reverted a bunch of the minor refactors that are in other PRs or aren't critical to the functionality.

karlkfi:
Alright, the general consensus talking to stakeholders seems to be that everyone expects JSONPath, even if no one can really justify why JSONPath is better than yq.

So following the principal of least surprise, I've refactored this to use https://github.com/spyzhov/ajson as the expression language, because that's the only existing Go JSONPath lib that supports both retrieval and mutation.

The one significant downside of this approach is that ajson does not auto-generate intermediate arrays/maps when setting a value. So the value needs to exist in the input object before it can be mutated. I've filed a feature request for that: https://github.com/spyzhov/ajson/issues/35

karlkfi:
One benefit of using ajson over yq is that ajson doesn't use any logging library (except in the main.go, which we aren't using). So this means I can drop the logtuil for translation.

pkg/kstatus/polling/statusreaders/testing.go
karlkfi: fixed.

karlkfi:
This is ready to review again.

The biggest outstanding question, IMO, is the ResourceReference syntax.

1. Should it exactly match ObjMetadata, without the option for apiVersion?
    This might make it hard to set the value with a kpt setter, which doesn't allow for processing the user's input value.
2. Should it use group or apiGroup? There are competing precedents in Kubernetes.

Prior Art:
*   [Cluster][RoleBinding roleRef](https://kubernetes.io/docs/reference/kubernetes-api/authorization-resources/role-binding-v1/):
    *   apiGroup + kind + name
*   [HorizontalPodAutoscaler scaleTargetRef](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/horizontal-pod-autoscaler-v1/#HorizontalPodAutoscalerSpec)
    *   apiVersion + kind + name
*   [PersistentVolume claimRef](https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/persistent-volume-v1/#PersistentVolumeSpec) ([ObjectReference](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-reference/#ObjectReference))
    *   apiVersion + kind + name [+ namespace] [+ resourceVersion] [+ uid] [+ fieldPath]
*   [PersistentVolumeClaim dataSourceRef](https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/persistent-volume-claim-v1/#alpha-level) ([TypedLocalObjectReference](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/typed-local-object-reference/#TypedLocalObjectReference))
    *   apiGroup + kind + name
*   ObjectMeta [OwnerReference](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L307)
    *   apiVersion + kind + name + uid [+ controller] [+ blockOwnerDeletion]
    *   [SIG-Architecture API Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#object-references)
*   [Config Connector resource references](https://cloud.google.com/config-connector/docs/how-to/creating-resource-references)
    *   apiVersion + kind + (external or (name [+ namespace]))
    *   Design: go/kcc-resource-references
*   [cli-utils ObjMetadata](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/object/objmetadata.go#L57)
    *   group + kind + name + namespace

pkg/object/mutation/types.go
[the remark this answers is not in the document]
mortent: I agree that keeping the ObjMetadata as an internal type and define a new one here makes sense. I'm a bit skeptical about supporting apiVersion here, as we are otherwise trying to avoid using the version. If I specify a version here, would we then also require that we fetch the resource from the apiserver with that exact version?

I think we have used just `Group` pretty consistently rather than `apiGroup`.

mortent (commented):
This looks good to me. I'll give others a chance to review it too.

I think it is ok to use the simple cache for now, but we should eventually think about how we can combine leverage the resources we already fetch from the cluster as part of determining reconcile status. But we might want to look at ways to improve that in other ways too.

pkg/apply/cache/resource_cache_map.go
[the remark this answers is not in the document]
seans3: Yes

pkg/object/mutation/types.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	if r.Namespace != "" {
+		return fmt.Sprintf("%s/namespaces/%s/%s/%s", group, r.Namespace, r.Kind, r.Name)
+	}
+	return fmt.Sprintf("%s/%s/%s", group, r.Kind, r.Name)
+}
+
+// Equal returns true if the ResourceReference sets are eqivelent, ignoring version.
+// Fulfills Equal interface from github.com/google/go-cmp
+func (r ResourceReference) Equal(b ResourceReference) bool {
```
seans3, line 140: `Equal` will mostly work when ignoring version. I think this will probably work for what it is being used for. The only true way to compare resources is by comparing their API Server assigned `UID`. For example, the following corner case will incorrectly say the two resources are not the same when they are:

Resource 1: 
Group/Version: `apiextensions.k8s.io/v1beta1`
Kind: `Deployment`
Namespace: `foo`,
Name: `bar`

Resource 2 (same resource):
Group/Version: `apps/v1`
Kind: `Deployment`
Namespace: `foo`,
Name: `bar`

Most of these corner cases have been deprecated by version 1.16 and later.

pkg/object/mutation/types.go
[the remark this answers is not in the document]
seans3: I still think we should stick with `ObjMetadata`, but I will not block on this. Both of these types should not be part of the API, and they should not be exported. The API is the annotation--not these two reference types.

seans3 (commented):
I think this looks good and it is ready to merge. I disagree on the two types which are basically the same thing (`ObjMetadata` and `ResourceReference`), but I will not block the PR on that. The only thing missing appears to be an end-to-end test.

As far as the version in the `ResourceReference`, I think it will not break anything as long as there is only one `ResourceReference` per mutation. But if we ever have to identify if two `ResourceReference` point to the same resource, then we must not use version (the `Equal` implements this correctly). I also think the versioned fetch may be a more fragile approach, since fetching with a version has a higher likelihood of returning an error. A version would only be **required** if the referenced field in the resource is dependent on the version (i.e. the field only exists in a particular version of the resource).

pkg/object/mutation/types.go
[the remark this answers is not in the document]
karlkfi: It's useful to have the structs exported for clients (like kpt functions) to use when building the annotation yaml.

go.sum
karlkfi: outdated

pkg/kyq/kyq.go
karlkfi: switched to jsonpath

pkg/logutil/logutil.go
karlkfi: The ajson library only logs in the main file, which we're not using. So no log translation is needed now. Removed.

pkg/object/mutation/types.go
karlkfi: The only two uses of Equal right now are for detecting self-references and test comparisons. In these cases, we don't really care about the api version.

But you're right that group mismatch would break the self-reference detection if the group changed between versions. This is an interesting edge case. I suspect that it would cause the resource to be treated like two separate resources, with one depending on the other, and both being added to the dependency graph. The graph doesn't look like it checks for duplicates. I'm not even really sure how we would do client-side deduplication when the group differs like this.

pkg/apply/mutator/apply_time_mutator.go
```diff
@@ -0,0 +1,236 @@
...
+	// get resource from cache
+	if atm.ResourceCache != nil {
+		obj, found := atm.ResourceCache.Get(id)
+		if found && obj != nil {
+			return obj, nil
+		}
+	}
+
+	// lookup resource using group api version, if specified
+	sourceGvk := ref.GroupVersionKind()
```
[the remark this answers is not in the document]
karlkfi: > adding the version makes the field substitution more fragile

Yes. This is intentional. I want to allow apiVersion without parsing, but you're right that it's more fragile. If users run into problems, they have the option to switch to using group instead.

> Let's go with this for now--I'm sure we'll run into subtle corner cases as we run it.

If this becomes a problem users complain about, I'm ok with re-evaluating later and removing apiVersion, or accepting apiVersion and just discarding the version (which would be less disruptive, but also less intuitive).

pkg/object/mutation/types.go
```diff
@@ -0,0 +1,151 @@
...
+	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+// ApplyTimeMutation is a list of substitutions to perform in the target
+// resource before applying, after waiting for the source resources to be
+// reconciled.
+// This most notibly allows status fields to be substituted into spec fields.
+type ApplyTimeMutation []FieldSubstitution
+
+// Equal returns true if the substitutions are eqivelent, ignoring order.
```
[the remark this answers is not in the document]
karlkfi: fixed

karlkfi:
Alright, I added new e2e tests. I don't have any interesting use cases that use the default K8s resources, so it's a bit of a toy case. But I wasn't sure if it would be a good idea to copy in the [Config Connector CRDs](https://github.com/GoogleCloudPlatform/k8s-config-connector/).

I also made the jsonpath lib a little more useful by adding support for multi-get and multi-set, but I kept the apply-time-mutation requiring exactly one match. We can add fan-out and/or fan-in later if users ask for it.

karlkfi:
Do I need to swash the commits myself? Probably don't want all the revert history in there.

seans3:
> Do I need to swash the commits myself? Probably don't want all the revert history in there.

I usually squash them myself, although I think there is a merge option to squash them too. We do not want the merge history. Almost always we merge with one commit.

karlkfi:
squashed

seans3 (approved):
/lgtm

PR #401, opened 2021-09-07
Ignore prune object "not found" error

seans3:
* Fixes bug where apply fatally errors if a prune object is not in cluster. Skips prune objects which are already deleted from cluster (as determined by "Not Found" error).
* Adds end-to-end test to validate the fix.

### DIFF
--- a/pkg/apply/prune/prune.go
+++ b/pkg/apply/prune/prune.go
@@ -15,6 +15,7 @@ import (
 	"context"
 	"sort"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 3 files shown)


examples/alphaTestExamples/pruneRetrieveError.md
```diff
@@ -0,0 +1,201 @@
+[kind]: https://github.com/kubernetes-sigs/kind
```
mortent, line 1: I don't think we should keep adding tests using this framework. The mdrip framework is useful for providing executable examples, but I don't think these are being used for that.

While the Ginkgo tests aren't perfect, they are definitely better than using the mdrip framework.

pkg/apply/prune/prune.go
```diff
@@ -164,8 +164,12 @@ func (po *PruneOptions) GetPruneObjs(inv inventory.InventoryInfo,
 	pruneObjs := []*unstructured.Unstructured{}
 	for _, pruneID := range pruneIds {
 		pruneObj, err := po.GetObject(pruneID)
+		// If we can not retrieve the object--do not attempt to prune it.
+		// TODO(seans): distinguish between "not found" and other errors.
 		if err != nil {
-			return nil, err
+			klog.V(4).Infof("error (%s) retrieving prune obj (%s/%s)--skip",
```
mortent, line 170: Why are we punting on distinguishing between NotFound an other errors?

seans3: Resolved by skipping not found object, and return fatal error for other errors as discussed offline.

mortent (approved):
/lgtm

PR #402, opened 2021-09-13
Refactor depends-on code

karlkfi:
- Move from `pkg/object` to `pkg/object/dependson`
- Add DependencySet type to abstract complexity specific to the depends-on annotation (set equality, format/parse)
- Add Parse/Format funcs for ObjMetadata and DependencySet
- Refactor primary API to ReadAnnotation and WriteAnnotation
- Move Write/Format behavior from testutil into dependson pkg
- Make constants private to encourage using functions as primary API

The following optional refactors may make good followup PRs:
- Refactor non-annotation uses of []ObjMetadata to use DependencySet, and replace object.SetEquals with DependencySet.Equal
- Replace object.ObjMetas (used only twice for Contains method) with DependencySet
- Refactor logging to use MarshalObjMetadata and MarshalDependencySet with a `String()` wrapper, to replace the `%s/%s` used in many places.
- Rename ObjMetadata to ResourceReference
- Move ObjMetadata to `pkg/object/reference` or ResourceReference to `pkg/object/reference` to reduce package complexity and simplify naming of functions


seans3 (approved):
/lgtm

PR #403, opened 2021-09-13
[DEMO] Updates TaskContext to store entire object

seans3:
/hold

* Example-only code to demonstrate `TaskContext` as an object cache
* Updates `applyInfo` to `*unstructured.Unstructured` within `TaskContext`

Example Output

```
$ kapply apply ~/go/src/github.com/GoogleContainerTools/kpt/e2e/live/testdata/depends-on-ordering/
namespace/depends-on-namespace unchanged
pod/pod-c unchanged
I0912 22:44:00.425225  465283 apply_task.go:182] mutate depends-on-namespace/pod-a field here: source depends-on object: default/pod-c
pod/pod-a unchanged
I0912 22:44:00.428305  465283 apply_task.go:182] mutate depends-on-namespace/pod-b field here: source depends-on object: depends-on-namespace/pod-a
pod/pod-b unchanged
4 resource(s) applied. 0 created, 4 unchanged, 0 configured, 0 failed
```

/assign @karlkfi

### DIFF
--- a/pkg/apply/task/apply_task.go
+++ b/pkg/apply/task/apply_task.go
@@ -4,6 +4,7 @@
 package task
 
 import (
+	"fmt"
 	"io/ioutil"
 	"strings"
 
@@ -23,6 +24,7 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/apply/taskrunner"
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/cli-utils/pkg/object"
+	"sigs.k8s.io/cli-utils/pkg/object/dependson"
 )
 
 // applyOptions defines the two key functions on the ApplyOptions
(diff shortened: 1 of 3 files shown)


karlkfi:
After in-person discussion, I don't think this is the direction we should take.

1. If the mutation contract is that it wants the latest value after reconciliation, then there's no need to cache the values at apply time. Without using the poller results, there's no way to invalidate that cache. So we would need to make a GET every time anyway.
2. If the desire is to not cache the result from the poller, because of fear of multi-threaded complexity, then we need to do a GET every time anyway, or make a cache that lives entirely in the mutator (no multi-threading).

I will update https://github.com/kubernetes-sigs/cli-utils/pull/400 to remove the ResourceCache from the task context for now and GET the resource in the same thread instead. I may make a follow-up PR to add a single-threaded cache to the mutator, tho. Because I anticipate that having multiple substitutions with the same source resource will be fairly common both in the same target resource, and in the apply set. 

A common example is creating a VPC Project, a platform Project and a GKE cluster in the same apply set. When used with a Shared VPC, GKE needs 3 different IAMPolicyMembers for each cluster using Google-managed service accounts (with project-number in their name). So even for a single cluster, the Project is used as a source for at least 3 different mutations. Expand that out to 20 or 200 clusters and having a cache makes clear sense.

PR #405, opened 2021-09-15
Add more robust event list testing in e2e tests

karlkfi:
- Add testutil.Equal to show comparisons with cmp.Diff output
- Add github.com/google/go-cmp for cmp.Equal and cmp.Diff
  These comparison tools make debugging easier and properly handle
  error comparisons, using cmpopts.EquateErrors()
- Added EventsToExpEvents, EventToExpEvent, and RemoveEqualEvents
  to testutil to help with comparing event lists
- Add InitEvent to ExpEvent, to enable validating it exists
- e2e tests have all been upgraded to validate a full list of
  events, rather than just a partial list with partial contents.
  This should increase confidence that new changes don't break things.
- Add test cleanup in AfterSuite and AfterEach to allow re-using a kind
  cluster between multiple test runs without test failures.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module sigs.k8s.io/cli-utils
 go 1.16
 
 require (
+	github.com/google/go-cmp v0.5.6
 	github.com/google/uuid v1.2.0
 	github.com/onsi/ginkgo v1.16.2
 	github.com/onsi/gomega v1.12.0
--- a/go.sum
+++ b/go.sum
@@ -268,8 +268,9 @@ github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/
 github.com/google/go-cmp v0.5.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.2/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.4/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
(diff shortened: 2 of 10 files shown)


karlkfi:
I've been getting flaky results from the "Apply and destroy" e2e tests locally. I'm not sure if it's a bug or not, but it's new to the expanded event list testing.

The `DeleteEvent` in `apply_and_destroy_test.go` seems to oscillate between `Deleted` and `DeleteSkipped`.

Also, the resulting event list from the destroyer in the same test doesn't have any WaitTasks after the delete, which is confusing, because I thought it always waits for deletion.

@seans3 or @mortent perhaps you can speak to what the desired behavior is? The existing test code didn't specify.

karlkfi:
Example diff output from the  new Equal matcher when there's a test error with the event lists:

```
      Diff:
            []testutil.ExpEvent{
            	{InitEvent: &{}},
            	{EventType: s"ActionGroupType", ActionGroupEvent: &{Name: "prune-0", Action: s"DeleteAction"}},
            	{
            		... // 4 identical fields
            		StatusEvent: nil,
            		PruneEvent:  nil,
            		DeleteEvent: &testutil.ExpDeleteEvent{
          - 			Operation:  s"DeleteSkipped",
          + 			Operation:  s"Deleted",
            			Identifier: {Namespace: "e2e-test-78635689", Name: "nginx-deployment", GroupKind: {Group: "apps", Kind: "Deployment"}},
            			Error:      nil,
            		},
            	},
            	{EventType: s"ActionGroupType", ActionGroupEvent: &{Name: "prune-0", Action: s"DeleteAction", Type: s"Finished"}},
            	{EventType: s"ActionGroupType", ActionGroupEvent: &{Name: "delete-inventory-0", Action: s"InventoryAction"}},
            	{EventType: s"ActionGroupType", ActionGroupEvent: &{Name: "delete-inventory-0", Action: s"InventoryAction", Type: s"Finished"}},
            }

```

karlkfi:
I fixed the DeleteSkipped test bug. Thanks @mortent for tracking down the cause!

PR #407, opened 2021-09-23
Add a new flag value `force-adopt` to the `--inventory-policy` flag

haiyanmeng:
`force-adopt` allows the current inventory take ownership of any objects.


Liujingfang1:
What is the motivation to expose the AdoptAll policy to the cli. I thought we only expect the library consumer to use AdoptAll.

haiyanmeng:
> What is the motivation to expose the AdoptAll policy to the cli. I thought we only expect the library consumer to use AdoptAll.

Quick testing.

seans3 (approved):
This looks good. I'll let others chime in, then we can merge.

seans3 (approved):
/lgtm

PR #408, opened 2021-09-23
e2e test cleanup

karlkfi:
- Unstructured() added to ResourceReference to aid test operations
- Test artifacts moved into artifacts_test.go
- deploymentManifest and apiserviceManifest replaced with
  yaml->unstructured, for consistency with other test artifacts
- Added withReplicas, withNamespace, and withDependsOn for easier inline
  mutation of test artifacts
- Added deleteUnstructuredAndWait, assertUnstructuredExists, and
  assertUnstructuredDoesNotExist to validate actual cluster state,
  not just events. Also reduced flakiness by waiting for delete.
- Removed error check before event validation. The event diff
  provides much more context for debugging.
- Replaced typed resources with unstructured and used
  testutil.NestedField to extract values.
- Updated prune error test to use full event list comparison and
  validate server state.
- Add 2nd resource to continue on error test
- Don't fail early on error from event list
- Fix flakey prune error test
  Add wait for pod2 after creation
  Add wait for pod2 after deletion
  Re-use inventory to ensure prune of pod2 not skipped
- Add optional klog support for e2e tests

### DIFF
--- a/test/e2e/continue_on_error_test.go
+++ b/test/e2e/continue_on_error_test.go
@@ -6,7 +6,6 @@ package e2e
 import (
 	"context"
 	"errors"
-	"strings"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -19,7 +18,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func continueOnErrorTest(_ client.Client, invConfig InventoryConfig, inventoryName, namespaceName string) {
+func continueOnErrorTest(c client.Client, invConfig InventoryConfig, inventoryName, namespaceName string) {
 	By("apply an invalid CRD")
 	applier := invConfig.ApplierFactoryFunc()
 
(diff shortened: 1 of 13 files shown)


seans3 (commented):
This is a good cleanup--thanks. Let me know after rebase when you're ready to merge.

karlkfi:
After talking to @haiyanmeng I'm convinced that ConfigSync and kpt would both be better if the WaitTask always happens after apply/prune/delete. But I don't want to make that behavioral change in this testing PR. So I'll add some waits into the tests for now and make another PR for the behavioral change.

karlkfi:
I figured out the bugs with the prune_retrieve_error_test:
1. apply wasn't waiting for pod2 to be created (flakey)
2. destroy wasn't waiting for pod2 to be deleted (flakey)
3. the destroyer was using a new inventory with a different random ID, so the delete task was running but the delete was being skipped/filtered for non-matching inventory ID

I've fixed all 3 issues in the tests.

karlkfi:
squashed. ready to merge once tests pass.

test/e2e/continue_on_error_test.go
```diff
@@ -90,6 +98,25 @@ func continueOnErrorTest(_ client.Client, invConfig InventoryConfig, inventoryNa
 			},
 		},
 		// Note: No WaitTask when apply fails
+		// TODO: why no wait after create tho?
```
seans3, line 101: Should this commented out stuff be removed?

karlkfi: I'm still not exactly sure why the WaitTask isn't being triggered.
Is it not waiting because it's the last apply in the task list?
Is it not waiting because there's no timeout configured?
Why does it wait sometimes and not others?

I started a chat convo about this last week and the consensus from Config Sync participants seemed to indicate that we should change it to always wait. I think that makes the most sense from a user perspective and is both intuitive and less surprising. But I didn't want to make that change in this commit. So I just left the WaitTasks commented out (as I had in other tests).

Do you want me to rip out the comment and just add it in the next PR that enables waiting?

seans3 (approved):
/lgtm

PR #410, opened 2021-09-25
Improve clarity of filter and mutator reasons

karlkfi:
- Use stringer for all enums (to translate ids into names)
- Remove redundant effect from reason strings
- Use consistent pattern for variables in reason strings.
  We can't use structured json, but this is close, while still human
  readable.
- Remove copyright on generated files

/cc @seans3 @mortent


mortent (commented):
/lgtm
I'll let @seans take a look and approve.

seans3 (approved):
/lgtm

PR #411, opened 2021-09-27
Fixes prune timeout value

seans3:
* Fixes bug where `PruneTimeout` does not add `WaitTask` to wait for deletion. It incorrectly set the `WaitTask` when the `ReconcileTimeout` was set using this value as the timeout value.
* Adds unit tests.


karlkfi:
What's the diff between PruneTimeout and ReconcileTimeout?
Is ReconcileTimeout supposed to be for create/update?
If so, aren't they both waiting for the desired state to become the current state? Should they be combined?
Should this change affect e2e tests?

karlkfi:
Per discussion offline:
- PruneTimeout is for waiting after any delete (applier or destroyer)
- ReconcileTimeout is for waiting after any create or update (applier)

PR #413, opened 2021-09-29
Merge Collector and ResourceCache

karlkfi:
- Make ResourceCache thread-safe
- Make ResourceCache store status and messagei
- Add ResourceCache to baseRunner and TaskContext
- Make Mutator compute resource status for uncached resources
- Share cache between StatusPoller and Mutator
- Move Condition and conditionMet() to its own file
- Simplify WaitTask.checkCondition
- Simplify baseRunner.amendTimeoutError


pkg/apply/cache/resource_cache.go
```diff
@@ -5,20 +5,39 @@ package cache
...
+}
+
+// ResourceCache stores CachedResource objects
 type ResourceCache interface {
+	ResourceCacheReader
+	// Load the resources into the cache, generating the ObjMetadata from the objects.
+	Load(...ResourceStatus) error
 	// Put the resource into the cache, generating the ObjMetadata from the object.
-	Put(obj *unstructured.Unstructured) error
+	Put(obj ResourceStatus) error
```
mortent, line 29: What is the difference between `Load`, `Put`, and `Set`? They seem very similar.

pkg/apply/cache/resource_cache_map.go
```diff
@@ -50,13 +56,19 @@ func (rc *ResourceCacheMap) Put(value *unstructured.Unstructured) error {
 
 // Set the resource in the cache using the supplied key, and replacing any
 // existing resource with the same key.
-func (rc *ResourceCacheMap) Set(key object.ObjMetadata, value *unstructured.Unstructured) {
+func (rc *ResourceCacheMap) Set(key object.ObjMetadata, value ResourceStatus) {
```
mortent, line 59: Do we need this function, or can we compute the key based on the Unstructured in the ResourceStatus?

pkg/apply/taskrunner/runner.go
```diff
@@ -71,10 +71,10 @@ func (tsr *taskStatusRunner) Run(ctx context.Context, taskQueue chan Task,
 
 // NewTaskRunner returns a new taskRunner. It can process taskqueues
 // that does not contain any wait tasks.
+// TODO: Do we need this abstraction layer now that baseRunner doesn't need a collector?
```
mortent, line 74: Probably not. I'm ok with simplifying things here.

pkg/apply/taskrunner/task.go
```diff
@@ -208,34 +198,6 @@ type resourceWaitData struct {
 	generation int64
 }
```
mortent, line 199: I think the `resourceWaitData` struct is no longer in use and can be deleted.

mortent (commented):
This looks good. It makes the handling of tasks more efficient and easier to understand.

pkg/apply/cache/resource_cache.go
karlkfi: Load is just a wrapper for Put to add multiple atomically. Mostly just used for tests to avoid a for loop.
Put wraps Set and handles generating the ObjMetadata from the Unstructured.
Set accepts a user specified ObjMetadata as the key, since we mostly have it already when adding objects to the cache.

pkg/apply/cache/resource_cache_map.go
karlkfi: That's what Put does. But in most cases we already have the ObjMetadata computed.

pkg/apply/taskrunner/runner.go
karlkfi: I think I'll leave that for a follow up PR. This one's already big enough.

pkg/apply/taskrunner/task.go
karlkfi: done.

pkg/apply/cache/resource_cache.go
karlkfi: I've removed Put in favor of just using Load, and then renamed Set to Put to match normal map operator naming.

karlkfi:
Pretty sure this is a flaky error related to apply ordering not actually being consistent. The apply sets are sorted, but within each set the set itself isn't sorted.

/test cli-utils-presubmit-master-e2e

mortent (approved):
/lgtm
/approve

PR #417, opened 2021-10-05
Skip pruning an object whose Group/Kind is not found

haiyanmeng:


mortent (commented):
Change looks good, but can we add a test?

pkg/apply/prune/prune.go
```diff
@@ -166,14 +166,17 @@ func (po *PruneOptions) GetPruneObjs(inv inventory.InventoryInfo,
 	for _, pruneID := range pruneIds {
 		pruneObj, err := po.GetObject(pruneID)
 		if err != nil {
-			// If prune object is not in cluster, no need to prune it--skip.
-			if apierrors.IsNotFound(err) {
-				klog.V(4).Infof("prune obj not in cluster--skip (%s/%s)",
+			if meta.IsNoMatchError(err) {
+				klog.V(4).Infof("skip pruning obj %s/%s: no match for the kind %q in Group %q",
```
karlkfi, line 170: "no match" sounds ambiguous. 
How about `skip pruning obj %s/%s: the resource type is unrecognized by the cluster (kind: %q, group %q)`.

karlkfi:
Do we need to support this for Apply too? 
Or is failure in that case the desirable behavior?

seans3 (commented):
This looks good. Can we add a unit test that triggers this new error ?

pkg/apply/prune/prune.go
haiyanmeng: Done

seans3 (commented):
/lgtm

I'll approve and merge this shortly if no one has any further concerns

PR #430, opened 2021-10-12
Check versions a CRD defines

ash2k:
If a CRD does not define a version that a CR has, then it's an error and CR scope cannot be decided.

### DIFF
--- a/pkg/manifestreader/common.go
+++ b/pkg/manifestreader/common.go
@@ -21,12 +21,12 @@ import (
 
 // UnknownTypesError captures information about unknown types encountered.
 type UnknownTypesError struct {
-	GroupKinds []schema.GroupKind
+	GroupVersionKinds []schema.GroupVersionKind
 }
 
 func (e *UnknownTypesError) Error() string {
 	var gks []string
-	for _, gk := range e.GroupKinds {
+	for _, gk := range e.GroupVersionKinds {
 		gks = append(gks, gk.String())
 	}
 	return fmt.Sprintf("unknown resource types: %s", strings.Join(gks, ","))
@@ -62,7 +62,7 @@ func SetNamespaces(mapper meta.RESTMapper, objs []*unstructured.Unstructured,
 		}
 	}
 
-	var unknownGKs []schema.GroupKind
+	var unknownGVKs []schema.GroupVersionKind
 	for _, obj := range objs {
 		// Exclude any inventory objects here since we don't want to change
 		// their namespace.
@@ -78,7 +78,7 @@ func SetNamespaces(mapper meta.RESTMapper, objs []*unstructured.Unstructured,
 			if errors.As(err, &unknownTypeError) {
 				// If no scope was found, just add the resource type to the list
 				// of unknown types.
-				unknownGKs = append(unknownGKs, unknownTypeError.GroupKind)
+				unknownGVKs = append(unknownGVKs, unknownTypeError.GroupVersionKind)
 				continue
 			} else {
 				// If something went wrong when looking up the scope, just
@@ -108,9 +108,9 @@ func SetNamespaces(mapper meta.RESTMapper, objs []*unstructured.Unstructured,
 			return fmt.Errorf("unknown RESTScope %q", scope.Name())
 		}
 	}
-	if len(unknownGKs) > 0 {
+	if len(unknownGVKs) > 0 {
 		return &UnknownTypesError{
-			GroupKinds: unknownGKs,
+			GroupVersionKinds: unknownGVKs,
 		}
 	}
 	return nil
--- a/pkg/manifestreader/common_test.go
+++ b/pkg/manifestreader/common_test.go
@@ -22,7 +22,7 @@ import (
 func TestSetNamespaces(t *testing.T) {
 	testCases := map[string]struct {
 		objs             []*unstructured.Unstructured
-		defaultNamspace  string
+		defaultNamespace string
 		enforceNamespace bool
 
 		expectedNamespaces []string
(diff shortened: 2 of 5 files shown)


ash2k:
I think whatever we do should be consistent with how rest mappers work. How they work is they try to find an exact match for the group+kind and one of the versions that it's asked about - https://github.com/kubernetes/kubernetes/blob/e054181e517b48a3c862207537092c28604aaad9/staging/src/k8s.io/apimachinery/pkg/api/meta/interfaces.go#L127.

In our case, we are using the rest mapper [here](https://github.com/kubernetes-sigs/cli-utils/pull/430/files#diff-e6111d1a995fec892e4ffc06d21f013a2ca3e96f8a8e77ea4a9496dd3266b2d0R135), but we are not providing a version. That means it uses a preferred version. I think this needs to be updated (I missed it) to provide the version from the actual object (will do that now).

The idea is that if a CRD defines v1 and v1beta1 (or any set of versions, I used v1 and v2 in test), CR must use one of them. It cannot use an undefined version because such an object cannot be put into a cluster. So it makes sense to reject such object as early as possible to prevent a partial apply.

Does it make sense?

ash2k:
This works the same way for builtin type and for CRD+CR as the API rules are the same for all objects.

ash2k:
@seans3 Hm, but do you agree with this:

> If a CRD does not define a version that a CR has, then it's an error and CR scope cannot be decided.

And the second commit does the same for builtin resources - when querying the mapper we pass the version of the object that we have. If such version is not defined then the server does not know about it and hence conversion/apply/etc will fail so there is no point in proceeding. It's better to fail earlier than in the middle of an apple/etc.

WDYT?

ash2k:
Put another way: `apps/v10`/`Deployment` is not a valid resource and apiserver will reject it. Why not reject it sooner? Same for CRD+CR.

Put yet another way: can we write a test for a desired behavior that will be broken by this PR?

seans3:
> Put another way: `apps/v10`/`Deployment` is not a valid resource and apiserver will reject it. Why not reject it sooner? Same for CRD+CR.
> 
> Put yet another way: can we write a test for a desired behavior that will be broken by this PR?

These are good questions. I will add these explanations to the developer documentation.

The short answer to your question is we do not want to be in the business of validating on the client. Client-side validation is extra complexity that could possibly cause version skew problems. The server will always validate better than the client (e.g. `ValidatingWebhooks` only exist on the server)--so we'll simply defer validation to when we apply. The optimization of one less network call is not worth the extra complexity and version skew risk.

**Question**: What about invalid object metadata in the inventory?  **Answer**: The inventory algorithm is tolerant to invalid metadata. Even though this invalid metadata might be inserted into the inventory initially, it will be cleaned up with the final inventory task which only add inventory metadata to objects which successfully applied (or prune failed). We still need to make sure the final inventory task always gets run, because wait timeouts do not run this final inventory task yet.

**Question**: What about client-side validation that already exists? **Answer**: We will remove it when we've ensured the final inventory task is always run.

**Somewhat related topic**: Why do we only have four fields (`Group/Kind/Namespace/Name`) for object metadata, not including version. **Answer**: These four fields uniquely describe one and only one object stored in the API server (Kind ~ Resource; see GVK versus GVR). When we add the version, we can end up with multiple entries in the inventory describing the same object stored in the API Server. And these duplicates in the inventory can cause the inventory algorithm to accidentally prune (delete) objects.

seans3:
> Thanks for clarifying. I agree re. client vs server validation. However, I think what I'm suggesting is a very basic thing, perhaps similar to valid/invalid YAML validation.
> 
> If we don't provide the version to the rest mapper, it just uses the server-preferred version of the resource rather than the one the user provided (which may be invalid). I don't see how asking rest mapper a more precise question (GK+any version vs GVK from user input) can cause any problems? Could you clarify what version skew you are talking about? I'm not suggesting to use GVK instead of GK in other places.
> 
> p.s. another thing to check is that we are using a correctly configured client. I suspect we may use client with server-preferred version and not with user-specified version. I.e. we need to ensure we are sending e.g. `apps/v1beta1/Deployment` to `apps/v1beta1` API endpoint and not to `apps/v1` API endpoint if that version is preferred by the server.

You are correct that we can ask a more precise question, and it will be a stronger validation. As mentioned before, though, I don't think we want to get in the validation business. I believe here we are doing validation reluctantly in order to get the `RESTScope` (i.e. namespaced, cluster scoped); not for validation itself. I will leave this final resolution up to you. But if we do add this extra validation, we'll need a very visible comment (something like `NOTE:`) that the version validation does NOT apply to inventory items.

Version skew is compiling in version-specific code, which can differ from the version understood by the server. This is not an issue with the `RESTMapper` (since it is created from server discovery), and it is also usually not a problem in this library because we use `Unstructured` objects which are just byte blobs instead of the concrete golang struct types. We need to make sure any validation does not compile in specific types, which should be a low bar. 

Finally, as far as the client, the server-preferred version is what we want (at least it is the easiest). The different API endpoints will **ALWAYS** point to the exact same object. The `apps/v1beta1/Deployment` API endpoint points to the exact same object as the `apps/v1/Deployment` version of the object. The API Server will do internal conversion on the storage version of the object to return the requested version of the object. This is what I mean when I say the the `Group/Kind/Namespace/Name` defines one and exactly one object in the API server.

seans3 (approved):
/lgtm

ash2k:
Thank you, @seans3!

karlkfi:
Looks like this introduced a minor regression in terms of the error string output by `manifestreader.UnknownTypesError`.

By changing from GroupKinds to GroupVersionKinds, the String() result changed from `Custom.custom.io` to `custom.io/v1, Kind=Custom`.

While adding the version to the error is probably fine, it should probably be changed to `Custom.custom.io/v1`, given that multiple are delimited with a comma.

OTOH, `Custom.custom.io` is technically an invalid resource name, because it uses the Kind instead of the (usually plural) Resource name...

I'm open to other alternative string formats, but `unknown resource types: custom1.io/v1, Kind=Custom1, custom2.io/v1, Kind=Custom2` is pretty terrible.

karlkfi:
Fix: https://github.com/kubernetes-sigs/cli-utils/pull/479

PR #435, opened 2021-10-14
Update to latest version of kyaml and k8s libs

mortent:
The latest version of kyaml has changes to the API, which requires that we also update the k8s libraries since kubectl has a dependency on the kustomize api package.


mortent:
/hold
Make sure everyone gets a chance to weigh in on this change.

Issue #433, opened 2021-10-12
Add global timeout to kapply cmds that use applier/destroyer and update example tests


karlkfi:
/assign @karlkfi

PR #439, opened 2021-10-16
Add timeout option & error on timeout/cancel

karlkfi:
- Add timeout flag to apply, destroy, & status cmds (for testing)
- Return error & send error event on cancel/timeout (for improved UX when context is cancelled)

With this change, callers of `Applier.Run` and `Destroyer.Run` will receive both a returned error and an error event when the context is cancelled. These errors can be tested against `context.Canceled` and `context.DeadlineExceeded` to see if the error was cancellation or timeout.

The error event allows the Printer processing the events to print the error to the user.

These errors print the following by default:

```
// cancellation
error: context canceled

// timeout
error: context deadline exceeded
```

Fixes: https://github.com/kubernetes-sigs/cli-utils/issues/433

### DIFF
--- a/cmd/apply/cmdapply.go
+++ b/cmd/apply/cmdapply.go
@@ -91,11 +91,7 @@ type ApplyRunner struct {
 }
 
 func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
-	// If specified, use command context.
 	ctx := cmd.Context()
-	if ctx == nil {
-		ctx = context.Background()
-	}
 	// If specified, cancel with timeout.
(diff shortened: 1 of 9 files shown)


karlkfi:
Wrote a TODO for cancellation of tasks during execution: https://github.com/kubernetes-sigs/cli-utils/issues/440

ash2k:
@karlkfi Thanks for a heads up! I'm fine with the behavior change.

mortent:
This looks good to me. Once we have addressed/responded to the comment from @ash2k I can lgtm.

PR #444, opened 2021-10-19
[WIP] Support apply of Gatekeeper Template and Constraint in the same set

mortent:
cli-utils currently doesn't support applying Gatekeeper templates and constraints in the same set, even with the `depends-on` annotation. This is a POC on how to implement support directly in the library and meant for discussion rather than merging it as is.

There are three challenges here:

**The library needs to be able to determine the namespace of gatekeeper constraints, even when the CRD for the template doesn't exist in the cluster**
This is very similar as for CRs and CRDs, although it is somewhat easier here since we know that all gatekeeper constraints are cluster-scoped. This PR doesn't actually check that we have the template in the same apply set, but just uses the group for gatekeeper constraints to determine the scope. We could make the check more thorough and error out early if we can't find the template in either the apply set or in the cluster.
A related question here is whether the library should verify the namespace field for provided resources or just accept what is passed in by the user.

**The library needs to apply the template before the constrain and make sure the CRD for the template has been generated**
This PR handles this in the solver, similar to how we manage CRDs and namespaces. But ideally I think we would rather let users use the `depends-on` annotation for this rather than us having to code support for this specifically in the library.
This also adds a rule for determining when a Gatekeeper template has reconciled. Adding rules for all possible types to the library obviously doesn't scale, so we should allow users to provide custom rules for this. It has always been the goal of the library, we just have never gotten around to it.

**Refresh the RESTMapper to pick up the type generated from a template**
Since Gatekeeper templates results in a CRD, we need to make sure the new type is seen by the RESTMapper after a template has reconciled. This PR does this the same way as for CRDs, but this is not a good long-term solution. We have discussed letting the library just check for new types on-demand, rather than invalidating the full RESTMapper after specific types has applied and reconciled.

### DIFF
--- a/pkg/apply/taskrunner/task.go
+++ b/pkg/apply/taskrunner/task.go
@@ -8,8 +8,6 @@ import (
 	"reflect"
 	"time"
 
-	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/client-go/restmapper"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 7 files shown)


ash2k:
I think it'd be better to have a way to extend cli-utils to support special objects than to try to build in support for all possible things that exist in the ecosystem :)

PR #445, opened 2021-10-19
Clean up prune package

karlkfi:
- Rename PruneOptions to Pruner (it does the work itself)
- Rename variables and internal methods for clarity
- Extract deleteObject method to improve readability
- Rename GetObject to getObject (only used internally and by test)
- Modify log messages for readability
- Add log support for prune tests

This is mostly just reducing tech debt to improve my sanity and make it easier to tackle harder problems.

### DIFF
--- a/pkg/apply/applier.go
+++ b/pkg/apply/applier.go
@@ -31,7 +31,7 @@ import (
 
 // NewApplier returns a new Applier.
 func NewApplier(factory cmdutil.Factory, invClient inventory.InventoryClient, statusPoller poller.Poller) (*Applier, error) {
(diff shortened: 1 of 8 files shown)


karlkfi:
> you'll see the Options pattern is very common in kubectl.

I haven't dug into kubectl code recently, but IMO Options are structs you pass to a method or constructor to make those functions accept fewer arguments in an extensible way that allows adding options without changing the interface of the method or constructor. Having the Options struct actually implement the methods that do the work seems like a confusing anti-pattern.

karlkfi:
Gotcha. I've now switched to %v for consistency. Thanks for the reasoning.

I also cleaned up a few more private variable/method/argument names for consistency.

PR #451, opened 2021-10-27
Change the behavior of hanlding a WaitTask timeout

haiyanmeng:
Today, when a WaitTask timeout happens, the WaitTask sends the
TimeoutError on the TaskChannel. After receiving the TimeoutError,
`baseRunner.run` terminates immediately by returning the error to its
caller (Applier.Run or Destroyer.Run). The caller then sends the error
onto the EventChannel and terminates.

With this PR, when a WaitTask timeout happens, the WaitTask sends a
WaitType Event including the TimeoutError on the EventChannel, and then
sends an empty TaskResult on the TaskChannel. An empty TaskResult
suggests that the task finished successfully, and therefore
`baseRunner.run` would continue instead of terminate.

The motivation of this change is to make sure that cli-utils only
terminates on fatal errors (such as inventory-related errors, and
ApplyOptions creation errors). A WaitTask timeout may not always mean a
fatal error (it may happen because the StatusPoller has not finished
polling everything, or some but not all the resources have not reached
the desired status), and therefore should not terminate cli-utils.

### DIFF
--- a/pkg/apply/event/event.go
+++ b/pkg/apply/event/event.go
@@ -21,6 +21,7 @@ const (
 	StatusType
 	PruneType
 	DeleteType
+	WaitType
 )
 
 // Event is the type of the objects that will be returned through
@@ -58,6 +59,9 @@ type Event struct {
 	// DeleteEvent contains information about object that have been
 	// deleted.
 	DeleteEvent DeleteEvent
+
+	// WaitEvent contains information about any errors encountered in a WaitTask.
+	WaitEvent WaitEvent
 }
 
 type InitEvent struct {
@@ -83,6 +87,11 @@ type ActionGroup struct {
 
 type ErrorEvent struct {
 	Err error
+}
+
+type WaitEvent struct {
+	GroupName string
+	Error     error
 }
 
 //go:generate stringer -type=ActionGroupEventType
--- a/pkg/apply/taskrunner/runner.go
+++ b/pkg/apply/taskrunner/runner.go
@@ -222,8 +222,9 @@ func (b *baseRunner) run(ctx context.Context, taskQueue chan Task,
 				}
 			}
 		// A message on the taskChannel means that the current task
-		// has either completed or failed. If it has failed, we return
-		// the error. If the abort flag is true, which means something
+		// has either completed or failed.
+		// If it has failed, we return the error.
+		// If the abort flag is true, which means something
 		// else has gone wrong and we are waiting for the current task to
 		// finish, we exit.
 		// If everything is ok, we fetch and start the next task.
@@ -238,7 +239,6 @@ func (b *baseRunner) run(ctx context.Context, taskQueue chan Task,
 				},
 			}
 			if msg.Err != nil {
-				b.amendTimeoutError(taskContext, msg.Err)
 				return msg.Err
 			}
 			if abort {
@@ -259,24 +259,6 @@ func (b *baseRunner) run(ctx context.Context, taskQueue chan Task,
 			abortReason = ctx.Err() // always non-nil when doneCh is closed
 			completeIfWaitTask(currentTask, taskContext)
 		}
-	}
-}
-
-func (b *baseRunner) amendTimeoutError(taskContext *TaskContext, err error) {
-	if timeoutErr, ok := err.(*TimeoutError); ok {
-		var timedOutResources []TimedOutResource
-		for _, id := range timeoutErr.Identifiers {
-			result := taskContext.ResourceCache().Get(id)
-			if timeoutErr.Condition.Meets(result.Status) {
-				continue
-			}
-			timedOutResources = append(timedOutResources, TimedOutResource{
-				Identifier: id,
-				Status:     result.Status,
-				Message:    result.StatusMessage,
-			})
-		}
-		timeoutErr.TimedOutResources = timedOutResources
 	}
 }
 
--- a/pkg/apply/taskrunner/runner_test.go
+++ b/pkg/apply/taskrunner/runner_test.go
@@ -45,9 +45,8 @@ func TestBaseRunner(t *testing.T) {
 		statusEventsDelay         time.Duration
 		statusEvents              []pollevent.Event
 		expectedEventTypes        []event.Type
-		expectedError             error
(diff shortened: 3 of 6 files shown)


pkg/apply/taskrunner/runner.go
```diff
@@ -236,7 +243,15 @@ func (b *baseRunner) run(ctx context.Context, taskQueue chan Task,
...
 				b.amendTimeoutError(taskContext, msg.Err)
-				return msg.Err
+				// If `o.continueOnError` is `false`, we return `msg.Err` immediately, and the caller of the function
+				// would send a ErrorType Event including `msg.Err` to eventChannel.
+				// If `o.continueOnError` is `true`, we don't not return immediately, but
+				// instead log the error and send a WaitType Event including `msg.Err` to eventChannel before continuing.
+				if !o.continueOnError {
+					return msg.Err
+				}
+				klog.Infof("ignore the TimeoutError: %v", msg.Err)
```
karlkfi, line 253: This log message is inaccurate. Errors from the task channel are not always wait task timeouts. Suggest: `continuing on error (task: %q): %v` with the `currentTask.Name()`.

pkg/apply/taskrunner/runner.go
```diff
@@ -258,6 +273,15 @@ func (b *baseRunner) run(ctx context.Context, taskQueue chan Task,
 	}
 }
 
+func handleWaitError(eventChannel chan event.Event, err error) {
```
karlkfi, line 276: Why are we sending a new different event when continue on error is enabled?
If continue on error is disabled this would be an ErrorEvent, right?

pkg/apply/event/event.go
```diff
@@ -85,6 +89,10 @@ type ErrorEvent struct {
 	Err error
 }
 
+type WaitEvent struct {
+	Error error
```
karlkfi, line 93: If we decide to keep this new event, please add a GroupName to match other events and can be associated with a parent ActionGroup

karlkfi (changes requested):
I'm a little confused about the new WaitEvent and why it's not just an ErrorEvent. Do we really want the events sent to be different when ContinueOnError is true vs false?

pkg/apply/taskrunner/runner.go
haiyanmeng: Good catch. Done.

pkg/apply/event/event.go
haiyanmeng: Not all the event types have GroupName.
What is the purpose of GroupName?

karlkfi: Group name is used to correlate apply/prune/delete/wait events (usually for individual objects) with action groups (tasks on groups of objects). It's required to differentiate when there are multiple task of the same type and helps sorting for testing comparison.

pkg/apply/taskrunner/context.go
```diff
@@ -13,14 +13,15 @@ import (
 )
 
 // NewTaskContext returns a new TaskContext
-func NewTaskContext(eventChannel chan event.Event, resourceCache cache.ResourceCache) *TaskContext {
+func NewTaskContext(eventChannel chan event.Event, resourceCache cache.ResourceCache, continueOnError bool) *TaskContext {
```
karlkfi, line 16: Why do we need the continue one error bool in the TaskContext?
Isn't it only needed by the baseruneer?

haiyanmeng: > Why do we need the continue one error bool in the TaskContext?
> Isn't it only needed by the baseruneer?

This is from an older version. Please review the latest version.

pkg/apply/event/event.go
haiyanmeng: Thanks for the explanation. Added it.

haiyanmeng:
@ash2k , let me know if this change breaks your use case.

haiyanmeng:
@seans3 , @mortent , @karlkfi , I updated the PR based on our last discussion. PTAL.

ash2k:
@haiyanmeng I haven't checked the code closely, but I think it shouldn't affect me. Thanks for the ping.

pkg/apply/taskrunner/task.go
```diff
@@ -111,16 +111,23 @@ func (w *WaitTask) setTimer(taskContext *TaskContext) {
...
-		// We only send the taskResult if no one has gotten
+		// We only send the TimeoutError to the eventChannel if no one has gotten
 		// to the token first.
 		case <-w.token:
-			taskContext.TaskChannel() <- TaskResult{
-				Err: &TimeoutError{
-					Identifiers: w.Ids,
-					Timeout:     w.Timeout,
-					Condition:   w.Condition,
+			err := &TimeoutError{
```
mortent, line 117: Nit: I like that we no longer need to amend certain error types in the runner itself. In this case, it seems like we can just create the `TimeoutError` directly without having to have the two steps here.

mortent (approved):
/approve
This looks good to me. I had just one small comment. I'll let @karlkfi or @seans3 do the lgtm

seans3 (approved):
/lgtm

Karl: if you had any other feedback, I can address it quickly (I hope) in a separate PR.

pkg/apply/taskrunner/runner_test.go
```diff
@@ -244,18 +248,13 @@ func TestBaseRunner(t *testing.T) {
...
-						timeoutError.TimedOutResources)
-					assert.Equal(t, timeoutError.Error(), tc.expectedErrorMsg)
-				}
-				return
-			} else if err != nil {
+			if err != nil {
 				t.Errorf("expected no error, but got %v", err)
 			}
 
+			for _, event := range events {
```
karlkfi, line 255: Not sure if you meant to leave this in here. Looks like debugging.
I plan to go through and convert unit tests to test the full event log with the new methods I used for the e2e tests. I can probably clean this up later.

pkg/apply/taskrunner/task.go
karlkfi: `amendTimeoutError` should probably be refactored to just have a function that returns the IDs that timed out, like `pending` except also strip out the ones that are Current or NotFound (for apply or prune).

PR #453, opened 2021-10-28
Ignore receiver name lint error

seans3:
* Ignores lint error caused by string generation, which creates a function receiver name of `i`.
* Lint error is:
```
pkg/common/common.go:94:27: ST1016: methods on the same type should have the same receiver name (seen 4x "drs", 1x "i") (stylecheck)
```

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -90,6 +90,7 @@ const (
 )
 
 // ClientDryRun returns true if input drs is DryRunClient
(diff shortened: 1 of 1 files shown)


dominikh:
> nit: I wish there were a way to ignore a specific rule and not the whole linter, but I don't see one documented.

You could run Staticcheck instead of golangci-lint ;)

PR #454, opened 2021-10-28
Applier constructor and run: functional options

ash2k:
This is a draft for API refactoring using the functional options pattern. The purpose of this PR is to start the conversation on how this can look.

It's easier to look at with the "hide whitespace" diff option enabled.

I suggest to look at this PR in the following order:
1. pkg/apply/applier.go `NewApplier()`. See how the new constructor looks.
1. pkg/apply/applier_options.go. See the new options to configure the applier. New ones may be added later as necessary.
1. pkg/apply/applier.go `Run()`. See how the new signature looks.
1. pkg/apply/applier_run_options.go. See the new options to configure the Run of the applier. New ones may be added later as necessary.
1. See cmd/* and then tests for usage examples.

TODO:
- [ ] Refactor `NewDestroyer()` to use functional options (in a separate PR).
- [ ] Refactor printers to be callbacks rather than channel consumers (in a separate PR). That would allow to remove some goroutines and cleanup a few places with Run() invocations.
- [x] Refactor `Applier.Run()` to use functional options. This is the most interesting part. Ran out of time today.
- [x] Fix tests

WDYT @karlkfi @mortent @seans3

### DIFF
--- a/cmd/apply/cmdapply.go
+++ b/cmd/apply/cmdapply.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/kubectl/pkg/util/i18n"
 	"sigs.k8s.io/cli-utils/cmd/flagutils"
 	"sigs.k8s.io/cli-utils/pkg/apply"
+	"sigs.k8s.io/cli-utils/pkg/apply/event"
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/cli-utils/pkg/inventory"
 	"sigs.k8s.io/cli-utils/pkg/manifestreader"
@@ -137,7 +138,10 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
 
 	// Run the applier. It will return a channel where we can receive updates
 	// to keep track of progress and any issues.
-	a, err := apply.NewApplier(r.factory, invClient)
+	a, err := apply.NewApplier(
+		apply.WithFactory(r.factory),
+		apply.WithInventoryClient(invClient),
+	)
 	if err != nil {
 		return err
 	}
@@ -147,20 +151,24 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
 		r.printStatusEvents = true
 	}
 
-	ch := a.Run(ctx, inv, objs, apply.Options{
-		ServerSideOptions: r.serverSideOptions,
-		PollInterval:      r.period,
-		ReconcileTimeout:  r.reconcileTimeout,
-		// If we are not waiting for status, tell the applier to not
-		// emit the events.
-		EmitStatusEvents:       r.printStatusEvents,
-		NoPrune:                r.noPrune,
-		DryRunStrategy:         common.DryRunNone,
-		PrunePropagationPolicy: prunePropPolicy,
-		PruneTimeout:           r.pruneTimeout,
-		InventoryPolicy:        inventoryPolicy,
-	})
-
+	ch := make(chan event.Event)
+	go func() {
+		defer close(ch)
+		a.Run(ctx, inv, objs,
+			apply.ServerSideOptions(r.serverSideOptions),
+			apply.PollInterval(r.period),
+			apply.ReconcileTimeout(r.reconcileTimeout),
+			// If we are not waiting for status, tell the applier to not
+			// emit the events.
+			apply.EmitStatusEvents(r.printStatusEvents),
+			apply.Prune(!r.noPrune),
+			apply.DryRunStrategy(common.DryRunNone),
+			apply.PrunePropagationPolicy(prunePropPolicy),
+			apply.PruneTimeout(r.pruneTimeout),
+			apply.InventoryPolicy(inventoryPolicy),
+			apply.EventChannelListener(ch),
+		)
+	}()
 	// The printer will print updates from the channel. It will block
 	// until the channel is closed.
 	printer := printers.GetPrinter(r.output, r.ioStreams)
--- a/pkg/apply/applier.go
+++ b/pkg/apply/applier.go
@@ -7,12 +7,12 @@ import (
 	"context"
 	"fmt"
 	"sort"
-	"time"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/klog/v2"
-	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"sigs.k8s.io/cli-utils/pkg/apply/cache"
 	"sigs.k8s.io/cli-utils/pkg/apply/event"
 	"sigs.k8s.io/cli-utils/pkg/apply/filter"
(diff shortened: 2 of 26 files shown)


cmd/apply/cmdapply.go
```diff
@@ -147,20 +151,24 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
...
-		// emit the events.
-		EmitStatusEvents:       r.printStatusEvents,
-		NoPrune:                r.noPrune,
-		DryRunStrategy:         common.DryRunNone,
-		PrunePropagationPolicy: prunePropPolicy,
-		PruneTimeout:           r.pruneTimeout,
-		InventoryPolicy:        inventoryPolicy,
-	})
-
+	ch := make(chan event.Event)
```
ash2k, line 154: Once printers are refactored, this goroutine and channel will go away. A printer (0 or more) will be passed to Run as a callback. Not done here to reduce scope.

pkg/apply/applier.go
```diff
@@ -114,192 +113,139 @@ func (a *Applier) prepareObjects(localInv inventory.InventoryInfo, localObjs obj
 	return localObjs, pruneObjs, nil
 }
 
-// Run performs the Apply step. This happens asynchronously with updates
-// on progress and any errors are reported back on the event channel.
+// Run performs the Apply step. This synchronously with updates
+// on progress and any errors reported back via callbacks.
```
ash2k, line 117: Run is a blocking method now. If caller needs concurrency they can always launch it in a goroutine. A typical caller doesn't need concurrency as it waits for the operation to be done. See tests for an example of a concurrent invocation.

pkg/apply/destroyer.go
```diff
@@ -171,3 +184,12 @@ func (d *Destroyer) Run(ctx context.Context, inv inventory.InventoryInfo, option
 	}()
 	return eventChannel
 }
+
+func handleError(eventChannel chan event.Event, err error) {
```
ash2k, line 188: This will go away once destroyer is refactored to use functional options too.

karlkfi:
Gotta say, I don't love this pattern. Why would we want to use this over just an Options struct?

Down sides:
1. It makes it more difficult to see which options are available when using an IDE, which currently shows me all the fields in the Option struct. 
2. It means that anyone can implement a new option struct that satisfies the interface, but Run will just ignore it, right?
3. Feels less intuitive and more "magic" you need to know to use it.

seans3:
> Gotta say, I don't love this pattern. Why would we want to use this over just an Options struct?
> 
> Down sides:
> 
> 1. It makes it more difficult to see which options are available when using an IDE, which currently shows me all the fields in the Option struct.
> 2. It means that anyone can implement a new option struct that satisfies the interface, but Run will just ignore it, right?
> 3. Feels less intuitive and more "magic" you need to know to use it.

According to Dave Cheney when discussing the Config struct pattern:
```
It has trouble with defaults, especially if the zero value has a well understood meaning.
```

Here's the full `Functional Options` post: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

karlkfi:
I have an alternate suggestion that i think would solve both the defaults issues and still make the API discoverable:

- Use an options struct, but make it private so it can't be constructed with zero values by the user. 
- Use an interface for the options that the options struct satisfies, for use as the function argument type, instead of the options struct.
- Create a public constructor that sets the defaults and returns the options struct
- Create functional option setters that update and return the options struct using the `opts.WithX(...)` pattern so they can be chained.

This has the advantages:
- defaults are always set, don't need to match the zero value in Go, and don't always need to be specified by the user.
- all possible options are discoverable using IDE auto-complete on the options struct.
- option setters allow custom logic based on user input, so the user input doesn't need to exactly match the field type stored in the options struct.

Example usage:
```
applier.Run(ctx, inv, objs, apply.DefaultOptions().
    WithServerSideOptions(r.serverSideOptions).
    WithReconcileTimeout(r.reconcileTimeout).
    WithDryRunStrategy(common.DryRunNone))
```

ash2k:
@karlkfi Thanks for your feedback!

> It makes it more difficult to see which options are available when using an IDE, which currently shows me all the fields in the Option struct.

I partially agree with this. In practice if we move all/most of the stuff, that shouldn't be public away from the `apply` package, it will almost exclusively contain relevant options (and the constructor).

> It means that anyone can implement a new option struct that satisfies the interface, but Run will just ignore it, right?

Not really. You cannot implement a function to implement the type because it takes a parameter of a private type - see https://github.com/kubernetes-sigs/cli-utils/pull/454/files#diff-44c477dc998e3b6be3729c3697d977a86de5fcd680dd8cee6a181e2d3f94ce9cR35 and https://github.com/kubernetes-sigs/cli-utils/pull/454/files#diff-ff0604c07cd58dd82246f24045636317c9ce34bb516d74c599511042f5e8313bR60.

Overall, this is a well known and widely used pattern in Go. For example, [grpc-go](https://github.com/grpc/grpc-go/) uses it. As a user of that library, I think it's quite good in terms of UX.

> Feels less intuitive and more "magic" you need to know to use it.

It depends if you have used it before. As I said above, it's a widely known and used pattern in Go. If you have worked with it before, you'll feel right at home.

> I have an alternate suggestion

To me as a user that reminds Java's builders :) I think this is fine, but less commonly used in Go.

> defaults are always set, don't need to match the zero value in Go, and don't always need to be specified by the user.

Same as with functional options.

> all possible options are discoverable using IDE auto-complete on the options struct.

This is good, but with func. options it's package name (e.g. `apply`) vs the private struct. Autocomplete works on both, but with functional options you get a bit of unrelated stuff too (depends on the IDE, I guess).

----

I think both approaches are better than what we have now. As I user, I personally would prefer functional options because it's more idiomatic in Go, but it's just me. 🤷

karlkfi:
How is the functional options pattern more idiomatic than the builder pattern?

I don't really buy that Functional Options is idiomatic just because Dave Cheney made a blog about it. And just because Builder pattern is popular in Java doesn't mean it's bad or not idiomatic Go. These seem like spurious arguments.

Consistency with other big projects is mildly compelling, but would be more compelling if it were something more closely related, like K8s itself. It's still sort of an appeal to authority tho.

I can sort of get on board with functional options being better than the existing options struct, but I see it more as a tradeoff than a solid win. And I'm not convinced it's better than the build pattern.

That said, none of this seems like a huge priority. Even if it's an improvement, it's still an API change that's going to require refactoring all the existing clients, without any obvious return on investment for existing users.

seans3:
> How is the functional options pattern more idiomatic than the builder pattern?
> 
> I don't really buy that Functional Options is idiomatic just because Dave Cheney made a blog about it. And just because Builder pattern is popular in Java doesn't mean it's bad or not idiomatic Go. These seem like spurious arguments.
> 
> Consistency with other big projects is mildly compelling, but would be more compelling if it were something more closely related, like K8s itself. It's still sort of an appeal to authority tho.
> 
> I can sort of get on board with functional options being better than the existing options struct, but I see it more as a tradeoff than a solid win. And I'm not convinced it's better than the build pattern.
> 
> That said, none of this seems like a huge priority. Even if it's an improvement, it's still an API change that's going to require refactoring all the existing clients, without any obvious return on investment for existing users.

Either builder, functional options, options struct, or a hybrid, we are going to simplify the API. The return on investment for the two current clients are that they will have a stable API. The two current existing clients are going to have to change no matter what, because we are certainly going to simplify the API to make it more accessible to other clients. The API is not even `alpha` yet. It is much better to change it now that there are only two known clients, instead of trying to update it later when there may be more. Specifically, the simplest default API will not assume a custom inventory object. In fact, the simplest case will not activate pruning functionality at all. The simplest case will be to merely activate the wrapped `kubectl apply` functionality--no pruning, depends on, or apply-time mutation. And the API will support this as the default. The SIG CLI refactored dependencies literally for years by moving the code base into `staging`, which finally allows `kubectl` code reuse. And this open source effort in a SIG CLI repository is one of the first known reuses of `kubectl apply`. The open source community deserves a simple, accessible API to this library, and it will get it.

ash2k:
I've opened https://github.com/kubernetes-sigs/cli-utils/pull/485 with an alternative approach with the builder pattern.

Issue #452, opened 2021-10-27
Bug: WaitTask waits for filtered resources


karlkfi:
WaitTask currently ignores failed applies and failed prunes, but doesn't ignore resources skipped by filters in the ApplyTask or WaitTask. 

Because WaitTasks are only added when there are dependencies (depends-on, CRDs, namespaces, etc.), this hasn't been much of a problem yet, but when using a sufficiently large/complex set of resources, it's feasible for the WaitTask to wait until timeout and fail because it's waiting for one of the following:
- [deletion prevention, inventory policy, local namespace] Expecting a delete that was skipped to cause that object to become NotFound
- [inventory policy] Expecting an apply that was skipped to cause the object (previously different or broken) to become Current

Suggested fix:
- Add tracking of skipped deletions and applies to the TaskContext (similar to the existing deletion tracking)
- Update WaitTask to ignore skipped deletions and applies

PR #456, opened 2021-10-29
fix: skipped deletes no longer cause waiting

karlkfi:
- Added SkippedApplies and SkippedDeletes to the TaskContext
- Modified tasks to use the new skipped tracking, replacing usage
  of failure tracking, where skipped is more accurate.
- Renamed some TaskContext methods for consistency
- Added ObjMetadataSetFromMap for use by TaskContext
- Added ObjectMetadataSet.Intersection for use by InvSetTask
- Cleaned up InvSetTask to be more readable with comments explaining
  intended behavior, including handling of skips.
- Added apply and prune tests for skipped, failure, and abandoned

Fixes: #452


mortent (approved):
/lgtm

PR #457, opened 2021-10-29
Removes inventory from manifestloader

seans3:
* Removes `InventoryInfo` from `ManifestLoader` interface
* Removes `NoInventoryObjError` from `inventory.SplitUnstructureds`

This is part of the ongoing effort to remove incorrect assumption that pruning functionality must be used when using the apply library. If clients want to make a missing inventory object an error, they will have to validate themselves now.


ash2k (approved):
/lgtm

PR #460, opened 2021-11-01
fix: TimeoutError only returns objs that timed out

karlkfi:
- Previous behavior returned all objects being waited on
- Updated WaitTask tests to wait for all TaskResults and Events

Fixes an old bug and cleans up some tech debt from https://github.com/kubernetes-sigs/cli-utils/pull/451

This change is probably mutually exclusive to https://github.com/kubernetes-sigs/cli-utils/pull/463, because with this change uses a single bulk WaitEvent and that change replaces it with individual WaitEvents.

### DIFF
--- a/pkg/apply/taskrunner/task.go
+++ b/pkg/apply/taskrunner/task.go
@@ -114,17 +114,11 @@ func (w *WaitTask) setTimer(taskContext *TaskContext) {
 		// We only send the TimeoutError to the eventChannel if no one has gotten
 		// to the token first.
 		case <-w.token:
-			err := &TimeoutError{
-				Identifiers: w.Ids,
-				Timeout:     w.Timeout,
-				Condition:   w.Condition,
-			}
-			amendTimeoutError(taskContext, err)
 			taskContext.EventChannel() <- event.Event{
 				Type: event.WaitType,
 				WaitEvent: event.WaitEvent{
 					GroupName: w.Name(),
-					Error:     err,
+					Error:     w.timeoutError(taskContext),
 				},
 			}
 			taskContext.TaskChannel() <- TaskResult{}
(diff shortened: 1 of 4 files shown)


pkg/apply/taskrunner/task.go
```diff
@@ -137,21 +131,29 @@ func (w *WaitTask) setTimer(taskContext *TaskContext) {
...
-				Message:    result.StatusMessage,
-			})
+// timeoutError returns an error with the objects that have not yet reached the
+// desired condition.
+func (w *WaitTask) timeoutError(taskContext *TaskContext) *TimeoutError {
+	expected := w.pending(taskContext)
+	ids := object.ObjMetadataSet{}
+	tors := []TimedOutResource{}
+	for _, id := range expected {
+		result := taskContext.ResourceCache().Get(id)
```
mortent, line 141: So will this call always return the same version of the resource as was used by the `WaitTask` to determine that it failed to reconcile? I'm wondering if there is a chance where we determine the resource has failed to reconcile before the timeout, but does so just afterwards and ends up not being included here?

karlkfi: This timeout error includes the IDs of all the objects that are still not reconciled at timeout.

If it reconciles after timeout, it won't be included. Tho there's a lag between timeout and all events being sent, because the caller needs to be draining the event channel.

If that's something we want to defend against, we could use two for loops, one to assess and make a list of IDs, and another to send the events.

Either way, the TaskResult wont be sent until after this returns. So if it reconciled while sending events it will still count as a timeout error.

karlkfi:
/hold

Need to decide between this PR and https://github.com/kubernetes-sigs/cli-utils/pull/463

karlkfi:
This is obsolete. We're going to go with https://github.com/kubernetes-sigs/cli-utils/pull/463 instead, which makes for better UX and simpler printers.

PR #461, opened 2021-11-01
Cherry-pick two commits

haiyanmeng:


mortent:
I think you need to run `make generate` after you added the new `WaitTask`.

seans3:
> I think you need to run `make generate` after you added the new `WaitTask`.

Usually best to just run `make all` before creating a PR, since it also formats, vets, and lints.

PR #449, opened 2021-10-27
feat: Always wait for reconciliation

karlkfi:
- Converted WaitTask to use a context for timeout/cancellation, to
  improve readability and reduce error cases. Now it only sends
  TaskResult events from one place, removing the need for a token to
  silence subsequent complete() calls.
- Add pending object tracking to WaitTask to ensure all objects are
  accounted for at least one WaitEvent.
- Upgraded applier tests to use full event comparison, for better signal
  on breaking changes.
- Enhanced task tests to consume all events before validating and test
  actual event output.
- Add set sorting to graph.SortObjs, to make wait event ordering more
  consistent, to make testing easier.

BREAKING CHANGE: wait tasks always execute after apply/prune/delete (except dry runs)
BREAKING CHANGE: wait tasks default to waiting until cancelled (previously 1m default)

fixes: https://github.com/kubernetes-sigs/cli-utils/issues/438

PR #463, opened 2021-11-02
feat: send WaitEvent for every resource

karlkfi:
- WaitEvent can be Pending, Reconciled, Skipped, or Timeout.
  Skipped, Pending, and Reconciled events are sent at task start.
  Reconciled events are sent later as status updates are recieved.
  Timeout events are sent for remaining events on timeout.
- Rewrite WaitTask.Start to use context.WithTimeout and a goroutine to
  handle task completion (replacing setTimer and the token hack).
- Replaced Task.ClearTimeout with Task.Cancel.
- Replaced WaitTask.complete & checkCondition with Task.StatusUpdate.
- Replaced WaitTask.startAndComplete with a check in WaitTask.Start.
- Replaced WaitTask.amendTimeoutError with WaitTask.sendTimeoutEvents
  to send multiple timeout events, instead of one event with a list of
  TimedOutResources.
- Updated all printers to handle WaitEvent.
  Event printer now includes reconcile events.
  JSON printer now includes resourceReconciled eventType.
  Table printer not includes reconcile column.
- Added JSON printer tests for error handling.
- Updated Formatter.FormatActionGroupEvent to collect WaitStats.
- Enable status events by default for kapply with table output

BREAKING CHANGE: WaitEvents now sent for each object

This fixes tech debt introduced in https://github.com/kubernetes-sigs/cli-utils/pull/451, which neglected to update the printers. Updating the printers made it clear that having a WaitEvent per object would make it much more consistent with the apply/prune/delete event handling.

This change is probably mutually exclusive to https://github.com/kubernetes-sigs/cli-utils/pull/460, because with this change the single bulk WaitEvents is replaced with individual WaitEvents.


karlkfi:
Example event printer output:
```
$ kapply apply examples/mutation/
pod/pod-b created
pod/pod-b reconcile pending
pod/pod-b reconciled
pod/pod-a created
2 resource(s) applied. 2 created, 0 unchanged, 0 configured, 0 failed
pod/pod-a reconcile pending
pod/pod-a reconciled
2 resource(s) reconciled, 0 skipped

$ kapply destroy examples/mutation/
pod/pod-a deleted
pod/pod-a reconcile pending
pod/pod-a reconciled
pod/pod-b deleted
2 resource(s) deleted, 0 skipped
pod/pod-b reconcile pending
pod/pod-b reconciled
2 resource(s) reconciled, 0 skipped
```

Example JSON printer output:
```
$ kapply apply examples/mutation/ --output json
{"eventType":"resourceApplied","group":"","kind":"Pod","name":"pod-b","namespace":"test","operation":"Created","timestamp":"2021-11-03T00:31:58Z","type":"apply"}
{"configuredCount":0,"count":1,"createdCount":1,"eventType":"completed","failedCount":0,"serverSideCount":0,"timestamp":"2021-11-03T00:31:58Z","type":"apply","unchangedCount":0}
{"eventType":"resourceReconciled","group":"","kind":"Pod","name":"pod-b","namespace":"test","operation":"Pending","timestamp":"2021-11-03T00:31:58Z","type":"wait"}
{"eventType":"resourceReconciled","group":"","kind":"Pod","name":"pod-b","namespace":"test","operation":"Reconciled","timestamp":"2021-11-03T00:32:00Z","type":"wait"}
{"eventType":"completed","reconciled":1,"skipped":0,"timestamp":"2021-11-03T00:32:00Z","type":"wait"}
{"eventType":"resourceApplied","group":"","kind":"Pod","name":"pod-a","namespace":"test","operation":"Created","timestamp":"2021-11-03T00:32:00Z","type":"apply"}
{"configuredCount":0,"count":2,"createdCount":2,"eventType":"completed","failedCount":0,"serverSideCount":0,"timestamp":"2021-11-03T00:32:00Z","type":"apply","unchangedCount":0}
{"eventType":"resourceReconciled","group":"","kind":"Pod","name":"pod-a","namespace":"test","operation":"Pending","timestamp":"2021-11-03T00:32:00Z","type":"wait"}
{"eventType":"resourceReconciled","group":"","kind":"Pod","name":"pod-a","namespace":"test","operation":"Reconciled","timestamp":"2021-11-03T00:32:02Z","type":"wait"}
{"eventType":"completed","reconciled":2,"skipped":0,"timestamp":"2021-11-03T00:32:02Z","type":"wait"}

$ kapply destroy examples/mutation/ --output json
{"eventType":"resourceDeleted","group":"","kind":"Pod","name":"pod-a","namespace":"test","operation":"Deleted","timestamp":"2021-11-03T00:32:12Z","type":"delete"}
{"deleted":1,"eventType":"completed","skipped":0,"timestamp":"2021-11-03T00:32:12Z","type":"delete"}
{"eventType":"resourceReconciled","group":"","kind":"Pod","name":"pod-a","namespace":"test","operation":"Pending","timestamp":"2021-11-03T00:32:12Z","type":"wait"}
{"eventType":"resourceReconciled","group":"","kind":"Pod","name":"pod-a","namespace":"test","operation":"Reconciled","timestamp":"2021-11-03T00:32:16Z","type":"wait"}
{"eventType":"completed","reconciled":1,"skipped":0,"timestamp":"2021-11-03T00:32:16Z","type":"wait"}
{"eventType":"resourceDeleted","group":"","kind":"Pod","name":"pod-b","namespace":"test","operation":"Deleted","timestamp":"2021-11-03T00:32:16Z","type":"delete"}
{"deleted":2,"eventType":"completed","skipped":0,"timestamp":"2021-11-03T00:32:16Z","type":"delete"}
{"eventType":"resourceReconciled","group":"","kind":"Pod","name":"pod-b","namespace":"test","operation":"Pending","timestamp":"2021-11-03T00:32:16Z","type":"wait"}
{"eventType":"resourceReconciled","group":"","kind":"Pod","name":"pod-b","namespace":"test","operation":"Reconciled","timestamp":"2021-11-03T00:32:20Z","type":"wait"}
{"eventType":"completed","reconciled":2,"skipped":0,"timestamp":"2021-11-03T00:32:20Z","type":"wait"}
```

Example table printer output:
```
$ kapply apply examples/mutation/ --output table
NAMESPACE   RESOURCE                                  ACTION        STATUS      RECONCILED  CONDITIONS                                AGE     MESSAGE                                 
test        Pod/pod-a                                               Current     Reconciled  Initialized,Ready,ContainersReady,PodSch  3s      Pod is Ready                            
test        Pod/pod-b                                               Current     Reconciled  Initialized,Ready,ContainersReady,PodSch  5s      Pod is Ready 

$ kapply destroy examples/mutation/ --output table
NAMESPACE   RESOURCE                                  ACTION        STATUS      RECONCILED  CONDITIONS                                AGE     MESSAGE                                 
test        Pod/pod-a                                               NotFound    Reconciled  -                                         -       Resource not found                      
test        Pod/pod-b                                               NotFound    Reconciled  -                                         -       Resource not found 
```

pkg/apply/taskrunner/task.go
```diff
@@ -4,6 +4,7 @@
 package taskrunner
```
karlkfi, line 4: Note: this file has so many changes that it's effectively a rewrite. It's probably easier to review the result, rather than the diff.

pkg/apply/taskrunner/task_test.go
```diff
@@ -4,98 +4,541 @@
 package taskrunner
```
karlkfi, line 4: Note: this file has so many changes that it's effectively a rewrite. It's probably easier to review the result, rather than the diff.

cmd/apply/cmdapply.go
```diff
@@ -108,14 +108,15 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	// Only print status events if we are waiting for status.
-	//TODO: This is not the right way to do this. There are situations where
-	// we do need status events event if we are not waiting for status. The
-	// printers should be updated to handle this.
 	var printStatusEvents bool
+	// Print status events if a wait timeout is specified
 	if r.reconcileTimeout != time.Duration(0) || r.pruneTimeout != time.Duration(0) {
```
karlkfi, line 113: This default doesn't make a lot of sense with the current default of 1min, but I wasn't sure what to change it to and it's just kapply so it's not super important.

cmd/destroy/cmddestroy.go
```diff
@@ -127,9 +127,19 @@ func (r *DestroyRunner) RunE(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+
+	var printStatusEvents bool
+	// Print status events if a wait timeout is specified
+	if r.deleteTimeout != time.Duration(0) {
```
karlkfi, line 133: This default doesn't make a lot of sense with the current default of 1min, but I wasn't sure what to change it to and it's just kapply so it's not super important.

pkg/apply/taskrunner/task.go
```diff
@@ -26,25 +27,20 @@ type Task interface {
 	Name() string
 	Action() event.ResourceAction
 	Identifiers() object.ObjMetadataSet
-	Start(taskContext *TaskContext)
-	ClearTimeout()
+	Start(*TaskContext)
+	StatusUpdate(*TaskContext, object.ObjMetadata)
+	Cancel(*TaskContext)
```
karlkfi, line 32: Note: this Cancel func is really just a placeholder. It will probably be removed when we pipe context through the TaskContext to handle cancellation and timeouts.

pkg/print/list/base.go
```diff
@@ -172,16 +200,36 @@ func (b *BaseListPrinter) Print(ch <-chan event.Event, previewStrategy common.Dr
 			if err := formatter.FormatDeleteEvent(e.DeleteEvent); err != nil {
 				return err
 			}
+		case event.WaitType:
```
karlkfi, line 203: TODO: This needs to be update for the new WaitEvent Operations. -Error +Pending +Timeout.

pkg/print/list/base.go
```diff
@@ -172,16 +200,36 @@ func (b *BaseListPrinter) Print(ch <-chan event.Event, previewStrategy common.Dr
...
+				statusCollector,
+			); err != nil {
 				return err
 			}
 		}
 	}
 	failedSum := applyStats.Failed + pruneStats.Failed + deleteStats.Failed
-	if failedSum > 0 {
-		return fmt.Errorf("%d resources failed", failedSum)
+	if failedSum > 0 || waitStats.Failed > 0 {
```
karlkfi, line 231: TODO: This needs to be update for the new WaitEvent Operations. There's no more Failed/Error. Do we want to return Timeouts in the error?

pkg/testutil/events.go
```diff
@@ -248,16 +249,21 @@ func isMatch(ee ExpEvent, e event.Event) bool {
...
 				return false
 			}
 		}
 
-		if wee.Error != nil {
-			return cmp.Equal(wee.Error, we.Error, cmpopts.EquateErrors())
+		if wee.Operation != we.Operation {
+			return false
 		}
-		return we.Error == nil
```
karlkfi, line 260: maybe add an explicit `return true`?

pkg/testutil/events.go
```diff
@@ -402,6 +409,12 @@ func (ape GroupedEventsByID) Less(i, j int) bool {
 			return false
 		}
 		return ape[i].DeleteEvent.Identifier.String() < ape[j].DeleteEvent.Identifier.String()
+	case event.WaitType:
```
karlkfi, line 412: I believe the WaitEvents are actually entirely predictable. So we might not need to sort them for test comparisons.

mortent (commented):
Overall I like the approach. Most of the changes seems to actually be in the printers.

On the output format, I have some thoughts:
- Do we want to print both the StatusEvents and the WaitEvents? The output becomes pretty busy with both. The example output in the PR is a bit misleading here, as it doesn't provide any reconcile/prune timeouts and therefore doesn't get the status events in the output. It does however get the wait behavior due to the 1min default.
- We do change the output here to include the WaitEvents, but as far as I can tell we don't actually change any of the other events. So for the `events` and `json` output formats, it is only adds new events. Is this correct? Of course, the behavior itself changes somewhat with the continue-on-error behavior for wait timeouts.

karlkfi:
@mortent 
> Do we want to print both the StatusEvents and the WaitEvents? The output becomes pretty busy with both. The example output in the PR is a bit misleading here, as it doesn't provide any reconcile/prune timeouts and therefore doesn't get the status events in the output. It does however get the wait behavior due to the 1min default.

By default, the status events aren't printed. It's opt-in. And I think now with the WaitEvents and https://github.com/kubernetes-sigs/cli-utils/pull/449 to make it always wait, I think the wait events are actually more useful to the end users. 
- for the event printer, status events aren't really required any more, but left in for completeness
- for the json printer, some clients might want status events, just to tell whether the resource is Current or NotFound (since the WaitEvent doesn't distinguish)
- for the table printer, I think the combination of permutations from the status and reconciled columns are pretty useful, but I could be convinced to remove the status column, if we get user feedback about it.

> We do change the output here to include the WaitEvents, but as far as I can tell we don't actually change any of the other events. So for the events and json output formats, it is only adds new events. Is this correct? Of course, the behavior itself changes somewhat with the continue-on-error behavior for wait timeouts.

This is correct for the json printer. The event printer also had the error text changed. And the table printer got a new column, but other columns weren't changed.

karlkfi:
Here's the event output with status events:
```
$ kapply apply examples/mutation/ --reconcile-timeout 1m
pod/pod-a is NotFound: Resource not found
pod/pod-b is NotFound: Resource not found
pod/pod-b created
pod/pod-b is NotFound: Resource not found
pod/pod-b reconcile pending
pod/pod-b is InProgress: Pod is in the Pending phase
pod/pod-b is Current: Pod is Ready
pod/pod-b reconciled
pod/pod-a created
2 resource(s) applied. 2 created, 0 unchanged, 0 configured, 0 failed
pod/pod-a is NotFound: Resource not found
pod/pod-a reconcile pending
pod/pod-a is Current: Pod is Ready
pod/pod-a reconciled
2 resource(s) reconciled, 0 skipped

$ kapply destroy examples/mutation/ --delete-timeout 1m
pod/pod-a deleted
pod/pod-a reconcile pending
pod/pod-b is Current: Pod is Ready
pod/pod-a is Current: Pod is Ready
pod/pod-a is NotFound: Resource not found
pod/pod-a reconciled
pod/pod-b deleted
2 resource(s) deleted, 0 skipped
pod/pod-b is Current: Pod is Ready
pod/pod-b reconcile pending
pod/pod-b is Terminating: Resource scheduled for deletion
pod/pod-b is NotFound: Resource not found
pod/pod-b reconciled
2 resource(s) reconciled, 0 skipped
```

mortent:
@karlkfi 
Yeah, but the opt-in mechanism here is providing a reconcile timeout or a prune timeout right? https://github.com/kubernetes-sigs/cli-utils/pull/463/files#diff-5d5c80857e24e895cf402523dabe1e2c4918cfafb25b582adbd7b3ba4f8cdd97R113. It seems like it might be useful to decouple it from providing custom timeout values (in particular as we want to change the default behavior here).

karlkfi:
@mortent 
> Yeah, but the opt-in mechanism here is providing a reconcile timeout or a prune timeout right? https://github.com/kubernetes-sigs/cli-utils/pull/463/files#diff-5d5c80857e24e895cf402523dabe1e2c4918cfafb25b582adbd7b3ba4f8cdd97R113. It seems like it might be useful to decouple it from providing custom timeout values (in particular as we want to change the default behavior here).

Correct, and I agree with you, but that's what the code was already doing. I just added the bit to make it auto-enable for table output, because otherwise the status column sits empty. I guess an alternative would be to add a new way to hide the status column in that printer. Or we could add new flags to the kapply commands like `--status-events`. I just didn't want to expand the PR any more than it already is. Could move it out to another tho. Got a preference?

karlkfi:
Pulled out race fix to https://github.com/kubernetes-sigs/cli-utils/pull/468

pkg/apply/event/type_string.go
```diff
@@ -15,11 +15,12 @@ func _() {
 	_ = x[StatusType-4]
 	_ = x[PruneType-5]
 	_ = x[DeleteType-6]
+	_ = x[WaitType-7]
 }
 
-const _Type_name = "InitTypeErrorTypeActionGroupTypeApplyTypeStatusTypePruneTypeDeleteType"
+const _Type_name = "InitTypeErrorTypeActionGroupTypeApplyTypeStatusTypePruneTypeDeleteTypeWaitType"
```
haiyanmeng, line 22: You need to rebase this PR. The change to this file has been in the master already: https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/apply/event/type_string.go#L21

pkg/apply/event/event.go
```diff
@@ -89,9 +89,20 @@ type ErrorEvent struct {
 	Err error
 }
 
+//go:generate stringer -type=WaitEventOperation -linecomment
+type WaitEventOperation int
+
+const (
+	ReconcilePending WaitEventOperation = iota // Pending
```
haiyanmeng, line 96: A WaitEvent may be generated when a resource to be pruned has not reached the `NotFound` status. Is it okay using the `Reconcile` prefix for both apply and prune?

pkg/apply/taskrunner/task.go
```diff
@@ -96,133 +84,180 @@ func (w *WaitTask) Identifiers() object.ObjMetadataSet {
...
+			w.sendEvent(taskContext, id, event.ReconcileTimeout)
 		}
-		timeoutErr.TimedOutResources = timedOutResources
 	}
 }
 
-// checkCondition checks whether the condition set in the task
-// is currently met given the status of resources in the cache.
-func (w *WaitTask) checkCondition(taskContext *TaskContext) bool {
+// reconciled checks whether the condition set in the task is currently met for
```
haiyanmeng, line 174: `reconciled` -> `reconciledByID`

pkg/apply/taskrunner/task.go
```diff
@@ -96,133 +84,180 @@ func (w *WaitTask) Identifiers() object.ObjMetadataSet {
...
-// ClearTimeout cancels the timeout for the wait task.
-func (w *WaitTask) ClearTimeout() {
+// Cancel exits early with a timeout error
+func (w *WaitTask) Cancel(_ *TaskContext) {
 	w.cancelFunc()
 }
 
+// StatusUpdate validates whether the update meets the conditions to stop
+// the wait task. If the status is for a watched object and that object now
+// meets the desired condition, a WaitEvent will is sent before exiting.
```
haiyanmeng, line 218: `is` -> `be`

pkg/print/list/base.go
```diff
@@ -16,8 +16,17 @@ type Formatter interface {
 	FormatStatusEvent(se event.StatusEvent) error
 	FormatPruneEvent(pe event.PruneEvent) error
 	FormatDeleteEvent(de event.DeleteEvent) error
+	FormatWaitEvent(we event.WaitEvent) error
 	FormatErrorEvent(ee event.ErrorEvent) error
-	FormatActionGroupEvent(age event.ActionGroupEvent, ags []event.ActionGroup, as *ApplyStats, ps *PruneStats, ds *DeleteStats, c Collector) error
+	FormatActionGroupEvent(
```
haiyanmeng, line 21: I am not sure it is necessary to put every parameter into a separate line.

pkg/print/list/base.go
```diff
@@ -172,16 +200,36 @@ func (b *BaseListPrinter) Print(ch <-chan event.Event, previewStrategy common.Dr
...
 				return err
 			}
+		case event.WaitType:
+			switch e.WaitEvent.Operation {
+			case event.Reconciled:
+				waitStats.incReconciled()
+			case event.ReconcileSkipped:
+				waitStats.incSkipped()
+			}
+			if e.DeleteEvent.Error != nil {
```
haiyanmeng, line 210: `DeleteEvent` -> `WaitEvent`

pkg/print/list/base.go
karlkfi: The conventional patterns are to either use a single line or one line per argument. But the line length convention in this repo wraps and creates something in between the two conventions. I tend to bias towards line per argument when reaching the character limit.

karlkfi: Looks like Google has conventions for this, but I don't know if there's a public equivalent: go/go-style/decisions#func-formatting

pkg/print/list/base.go
karlkfi: hows it look now?

pkg/apply/event/type_string.go
karlkfi: done

pkg/apply/event/event.go
karlkfi: I'm using "reconcile" to mean both here. Reconciled means the desired state is now the current state.

pkg/print/list/base.go
karlkfi: fixed by removing. there's no WaitEvent.Error now, just a Timeout operation

pkg/print/list/base.go
karlkfi: Unfortunately, the Google conventions don't use an 80 character max line (which this repo sort of does). So the conventions on when to wrap and when not to wrap don't cleanly apply.

pkg/apply/taskrunner/task.go
karlkfi: fixed.

pkg/apply/taskrunner/task.go
karlkfi: fixed.

pkg/apply/taskrunner/task_test.go
```diff
@@ -4,98 +4,541 @@
...
+kind: Deployment
+metadata:
+  name: d
+  namespace: default
+  uid: dep-uid-d
+  generation: 1
+spec:
+  replicas: 4
+`
+
```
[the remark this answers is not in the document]
karlkfi: `condition_test.go` definitely tests this, and we're now delegating to `conditionMet` everywhere. Before there was a manual check in `amendTimeoutError` doing `if timeoutErr.Condition.Meets(result.Status) {`. 

Still want an extra test?

karlkfi:
Pulled out another race fix to https://github.com/kubernetes-sigs/cli-utils/pull/469

pkg/printers/json/formatter.go
```diff
@@ -119,6 +132,14 @@ func (jf *formatter) FormatActionGroupEvent(age event.ActionGroupEvent, ags []ev
 			}
 		}
 	}
+
+	if age.Action == event.WaitAction && age.Type == event.Finished {
+		return jf.printEvent("wait", "completed", map[string]interface{}{
+			"reconciled": ws.Reconciled,
+			"skipped":    ds.Skipped,
```
karlkfi, line 139: TODO: add timeout count

karlkfi:
rebased

karlkfi:
Extracted the kapply cmd changes to https://github.com/kubernetes-sigs/cli-utils/pull/470

seans3 (approved):
Nice tests. It is a large PR, but it is good to see that a lot of it is tests.

This is cleaner, and it looks like there are more possible refactorings for future cleanups. Specifically, I believe we should not have conditional code for sending`Status` events. We should just probably pass those events, and let the consumer decide to process them or not. The idea of `ActionGroups` might do with cleanups as well.

You are correct to not prioritize code under `cmd` (e.g. cmdapply), since this is not part of the API. It is only used for documentation and testing.

The code under `pkg/printers` has only recently been moved from `cmd` to `pkg`, and should not be imported by any client yet. This code will be used as standard printers for upcoming function `RunAndPrint` (or something like it that does not rely on the low-level Events and the EventChannel).

/lgtm

pkg/print/list/base.go
karlkfi: Agreed. The stats objects should probably all be wrapped into a single object, but it didn't seem particularly high priority.

PR #465, opened 2021-11-02
Exports Applier/Destroyer StatusPoller field

seans3:
* Exports `StatusPoller` fields for `Applier` and `Destroyer` so they can be overwritten if necessary.
* Removes the status poller parameter from `NewApplier` and `NewDestroyer`, assigning the default status poller instead.
* The two known library clients use the default status poller, so this will result in only code removal.


karlkfi (approved):
/lgtm

PR #466, opened 2021-11-02
Fixes the error returned from running `make verify-kapply-e2e` locally

haiyanmeng:
Running `make verify-kapply-e2e` locally on kind v0.11.1 go1.16.7 linux/amd64 failed because
`kubectl get all -n hellospace` returns `No resources found.` instead of
`No resources found in hellospace namespace.`

### DIFF
--- a/examples/alphaTestExamples/helloapp.md
+++ b/examples/alphaTestExamples/helloapp.md
@@ -171,7 +171,7 @@ expectedOutputLine "3 resource(s) applied. 0 created, 0 unchanged, 0 configured,
 
 # Verify that preview didn't create any resources.
 kubectl get all -n hellospace > $OUTPUT/status 2>&1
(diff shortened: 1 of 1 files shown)


mortent:
Do we know why this changed and why it wasn't caught in previous PRs?

haiyanmeng:
My best guess is that the prow presubmit jobs may use a different version of kind, which may use a different version of kubectl.
Another factor may be that no one really has run `make verify-kapply-e2e` locally before.

karlkfi:
My example tests pass locally on kind 0.11.1. Are you sure it has to do with kind and not your kubectl version?

haiyanmeng:
> My example tests pass locally on kind 0.11.1. Are you sure it has to do with kind and not your kubectl version?

You are right. The problem went away after I upgraded my kubectl. Thanks!

PR #468, opened 2021-11-03
fix: race condition in BuildInfo

karlkfi:
Historically, the objects in ApplyTask were being modified in the
Start func by BuildInfo to remove the path annotation, but this can
cause a race condition in the task runner if it tried to read the
task objects.

Now a DeepCopy is made in BuildInfo instead. This avoids the race
condition but broke mutation, which executes after BuildInfo on the
original object. So we now extract the object from the into after
BuildInfo and use that instead for mutations, filters, and events.

### DIFF
--- a/pkg/apply/task/apply_task.go
+++ b/pkg/apply/task/apply_task.go
@@ -102,6 +102,9 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
 			// Set the client and mapping fields on the provided
 			// info so they can be applied to the cluster.
 			info, err := a.InfoHelper.BuildInfo(obj)
+			// BuildInfo strips path annotations.
+			// Use modified object for filters, mutations, and events.
+			obj = info.Object.(*unstructured.Unstructured)
 			id := object.UnstructuredToObjMetaOrDie(obj)
 			if err != nil {
 				if klog.V(4).Enabled() {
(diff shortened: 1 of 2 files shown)


pkg/apply/task/apply_task.go
```diff
@@ -102,6 +102,9 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
 			// Set the client and mapping fields on the provided
 			// info so they can be applied to the cluster.
 			info, err := a.InfoHelper.BuildInfo(obj)
+			// BuildInfo strips path annotations.
+			// Use modified object for filters, mutations, and events.
+			obj = info.Object.(*unstructured.Unstructured)
```
mortent, line 107: I agree that making a copy of the input resource to the library before making changes is a better way here.
Reusing the `obj` name here is a bit confusing. Could we maybe write this as:
```golang
for i := range objects {
    info, err := a.InfoHelper.BuildInfo(objects[i])
    obj := info.Object.(*unstructured.Unstructured)
```

karlkfi: Reusing the object from the for loop is technically slightly more efficient. It re-uses the variable instead of a new variable per iteration being garbage collected. Not sure the memory or performance impact really matters, but it seems like an arbitrary change. The comment should be enough to avoid confusion.

PR #469, opened 2021-11-03
fix: race condition in PruneTask

karlkfi:
The Pruner modifies the input object when the deletion prevention
annotation is present. This in turn modifies the object map, which
creates a race condition when reading the object map from the task
runner. To fix the race, a copy of the object is made, but only if
the deletion prevention filter prevents deletion.

### DIFF
--- a/pkg/apply/prune/prune.go
+++ b/pkg/apply/prune/prune.go
@@ -119,7 +119,7 @@ func (p *Pruner) Prune(
 				// If deletion was prevented, remove the inventory annotation.
 				if pruneFilter.Name() == filter.PreventRemoveFilterName {
 					if !opts.DryRunStrategy.ClientOrServerDryRun() {
-						err := p.removeInventoryAnnotation(obj)
+						obj, err = p.removeInventoryAnnotation(obj)
 						if err != nil {
 							if klog.V(4).Enabled() {
 								klog.Errorf("error removing annotation (object: %q, annotation: %q): %v", id, inventory.OwningInventoryKey, err)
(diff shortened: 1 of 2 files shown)


pkg/apply/prune/prune.go
```diff
@@ -163,7 +163,10 @@ func (p *Pruner) Prune(
 }
 
 // removeInventoryAnnotation removes the `config.k8s.io/owning-inventory` annotation from pruneObj.
-func (p *Pruner) removeInventoryAnnotation(obj *unstructured.Unstructured) error {
+func (p *Pruner) removeInventoryAnnotation(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
```
mortent, line 166: Do we need to return the cloned resource here? Unlike in the apply case, the object we mutate here isn't actually an input to the library but something that we read from the cluster.

karlkfi: The obj comes from the Prune objs input, which comes from PruneTask.Objects. I don't think we want to modify PruneTask.Objects while the task is running.

karlkfi: Another option would be to create a PATCH directly that removes the annotation, rather than modifying the object and then making a PATCH, but that would be a bigger change to the code here.

mortent: Agree that it is probably better to avoid mutating the resources. PATCH could be a good option, but agree that should be a separate PR.

haiyanmeng:
Pleas fix the `cli-utils-presubmit-master` failure. 

Otherwise, LGTM.

karlkfi:
Tests fixed.

PR #470, opened 2021-11-04
feat: Add apply/destroy --status-events flag

karlkfi:
- Status events disabled by default
- Status events always enabled for table output

The table output has a status column which doesn't work when status events are disabled. So they either need to be enabled or a way added to remove the column. I opted for the simpler solution.

The current "enable only if timeout is set" doesn't make any sense, because it defaults to a 1m timeout and we want to make it wait forever anyway. So I just removed that bit and disabled the status events by default.


karlkfi:
Fixed tests. Needs another lgtm @haiyanmeng

karlkfi:
rebased on top of #463

PR #472, opened 2021-11-05
Consistently use InfoHelper to get rest client

ash2k:
This started as a follow up for https://github.com/kubernetes-sigs/cli-utils/pull/430#issuecomment-953455611 to pass version to the mapper to construct a correctly versioned inventory client, but then I noticed that all the required information is already present in the `resource.Info` object. After that I got some failing tests and it turned out there was some duplication of functionality between `InfoHelper` and the factory. I've replaced duplicated code with factory function invocation. I think it's a nice cleanup in the end.

### DIFF
--- a/pkg/inventory/inventory-client_test.go
+++ b/pkg/inventory/inventory-client_test.go
@@ -9,6 +9,8 @@ import (
 	"net/http"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


pkg/inventory/inventory-client_test.go
```diff
@@ -248,18 +249,18 @@ func TestCreateInventory(t *testing.T) {
...
-			if !tc.isError && err != nil {
-				t.Fatalf("unexpected error received: %s", err)
-			}
-			if tc.isError && err == nil {
-				t.Fatalf("expected error but received none")
+			err = invClient.createInventoryObj(inv, common.DryRunNone)
+			if tc.error != "" {
+				assert.EqualError(t, err, tc.error)
+			} else {
+				assert.NoError(t, err)
```
ash2k, line 263: A better test here.

seans3 (approved):
/lgtm

PR #474, opened 2021-11-05
No need to lowercase namespaces

ash2k:
They are lowercase and that's enforced by Kubernetes. We [don't lowercase](https://github.com/kubernetes-sigs/cli-utils/blob/8da78378653820d12a12b9a92103ab302e4591b9/pkg/apply/filter/local-namespaces-filter.go#L38) the namespaces' names when we perform the check anyway.


seans3 (approved):
/lgtm

PR #475, opened 2021-11-10
fix: Improve task & event logging

karlkfi:
- Log tasks at start and end
- Log events before sending
- Add String funcs for readable event logs

### DIFF
--- a/pkg/apply/taskrunner/context.go
+++ b/pkg/apply/taskrunner/context.go
@@ -53,6 +53,12 @@ func (tc *TaskContext) ResourceCache() cache.ResourceCache {
 	return tc.resourceCache
 }
 
+// SendEvent sends an event on the event channel
+func (tc *TaskContext) SendEvent(e event.Event) {
+	klog.V(5).Infof("sending event: %s", e)
+	tc.eventChannel <- e
+}
+
 // IsSuccessfulApply returns true if the object apply was successful
 func (tc *TaskContext) IsSuccessfulApply(id object.ObjMetadata) bool {
 	_, found := tc.successfulApplies[id]
--- a/pkg/apply/event/event.go
+++ b/pkg/apply/event/event.go
@@ -4,6 +4,9 @@
 package event
 
 import (
+	"fmt"
+	"strings"
+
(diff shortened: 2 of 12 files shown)


seans3 (approved):

pkg/apply/taskrunner/context.go
```diff
@@ -53,6 +53,12 @@ func (tc *TaskContext) ResourceCache() cache.ResourceCache {
 	return tc.resourceCache
 }
 
+// SendEvent sends an event on the event channel
+func (tc *TaskContext) SendEvent(e event.Event) {
+	klog.V(5).Infof("sending event: %s", e)
```
ash2k, line 58: I'd make this debug level. Events are usually consumed by printers, so there is no need to log it and then also print it.

pkg/apply/taskrunner/context.go
```diff
@@ -53,6 +53,12 @@ func (tc *TaskContext) ResourceCache() cache.ResourceCache {
 	return tc.resourceCache
 }
 
+// SendEvent sends an event on the event channel
+func (tc *TaskContext) SendEvent(e event.Event) {
```
ash2k, line 57: I like this new method, it's a good idea. I wonder if we can delete the `EventChannel()` method to encapsulate the channel even more? No need to do it in this PR, of course.

karlkfi: The EventChannel method is used in a few places to read from the event channel, like in the ApplyTask which sends it to `applyOptionsFactoryFunc` (aka `newApplyOptions`) which passes it to `KubectlPrinterAdapter`. I don't really understand why it's building a printer, but if you wanna clean it up, be my guest.

I mostly just added the SendEvent convenience method so there was a single place to log from.

pkg/apply/taskrunner/context.go
karlkfi: AFAICT cli-utils has no debug or even warning logging. It's all info and error logs with various levels of verbosity.

It's not really clear to me when to use each type with klog or even really what the levels are intended for. But if we want to standardize it should probably be a different cleanup PR.

PR #478, opened 2021-11-11
fix: Make testutil.AssertEqual more like testify

karlkfi:
- Swap actual and expected arguments
- Add optional Sprintf arguments
- Add messages to existing usages


seans3 (approved):
/lgtm

PR #479, opened 2021-11-12
fix: Make UnknownTypesError error more readable

karlkfi:
- Fix regression from switch to GroupVersionKind from GroupKind
- New error string format: G/V/K, G/V/K
- Avoid K.G/V, which looks like but isn't R.G/V

Fixes: https://github.com/kubernetes-sigs/cli-utils/pull/430#issuecomment-966737188


mortent (approved):
/approve

ash2k (approved):
/lgtm

PR #480, opened 2021-11-12
Validate the WaitTask.Timeout

haiyanmeng:


seans3 (approved):

pkg/apply/solver/solver_test.go
```diff
@@ -429,6 +429,12 @@ func TestTaskQueueBuilder_AppendApplyWaitTasks(t *testing.T) {
 							expTsk.Ids, actWaitTask.Ids)
 					}
 					assert.Equal(t, taskrunner.AllCurrent, actWaitTask.Condition)
+					if tc.options.ReconcileTimeout != 0 {
```
karlkfi, line 432: Can we just replace this entire for loop with `testutil.AssertEqual(t, tc.expectedTasks, tq.tasks)`?

pkg/apply/solver/solver_test.go
```diff
@@ -429,6 +429,12 @@ func TestTaskQueueBuilder_AppendApplyWaitTasks(t *testing.T) {
...
+						assert.Equal(t, tc.options.ReconcileTimeout, expTsk.Timeout)
+						assert.Equal(t, expTsk.Timeout, actWaitTask.Timeout)
+					} else {
+						assert.Equal(t, tc.options.ReconcileTimeout, actWaitTask.Timeout)
```
karlkfi, line 436: This else seems redundant. Isn't this always being tested?

pkg/apply/solver/solver_test.go
```diff
@@ -746,7 +752,12 @@ func TestTaskQueueBuilder_AppendPruneWaitTasks(t *testing.T) {
 					}
 					assert.Equal(t, taskrunner.AllNotFound, actWaitTask.Condition)
 					// Validate the prune wait timeout.
-					assert.Equal(t, tc.options.PruneTimeout, actualTask.(*taskrunner.WaitTask).Timeout)
+					if tc.options.ReconcileTimeout != 0 {
+						assert.Equal(t, tc.options.PruneTimeout, expTsk.Timeout)
+						assert.Equal(t, expTsk.Timeout, actWaitTask.Timeout)
+					} else {
+						assert.Equal(t, tc.options.PruneTimeout, actWaitTask.Timeout)
```
karlkfi, line 759: This else seems redundant. Isn't this always being tested?

pkg/apply/solver/solver_test.go
```diff
@@ -183,7 +183,7 @@ func TestTaskQueueBuilder_AppendApplyWaitTasks(t *testing.T) {
 						testutil.ToIdentifier(t, resources["deployment"]),
 						testutil.ToIdentifier(t, resources["secret"]),
 					},
-					taskrunner.AllCurrent, 1*time.Second,
+					taskrunner.AllCurrent, 1*time.Minute,
```
karlkfi, line 186: Does it matter what the timeout is for this test? We're just checking that it was set, right?

karlkfi:
PR title and commit text seems inaccurate. What this is actually testing is just that the Solver is passing the timeout to the WaitTask, right?

pkg/apply/solver/solver_test.go
haiyanmeng: No. If `tc.Options.PruneTimeout` is not set, `WaitTask.Timeout` will be zero. We can either test `assert.Equal(t, tc.options.PruneTimeout, actWaitTask.Timeout)` or change the `WaitTask.Timeout` in `tc.expectedTasks` from `1*time.Second` to 0.

pkg/apply/solver/solver_test.go
haiyanmeng: No. If `tc.Options.ReconcileTimeout` is not set, `WaitTask.Timeout` will be zero. We can either test `assert.Equal(t, tc.options.ReconcileTimeout, actWaitTask.Timeout)` or change the `WaitTask.Timeout` in `tc.expectedTasks` from `1*time.Second` to 0.

pkg/apply/solver/solver_test.go
haiyanmeng: Before this PR, we didn't check the Timeout field at all.

pkg/apply/solver/solver_test.go
karlkfi: What i mean is that because of the transitive property `tc.options.ReconcileTimeout == expTsk.Timeout && expTsk.Timeout == actWaitTask.Timeout` means that `tc.options.ReconcileTimeout == actWaitTask.Timeout`. So the 2nd line in the if statement is equivalent to the else statement. We could just move it out, right?

```
if tc.options.ReconcileTimeout != 0 {
    assert.Equal(t, tc.options.ReconcileTimeout, expTsk.Timeout)
}
assert.Equal(t, tc.options.ReconcileTimeout, actWaitTask.Timeout)
```

pkg/apply/solver/solver_test.go
karlkfi: This can be simplified to the following, right?
```
if tc.options.ReconcileTimeout != 0 {
    assert.Equal(t, tc.options.PruneTimeout, expTsk.Timeout)
}
assert.Equal(t, tc.options.PruneTimeout, actWaitTask.Timeout)
```

Issue #481, opened 2021-11-12
Bug: dependency chain with namespace and apply-time-mutation fails


karlkfi:
Resources:
- namespace test
- pod-a
  - implicit dep: namespace test
- pob-b
  - implicit dep: namespace test
  - explicit dep: apply-time mutation with pod-a as source

Output (from kpt using cli-utils master):
```
$ kpt live apply
namespace/test unchanged
1 resource(s) applied. 0 created, 1 unchanged, 0 configured, 0 failed
pod/pod-a created
pod/pod-b apply failed: failed to mutate "test_pod-b__Pod" with "ApplyTimeMutator": failed to read field ($.status.podIP) from source resource (/namespaces/test/Pod/pod-a): expected 1 match, but found 0)
2 resource(s) applied. 1 created, 0 unchanged, 0 configured, 1 failed
error: 1 resources failed
```

karlkfi:
Explicit dep with depends-on works, so it's just apply-time-mutation that doesn't (weird!).

```
$ kpt live apply
namespace/test unchanged
1 resource(s) applied. 0 created, 1 unchanged, 0 configured, 0 failed
pod/pod-a created
1 resource(s) applied. 1 created, 0 unchanged, 0 configured, 0 failed
pod/pod-b created
1 resource(s) applied. 1 created, 0 unchanged, 0 configured, 0 failed
```

karlkfi:
apply-time-mutation works if the source namespace is specified. So it's just the implicit namespace resolution that breaks graph sorting.

karlkfi:
Root cause:
`graph.Sort` only uses `mutation.ReadAnnotation`, which returns an empty Namespace, if not specified by the user. The implicit namespace resolution is performed in `ApplyTimeMutator`.

Unfortunately, the easy solution of moving implicit namespace resolution into `mutation.ReadAnnotation` doesn't work, because it needs to know whether the Source and Target are both namespaced. This requires using the mapper to resolve the resource schemas. If we try to use the mapper in `graph.Sort` it would fail for resources that haven't had their CRD applied yet.

So this is another issue with mapping lookups happening up front instead of lazily as-needed. Changing that, to resolve sort order after every apply, would be a significant change...

Also unfortunate is that we can't reject SourceRefs without an explicit namespace without knowing if the resource is namespaced or not...

karlkfi:
Possible workarounds:
1. Resolve mapping up front, if possible, and error at apply time, if not
[...]

PR #482, opened 2021-11-15
[WIP] fix: Require src namespace in apply-time-mutation

karlkfi:
- This change dissables implicit namespace resolution for
  apply-time-mutation sourceRefs. This feature has been partially
  broken since it was implimented and fixing it is not currently
  possible. The sorting never worked properly because the namespace
  was being infered at apply time, while sorting was being done
  much earlier, before the tasks were scheduled.
- Unfortunately, this makes testing and usage harder when the
  namespace is not known in advance, requiring yaml templating.
- Avoid logging yaml at log lvl 5 (use lvl 7)
- Improve error messages when annotation read/write fails
- Warn if dependencies cannot be sorted
- Warn if dependency not in resource set

Fixes: https://github.com/kubernetes-sigs/cli-utils/issues/481

### DIFF
--- a/pkg/apply/mutator/apply_time_mutator.go
+++ b/pkg/apply/mutator/apply_time_mutator.go
@@ -57,7 +57,7 @@ func (atm *ApplyTimeMutator) Mutate(ctx context.Context, obj *unstructured.Unstr
 		return mutated, reason, fmt.Errorf("failed to read annotation in resource (%s): %w", targetRef, err)
 	}
 
-	klog.V(4).Infof("target resource (%s):\n%s", targetRef, object.YamlStringer{O: obj})
+	klog.V(7).Infof("target resource (%s):\n%s", targetRef, object.YamlStringer{O: obj})
 
 	// validate no self-references
 	// Early validation to avoid GETs, but won't catch sources with implicit namespace.
@@ -78,7 +78,7 @@ func (atm *ApplyTimeMutator) Mutate(ctx context.Context, obj *unstructured.Unstr
 
 		// Default source namespace to target namesapce, if namespace-scoped
 		if sourceRef.Namespace == "" && sourceMapping.Scope.Name() == meta.RESTScopeNameNamespace {
-			sourceRef.Namespace = targetRef.Namespace
+			return mutated, reason, fmt.Errorf("invalid source reference (%s): missing required namespace", sourceRef)
 		}
 
 		// validate no self-references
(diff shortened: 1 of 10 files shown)


pkg/object/graph/depends.go
```diff
@@ -83,20 +83,27 @@ func ReverseSortObjs(objs object.UnstructuredSet) ([]object.UnstructuredSet, err
...
+	for i, obj := range objs {
+		id := ids[i]
 		klog.V(3).Infof("adding vertex: %s", id)
 		g.AddVertex(id)
 		if mutation.HasAnnotation(obj) {
 			subs, err := mutation.ReadAnnotation(obj)
 			if err != nil {
-				// TODO: fail task if parse errors?
-				klog.V(3).Infof("failed to add edges from: %s: %s", id, err)
+				klog.Warningf("warning: failed to sort explicit dependency: %v",
```
karlkfi, line 94: `klog.Warning` prints directly to the user when using kapply or kpt, but for server-side usage it would just print to the service log. 

Should we make a new WarningEvent instead, so the printer can handle it?

test/e2e/mutation_test.go
```diff
@@ -38,7 +39,28 @@ func mutationTest(ctx context.Context, c client.Client, invConfig InventoryConfi
 
 	inv := invConfig.InvWrapperFunc(invConfig.InventoryFactoryFunc(inventoryName, namespaceName, "test"))
 
-	podAObj := withNamespace(manifestToUnstructured(podA), namespaceName)
+	podAObj := withMutation(withNamespace(manifestToUnstructured(podA), namespaceName), mutation.ApplyTimeMutation{
```
karlkfi, line 42: This test is a little too dependent on the impl. It might be better to use go-template to template the input yaml.

pkg/apply/mutator/apply_time_mutator.go
```diff
@@ -78,7 +78,7 @@ func (atm *ApplyTimeMutator) Mutate(ctx context.Context, obj *unstructured.Unstr
 
 		// Default source namespace to target namesapce, if namespace-scoped
```
mortent, line 79: We should probably fix the comment here.

pkg/object/graph/depends.go
mortent: The `Warning` event would be a new top-level event type then? It seems like it would be only for the user to understand what is going on, as it seems difficult to use it for anything useful from code? Another option (perhaps in addition to a `Warning` event) would be to allow users to specify if invalid references should lead to an error or not. Although that would require that we can fit the behavior into the continue-on-error strategy, but this happens before we apply any of the resources.

mortent: So I think the question here is what should happen if either the annotation can't be read or it doesn't specify a namespace in a reference to a namespace-scoped resource. If we print an error or send an error event, then presumably we will go on with applying the resource right? I think considering the resource as failed/skipped in this situation might be more appropriate, and we can provide information back to the caller through the `Error` property on the `ApplyEvent`. It seems that applying the resource without doing the apply-time mutation almost certainly will result in a different outcome than the user expected. And by just considering that particular resource as failed/skipped, we will still continue applying other resources.

karlkfi: I've defined the problem and a couple proposals here: https://github.com/kubernetes-sigs/cli-utils/issues/484

Issue #421, opened 2021-10-07
Bug: apply sets that include both CRD and CR fail when not using a caching RestMapper


ash2k:
When the set of applied resources contains a CRD and a CR of that new kind, applier seem to get stuck in a loop, re-trying. I think it's [waiting for the rest mapper to be updated with the new mapping](https://github.com/kubernetes-sigs/cli-utils/blob/231f3f4a699c56fe75aa65b1535b7e9ca3996704/pkg/kstatus/polling/clusterreader/caching_reader.go#L221-L232) from the CRD, but that never happens because nothing invalidates the rest mapper.

I wonder if it's a bug/missing feature or I'm just not using the library correctly?

FYI this is related to https://github.com/kubernetes/kubernetes/issues/64391

ash2k:
I've opened a PR to introduce a proper API for this https://github.com/kubernetes/kubernetes/pull/105623.

karlkfi:
FWIW, I think @justinsb has some code with a lazily populated caching RESTMapper that allows individual schema invalidation, rather than all or nothing rest. It also only grabs a schema as needed, instead of all at once (which currently triggers throttling).

We might want to use something like that instead of the existing CachingRESTClientGetter.

PR #485, opened 2021-11-20
Applier constructor and run: builder

ash2k:
This is an alternative approach to https://github.com/kubernetes-sigs/cli-utils/pull/454, as suggested in https://github.com/kubernetes-sigs/cli-utils/pull/454#issuecomment-965812739.

/hold
/kind cleanup

### DIFF
--- a/cmd/apply/cmdapply.go
+++ b/cmd/apply/cmdapply.go
@@ -141,7 +141,10 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
 
 	// Run the applier. It will return a channel where we can receive updates
 	// to keep track of progress and any issues.
-	a, err := apply.NewApplier(r.factory, invClient)
+	a, err := apply.NewApplierBuilder().
+		WithFactory(r.factory).
+		WithInventoryClient(invClient).
+		Build()
 	if err != nil {
 		return err
 	}
@@ -151,7 +154,7 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
 		r.printStatusEvents = true
 	}
 
-	ch := a.Run(ctx, inv, objs, apply.Options{
+	ch := a.Run(ctx, inv, objs, apply.ApplierOptions{
 		ServerSideOptions: r.serverSideOptions,
 		PollInterval:      r.period,
 		ReconcileTimeout:  r.reconcileTimeout,
--- a/pkg/apply/applier.go
+++ b/pkg/apply/applier.go
@@ -9,10 +9,12 @@ import (
 	"sort"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/klog/v2"
-	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"sigs.k8s.io/cli-utils/pkg/apply/cache"
 	"sigs.k8s.io/cli-utils/pkg/apply/event"
 	"sigs.k8s.io/cli-utils/pkg/apply/filter"
@@ -24,35 +26,12 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/apply/taskrunner"
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/cli-utils/pkg/inventory"
-	"sigs.k8s.io/cli-utils/pkg/kstatus/polling"
-	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/engine"
 	"sigs.k8s.io/cli-utils/pkg/object"
 	"sigs.k8s.io/cli-utils/pkg/object/validation"
 	"sigs.k8s.io/cli-utils/pkg/ordering"
 )
 
-// NewApplier returns a new Applier.
-func NewApplier(factory cmdutil.Factory, invClient inventory.InventoryClient) (*Applier, error) {
-	pruner, err := prune.NewPruner(factory, invClient)
-	if err != nil {
-		return nil, err
-	}
-	statusPoller, err := polling.NewStatusPollerFromFactory(factory, []engine.StatusReader{})
-	if err != nil {
-		return nil, err
-	}
-	mapper, err := factory.ToRESTMapper()
-	if err != nil {
-		return nil, err
-	}
-	return &Applier{
-		pruner:       pruner,
-		StatusPoller: statusPoller,
-		factory:      factory,
-		invClient:    invClient,
-		infoHelper:   info.NewInfoHelper(mapper, factory),
-	}, nil
-}
+const defaultPollInterval = 2 * time.Second
 
 // Applier performs the step of applying a set of resources into a cluster,
 // conditionally waits for all of them to be fully reconciled and finally
@@ -65,17 +44,19 @@ func NewApplier(factory cmdutil.Factory, invClient inventory.InventoryClient) (*
 // parameters and/or the set of resources that needs to be applied to the
 // cluster, different sets of tasks might be needed.
 type Applier struct {
-	pruner       *prune.Pruner
-	StatusPoller poller.Poller
-	factory      cmdutil.Factory
-	invClient    inventory.InventoryClient
-	infoHelper   info.InfoHelper
+	pruner        *prune.Pruner
+	statusPoller  poller.Poller
+	invClient     inventory.InventoryClient
+	client        dynamic.Interface
+	openAPIGetter discovery.OpenAPISchemaInterface
+	mapper        meta.RESTMapper
+	infoHelper    info.InfoHelper
 }
 
 // prepareObjects returns the set of objects to apply and to prune or
 // an error if one occurred.
 func (a *Applier) prepareObjects(localInv inventory.InventoryInfo, localObjs object.UnstructuredSet,
-	o Options) (object.UnstructuredSet, object.UnstructuredSet, error) {
+	o ApplierOptions) (object.UnstructuredSet, object.UnstructuredSet, error) {
 	if localInv == nil {
 		return nil, nil, fmt.Errorf("the local inventory can't be nil")
 	}
@@ -117,37 +98,25 @@ func (a *Applier) prepareObjects(localInv inventory.InventoryInfo, localObjs obj
 }
 
 // Run performs the Apply step. This happens asynchronously with updates
-// on progress and any errors are reported back on the event channel.
+// on progress and any errors reported back on the event channel.
 // Cancelling the operation or setting timeout on how long to Wait
 // for it complete can be done with the passed in context.
 // Note: There isn't currently any way to interrupt the operation
 // before all the given resources have been applied to the cluster. Any
 // cancellation or timeout will only affect how long we Wait for the
 // resources to become current.
-func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, objects object.UnstructuredSet, options Options) <-chan event.Event {
+func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, objects object.UnstructuredSet, options ApplierOptions) <-chan event.Event {
 	klog.V(4).Infof("apply run for %d objects", len(objects))
 	eventChannel := make(chan event.Event)
 	setDefaults(&options)
 	go func() {
 		defer close(eventChannel)
-
-		client, err := a.factory.DynamicClient()
-		if err != nil {
-			handleError(eventChannel, err)
-			return
-		}
-		mapper, err := a.factory.ToRESTMapper()
-		if err != nil {
-			handleError(eventChannel, err)
-			return
-		}
-
 		// Validate the resources to make sure we catch those problems early
 		// before anything has been updated in the cluster.
 		vCollector := &validation.Collector{}
 		validator := &validation.Validator{
 			Collector: vCollector,
-			Mapper:    mapper,
+			Mapper:    a.mapper,
 		}
 		validator.Validate(objects)
 
@@ -162,13 +131,14 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
 		// Fetch the queue (channel) of tasks that should be executed.
 		klog.V(4).Infoln("applier building task queue...")
 		taskBuilder := &solver.TaskQueueBuilder{
-			Pruner:     a.pruner,
-			Factory:    a.factory,
-			InfoHelper: a.infoHelper,
-			Mapper:     mapper,
-			InvClient:  a.invClient,
-			Destroy:    false,
-			Collector:  vCollector,
+			Pruner:        a.pruner,
+			DynamicClient: a.client,
+			OpenAPIGetter: a.openAPIGetter,
+			InfoHelper:    a.infoHelper,
+			Mapper:        a.mapper,
+			InvClient:     a.invClient,
+			Destroy:       false,
+			Collector:     vCollector,
 		}
 		opts := solver.Options{
 			ServerSideOptions:      options.ServerSideOptions,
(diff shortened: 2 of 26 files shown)


pkg/apply/applier.go
```diff
@@ -114,192 +92,138 @@ func (a *Applier) prepareObjects(localInv inventory.InventoryInfo, localObjs obj
...
-		o.PrunePropagationPolicy = metav1.DeletePropagationBackground
+	if err != nil {
+		postErrorEvent(options, err)
+		return
 	}
 }
 
-func handleError(eventChannel chan event.Event, err error) {
-	eventChannel <- event.Event{
+func postErrorEvent(cfg *Options, err error) {
```
karlkfi, line 215: i think we use "send" as the verb for events elsewhere

pkg/apply/applier_apply_options.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// inventoryPolicy defines the inventory policy of apply.
+	inventoryPolicy inventory.InventoryPolicy
+
+	// eventListeners is a list of event listeners that are called in specified order when an event happens.
+	// Listeners are never called concurrently with each other.
+	eventListeners []func(event.Event)
+}
+
+func NewOptions() *Options {
```
karlkfi, line 60: Would `NewDefaultOptions` be a better name? or is that implied? I'm on the fence.

pkg/apply/applier_apply_options.go
```diff
@@ -0,0 +1,141 @@
...
+	return &Options{
+		pollInterval:           defaultPollInterval,
+		prune:                  true,
+		prunePropagationPolicy: metav1.DeletePropagationBackground,
+	}
+}
+
+func (o *Options) ServerSideOptions(serverSideOptions common.ServerSideOptions) *Options {
```
karlkfi, line 68: Why did you decide not to use the `With` prefix here?

pkg/apply/applier_builder.go
```diff
@@ -0,0 +1,160 @@
...
+	client                       dynamic.Interface
+	discoClient                  discovery.CachedDiscoveryInterface
+	mapper                       meta.RESTMapper
+	restConfig                   *rest.Config
+	unstructuredClientForMapping func(*meta.RESTMapping) (resource.RESTClient, error)
+	statusPoller                 poller.Poller
+}
+
+// NewBuilder returns a new Applier builder.
+func NewBuilder() *Builder {
```
karlkfi, line 38: Since we probably need this for the Destroyer in the same package, this should maybe be `NewApplierBuilder`.

pkg/apply/applier_builder.go
```diff
@@ -0,0 +1,160 @@
...
+	return &Builder{
+		// Defaults, if any, go here.
+	}
+}
+
+func (b *Builder) New() (*Applier, error) {
```
karlkfi, line 44: I think `Build()` is the conventional terminus when using the builder pattern.

pkg/apply/applier_apply_options.go
```diff
@@ -0,0 +1,141 @@
...
+	o.pruneTimeout = pruneTimeout
+	return o
+}
+
+func (o *Options) InventoryPolicy(inventoryPolicy inventory.InventoryPolicy) *Options {
+	o.inventoryPolicy = inventoryPolicy
+	return o
+}
+
+func (o *Options) EventListener(eventListener ...func(event.Event)) *Options {
```
karlkfi, line 113: These event channel changes should probably go in a separate PR. They seem enabled by but not required for the Builder pattern. If you leave the applier.Run the way it was, it should significantly simplify this PR.

I'm not sure how I feel about these EventChannel changes, but it doesn't need to block the other changes.

karlkfi (changes requested):
- I like the way the apply options were changed here.
- I mostly like the applier builder too, with some changes.
- I'm on the fence on whether the event channel changes are an improvement. They can probably be punted to a subsequent PR.
- I made a few inline comments about conventional naming

pkg/apply/applier_apply_options.go
```diff
@@ -0,0 +1,141 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/cli-utils/pkg/apply/event"
+	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
+)
+
+const defaultPollInterval = 2 * time.Second
+
+type Options struct {
```
karlkfi, line 17: Given that we're changing how to build this, we might want to rename it to `apply.ApplierOptions` too, to be consistent with the `DestroyerOptions`. I think the applier options came first before the destroyer was added.

Also, do we want to make this a private struct so it has to be constructed with the builder?

pkg/apply/applier_apply_options.go
ash2k: `With` was a way to namespace functions when they all were in the same "package name namespace". Not that they are methods, perhaps it's better to drop `With` for brevity?

pkg/apply/applier_apply_options.go
ash2k: Makes sense, renamed and made private.

ash2k: Hm, actually, if we make it private it will be difficult to use mocks with it. Mocking library will not be able to refer to the type.

pkg/apply/applier_apply_options.go
ash2k: I think it should be `NewApplierOptions` if we are changing the type name to `ApplierOptions`.

ash2k:
@karlkfi I've made most of the changes you suggested, thanks for the review.

> I'm on the fence on whether the event channel changes are an improvement. They can probably be punted to a subsequent PR.

I'd like to keep this change:
- It allows to create reusable and composable consumers out of the box, without having to build fan-out on top of the event channel. For example, I'd like to refactor printers to be usable as callbacks:
  ```go
  type Printer interface {
          Print(Event)
  }
  func (o *ApplierOptions) PrintEvents(p Printer) *ApplierOptions {
	return o.EventListener(func(e event.Event) {
		p.Print(e)
	})
  }
  ```
- It allows library user to decide whether they want concurrency or not. As seen in tests, quite often consumer blocks while apply operation is running (no concurrency -> simpler API).

cmd/apply/cmdapply.go
```diff
@@ -137,7 +138,10 @@ func (r *ApplyRunner) RunE(cmd *cobra.Command, args []string) error {
 
 	// Run the applier. It will return a channel where we can receive updates
 	// to keep track of progress and any issues.
-	a, err := apply.NewApplier(r.factory, invClient)
+	a, err := apply.NewApplierBuilder().
+		WithFactory(r.factory).
```
seans3, line 142: Should this optional parameter be required?

pkg/apply/applier_builder.go
```diff
@@ -0,0 +1,160 @@
...
+	"sigs.k8s.io/cli-utils/pkg/apply/poller"
+	"sigs.k8s.io/cli-utils/pkg/apply/prune"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/polling"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ApplierBuilder struct {
+	// factory is only used to retrieve things that have not been provided explicitly.
+	factory                      util.Factory
```
mortent, line 27: Do we need all these options? I'm not even sure which ones are mutually exclusive. I think some of it we might be able to clean up by simplifying things internally. I'm also uncertain if we should let users provide things like the RESTMapper, since we in practice have some requirements on the implementation that isn't obvious.

pkg/apply/applier_apply_options.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// CollectEventsInto is a convenience helper that attaches a listener that appends events to the eventSink slice.
+func (o *ApplierOptions) CollectEventsInto(eventSink *[]event.Event) *ApplierOptions {
+	return o.EventListener(func(e event.Event) {
+		*eventSink = append(*eventSink, e)
+	})
+}
+
+// CollectErrorInto is a convenience helper that captures error from the first event of ErrorType type.
```
mortent, line 132: As far as I know, there should only ever be one Error event.

pkg/apply/applier_apply_options.go
```diff
@@ -0,0 +1,141 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/cli-utils/pkg/apply/event"
+	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
+)
+
+const defaultPollInterval = 2 * time.Second
+
+type ApplierOptions struct {
```
mortent, line 17: Do we need to change the name to ApplierOptions? It seems a bit verbose given that it is in the apply package.

pkg/apply/applier_apply_options.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// inventoryPolicy defines the inventory policy of apply.
+	inventoryPolicy inventory.InventoryPolicy
+
+	// eventListeners is a list of event listeners that are called in specified order when an event happens.
+	// Listeners are never called concurrently with each other.
+	eventListeners []func(event.Event)
+}
+
+func NewApplierOptions() *ApplierOptions {
```
mortent, line 60: Why would we have a pointer to the ApplierOptions? With the ApplierOptions just holding data, it seems unnecessary.

pkg/apply/applier.go
```diff
@@ -114,192 +92,138 @@ func (a *Applier) prepareObjects(localInv inventory.InventoryInfo, localObjs obj
...
+// Apply performs the apply step. This happens synchronously with updates
+// on progress and any errors reported back via callbacks.
 // Cancelling the operation or setting timeout on how long to Wait
 // for it complete can be done with the passed in context.
 // Note: There isn't currently any way to interrupt the operation
 // before all the given resources have been applied to the cluster. Any
 // cancellation or timeout will only affect how long we Wait for the
 // resources to become current.
-func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, objects object.UnstructuredSet, options Options) <-chan event.Event {
+func (a *Applier) Apply(ctx context.Context, invInfo inventory.InventoryInfo, objects object.UnstructuredSet, options *ApplierOptions) {
```
mortent, line 103: This makes for a somewhat strange API. If it is synchronous, I would expect the results (at least any errors) be return values rather than having to fetch it through the `ApplierOptions`.

Returning a channel does come with some complexity, but it is a pretty standard golang pattern. And it is pretty easy to write simple wrapper functions to make it do everything that the the approach in this PR allows for. 
I'm not convinced passing pointers to structs and errors into functions on the `ApplierOptions` is any better. Examples [here](https://github.com/kubernetes-sigs/cli-utils/pull/485/files#diff-35e39add2210ba4befff203297cc91784fc697568de88aca4194f4ca2f460b81R30) and [here](https://github.com/kubernetes-sigs/cli-utils/pull/485/files#diff-0367d69a6e70baf91f515166e4e2cc6c16b065323518fd37388e5c774a23a5f9R170).

mortent (changes requested):
This PR seems to combine several changes that are mostly unrelated. It seems like it could be split up into separate PRs.

pkg/apply/applier_apply_options.go
mortent: I didn't see the discussion around this until now. I'm ok with the change.

pkg/apply/applier.go
mortent: If we go down this path, it seems like we could let the `Apply` function return an error. The current `ErrorEvent` is considered fatal and should be the last event sent before the channel is closed. Then users don't have to worry about collecting results from the `ApplyOptions` unless they are interested in the events.
I thought I saw this mentioned in https://github.com/kubernetes-sigs/cli-utils/pull/454, but now I can't find it.

cmd/apply/cmdapply.go
ash2k: It is not required if all other things, that are derived from the factory, are provided explicitly. This approach allows user to customize what applier uses without creating their own `Factory` implementation (to inject desired objects).

pkg/apply/applier_builder.go
ash2k: I think it's good to allow users to inject dependencies directly. For those who don't need to do that, they can inject just the factory and nothing else. My main motivator is the desire to inject a custom `rest.Config`/`dynamic.Interface` with a dedicated rate limiter. Other parts of my program should use a separate rate limiter from the applier. I could do it by creating a custom factory, but that is quite unwieldy. It's an example use case, anyway.

> I'm also uncertain if we should let users provide things like the RESTMapper, since we in practice have some requirements on the implementation that isn't obvious.

If there is a requirement, ideally code should fail early if it's not met rather than silently not do what it is supposed to (current behavior). I'm going to fix that in https://github.com/kubernetes-sigs/cli-utils/issues/421 by using code from https://github.com/kubernetes/kubernetes/pull/105623. As part of that I'm going to change `meta.RESTMapper` to `meta.ResettableRESTMapper` here so the requirement is explicit.

pkg/apply/applier.go
```diff
@@ -245,7 +215,7 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
 	return eventChannel
 }
 
-type Options struct {
+type ApplierOptions struct {
```
ash2k, line 218: We have `DestroyerOptions`, so renamed this for consistency as suggested here https://github.com/kubernetes-sigs/cli-utils/pull/485#discussion_r755460760.

pkg/apply/applier.go
```diff
@@ -286,9 +256,9 @@ type Options struct {
 
 // setDefaults set the options to the default values if they
 // have not been provided.
-func setDefaults(o *Options) {
-	if o.PollInterval == time.Duration(0) {
-		o.PollInterval = poller.DefaultPollInterval
+func setDefaults(o *ApplierOptions) {
+	if o.PollInterval == 0 {
+		o.PollInterval = defaultPollInterval
```
ash2k, line 261: `poller.DefaultPollInterval` is not used anywhere else, so brought it here, made private and dropped the dependency on the `poller` package.

ash2k:
Thank you for feedback, everybody. I've just pushed an updated version. Removed the channel->callback change and brought back the options struct for the `Run` method. The only major change in this PR is builder for construction of the applier. Later we'd want to do the same for the destroyer, I suppose.

pkg/apply/applier_builder.go
```diff
@@ -0,0 +1,160 @@
...
+		statusPoller:  bx.statusPoller,
+		invClient:     bx.invClient,
+		client:        bx.client,
+		openAPIGetter: bx.discoClient,
+		mapper:        bx.mapper,
+		infoHelper:    info.NewInfoHelper(bx.mapper, bx.unstructuredClientForMapping),
+	}, nil
+}
+
+func (b *ApplierBuilder) finalize() (*ApplierBuilder, error) {
```
karlkfi, line 64: Do these all need to be finalized in the same function? Could we make a getter function for each option and just use those in the Build function to finalize each field individually?

karlkfi (commented):
Everything looks good except maybe the finalize function. LMK if the suggested approach sounds feasible.

karlkfi (changes requested):
Please amend the commit messages to have fix or feat prefixes. I want to try to start following https://www.conventionalcommits.org/en/v1.0.0/#summary to make change logs easier to produce. Since this is a breaking change, add a BREAKING CHANGE line at the end with detail for each external facing API change.

pkg/apply/applier_builder.go
ash2k: It can certainly be changed to work this way. I personally prefer the current style because I'm not sure what you are suggesting will look clearer in this particular case. I also like how you can put the cursor on a field and see where it's initialized and where it's used within a single method (to check that a field has been initialized before use).

If you don't feel strongly about it, I suggest we keep it this way.

karlkfi (approved):
/lgtm

pkg/apply/applier_builder.go
ash2k: We are making a copy on the next line so that the builder is reusable. I thought you were talking about improving readability of the code.

ash2k:
Rebased - no changes, just a conflict. PTAL @karlkfi @mortent

pkg/apply/applier_builder.go
karlkfi: cool cool. i was also talking about readability, but that's not a deal breaker.

karlkfi (approved):
/lgtm

ash2k:
Rebased once again.

cmd/apply/cmdapply.go
mortent: Yeah, allowing users to provide the different dependencies directly instead of through the `Factory` is quite useful.

mortent (approved):
/approve

mortent:
@karlkfi Can you lgtm this again so we can get it merged?

PR #486, opened 2021-11-21
Avoid waiting for failed resources

mortent:
This updates the WaitTask to handle resources that are reported as `Failed`. It prevents the library from waiting the full length of the wait timeout for resources that we know are unlikely to successfully reconcile.

The behavior is:
* If a resource is reported as `Failed`, the library will remove it from the set of resources we are waiting to be reconciled, add it to a set of failed resources, and send a `WaitEvent` with the `ReconciledFailed` operation.
* Any failed resources will not prevent a WaitTask from being completed if the other resources in the task have become `Current`. So a resource becoming `Failed` does not terminate a WaitTask, it just makes sure we don't end up waiting for the timeout to expire when we know resources are unlikely to successfully reconcile.
* If a resource becomes `Failed`, but happens to become unblocked before all other resources in the task has completed, it can transition out of the `Failed` state. It can move either into `Current` or `InProgress`.

### DIFF
--- a/pkg/apply/event/event.go
+++ b/pkg/apply/event/event.go
@@ -160,6 +160,7 @@ const (
 	Reconciled
 	ReconcileSkipped // Skipped
 	ReconcileTimeout // Timeout
+	ReconcileFailed  // Failed
 )
 
 type WaitEvent struct {
--- a/pkg/apply/taskrunner/task.go
+++ b/pkg/apply/taskrunner/task.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/client-go/restmapper"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/cli-utils/pkg/apply/event"
+	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/cli-utils/pkg/object"
 )
 
@@ -69,6 +70,9 @@ type WaitTask struct {
 	cancelFunc context.CancelFunc
 	// pending is the set of resources that we are still waiting for.
 	pending object.ObjMetadataSet
+	// failed is the set of resources that we are waiting for, but is considered
+	// failed, i.e. unlikely to successfully reconcile.
+	failed object.ObjMetadataSet
 	// mu protects the pending ObjMetadataSet
 	mu sync.RWMutex
 }
@@ -198,6 +202,12 @@ func (w *WaitTask) skipped(taskContext *TaskContext, id object.ObjMetadata) bool
 	return false
 }
 
+// failedByID returns true if the resource is failed.
+func (w *WaitTask) failedByID(taskContext *TaskContext, id object.ObjMetadata) bool {
+	cached := taskContext.ResourceCache().Get(id)
+	return cached.Status == status.FailedStatus
+}
+
 // Cancel exits early with a timeout error
 func (w *WaitTask) Cancel(_ *TaskContext) {
 	w.cancelFunc()
@@ -217,12 +227,18 @@ func (w *WaitTask) StatusUpdate(taskContext *TaskContext, id object.ObjMetadata)
 
 	switch {
 	case w.pending.Contains(id):
+		switch {
 		// pending - check if reconciled
-		if w.reconciledByID(taskContext, id) {
+		case w.reconciledByID(taskContext, id):
 			// reconciled - remove from pending & send event
 			w.pending = w.pending.Remove(id)
 			w.sendEvent(taskContext, id, event.Reconciled)
-		} else {
+		case w.failedByID(taskContext, id):
+			// failed - remove from pending & send event
+			w.pending = w.pending.Remove(id)
+			w.failed = append(w.failed, id)
+			w.sendEvent(taskContext, id, event.ReconcileFailed)
+		default:
 			// can't be all reconciled now, so don't bother checking
 			return
 		}
@@ -231,6 +247,24 @@ func (w *WaitTask) StatusUpdate(taskContext *TaskContext, id object.ObjMetadata)
 		return
 	case w.skipped(taskContext, id):
 		// skipped - ignore
+		return
+	case w.failed.Contains(id):
+		// If a failed resource becomes current before other
+		// resources have completed/timed out, we consider it
+		// current.
+		if w.reconciledByID(taskContext, id) {
+			w.failed = w.failed.Remove(id)
+			w.sendEvent(taskContext, id, event.Reconciled)
+		} else if !w.failedByID(taskContext, id) {
+			// If a resource is no longer reported as Failed and is not Reconciled,
+			// they should just go back to InProgress.
+			w.failed = w.failed.Remove(id)
+			w.pending = append(w.pending, id)
+			w.sendEvent(taskContext, id, event.ReconcilePending)
+		}
+		// can't be all reconciled, so don't bother checking. A failed
+		// resource doesn't prevent a WaitTask from completing, so there
+		// must be at least one InProgress resource we are still waiting for.
 		return
 	default:
 		// reconciled - check if unreconciled
@@ -243,10 +277,10 @@ func (w *WaitTask) StatusUpdate(taskContext *TaskContext, id object.ObjMetadata)
 		}
 	}
 
-	// if all conditions are met, complete the wait task
-	if conditionMet(taskContext, w.pending, w.Condition) {
-		// all reconciled - clear pending and exit
-		w.pending = object.ObjMetadataSet{}
+	// If we no longer have any pending resources, the WaitTask
+	// can be completed.
+	if len(w.pending) == 0 {
+		// all reconciled, so exit
 		klog.V(3).Infof("all objects reconciled or skipped (name: %q)", w.name)
 		w.cancelFunc()
 	}
--- a/pkg/apply/taskrunner/task_test.go
+++ b/pkg/apply/taskrunner/task_test.go
@@ -548,3 +548,343 @@ loop:
 	}
 	assert.Equal(t, expectedResults, receivedResults)
 }
+
+func TestWaitTask_Failed(t *testing.T) {
+	taskName := "wait-1"
+	testDeployment1ID := testutil.ToIdentifier(t, testDeployment1YAML)
+	testDeployment1 := testutil.Unstructured(t, testDeployment1YAML)
+	testDeployment2ID := testutil.ToIdentifier(t, testDeployment2YAML)
+	testDeployment2 := testutil.Unstructured(t, testDeployment2YAML)
+
+	testCases := map[string]struct {
+		configureTaskContextFunc func(taskContext *TaskContext)
+		eventsFunc               func(*cache.ResourceCacheMap, *WaitTask, *TaskContext)
+		waitTimeout              time.Duration
+		expectedEvents           []event.Event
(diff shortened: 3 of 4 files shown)


pkg/apply/taskrunner/task.go
```diff
@@ -178,6 +183,16 @@ func (w *WaitTask) sendTimeoutEvents(taskContext *TaskContext) {
 	}
 }
 
+// updateFailedReconciles updates the taskContext with resources that
+// are either stalled or still pending.
+func (w *WaitTask) updateFailedReconciles(taskContext *TaskContext) {
+	w.mu.RLock()
+	defer w.mu.RUnlock()
+
+	taskContext.AddFailedReconciles(w.pending)
```
karlkfi, line 192: Why add pending as failed? 

sendTimeoutEvents sends these as ReconcileTimeout, which is not the same as Failed.

pkg/apply/taskrunner/task.go
```diff
@@ -218,10 +239,15 @@ func (w *WaitTask) StatusUpdate(taskContext *TaskContext, id object.ObjMetadata)
 	switch {
 	case w.pending.Contains(id):
 		// pending - check if reconciled
-		if w.reconciledByID(taskContext, id) {
+		if w.reconciledByID(taskContext, id) { //nolint:gocritic
```
karlkfi, line 242: just use a switch statement

pkg/apply/taskrunner/task.go
```diff
@@ -218,10 +239,15 @@ func (w *WaitTask) StatusUpdate(taskContext *TaskContext, id object.ObjMetadata)
...
 			// reconciled - remove from pending & send event
 			w.pending = w.pending.Remove(id)
 			w.sendEvent(taskContext, id, event.Reconciled)
+		} else if w.resourceStalled(taskContext, id) {
+			// stalled - remove from pending & send event
+			w.pending = w.pending.Remove(id)
+			w.stalled = w.stalled.Union(object.ObjMetadataSet{id})
```
karlkfi, line 249: you don't need to union here. you can just append.

pkg/apply/taskrunner/task.go
```diff
@@ -232,6 +258,16 @@ func (w *WaitTask) StatusUpdate(taskContext *TaskContext, id object.ObjMetadata)
...
 		return
+	case w.stalled.Contains(id):
+		// If a stalled resource becomes current before other
+		// resources have completed/timed out, we consider it
+		// current.
+		if w.reconciledByID(taskContext, id) {
+			w.stalled = w.stalled.Remove(id)
+			w.sendEvent(taskContext, id, event.Reconciled)
+		}
+		// can't be all reconciled, so don't bother checking
```
karlkfi, line 269: why can't they be all reconciled here?

pkg/apply/taskrunner/task.go
```diff
@@ -244,6 +280,8 @@ func (w *WaitTask) StatusUpdate(taskContext *TaskContext, id object.ObjMetadata)
 	}
 
 	// if all conditions are met, complete the wait task
+	// TODO(mortent): Isn't it enough to just check the size of the
```
karlkfi, line 283: You're probably right, assuming we received a StatusUpdate for every object, then we could just check if `len(pending) == 0`

pkg/apply/taskrunner/task_test.go
```diff
@@ -127,7 +127,7 @@ func TestWaitTask_CompleteEventually(t *testing.T) {
 	}()
 
 	// wait for task result
-	timer := time.NewTimer(5 * time.Second)
+	timer := time.NewTimer(500 * time.Second)
```
karlkfi, line 130: Why make the timeout so big? Was it failing?
This is per test...

pkg/apply/event/event.go
```diff
@@ -160,6 +160,7 @@ const (
 	Reconciled
 	ReconcileSkipped // Skipped
 	ReconcileTimeout // Timeout
+	ReconcileStalled // Stalled
```
karlkfi, line 163: I'm not convinced we really want to introduce "Stalled" here when kapply is returning "Failed".

karlkfi (changes requested):
Thanks for taking a stab at this.

I'm not convinced we should conflate Timeout and Failed as Stalled tho. I would prefer to keep them separate.

pkg/apply/taskrunner/task.go
mortent: So the idea here is that we want to capture the resources that failed to reconcile in the TaskContext. I was thinking that once a WaitTask completes, we don't really care whether a resource failed to reconcile due to a timeout or due to the controller reporting it as Failed/Stalled. In both cases, we have a resource where the controller wasn't able to reconcile it.
I agree though that for reporting this to users, it makes sense to distinguish between them. But for how we should handle the resource in later tasks, I don't think it matters.

pkg/apply/event/event.go
mortent: I was hoping we could rename the "Failed" status from kstatus to "Stalled". It doesn't really change the behavior in any way, I just feel that "Failed" signals a terminal state, while "Stalled" better conveys that while we don't think reconciliation will succeed, it might happen as the underlying controllers will keep retrying.
Stalled is also more aligned with the guidance for Conditions.

pkg/apply/taskrunner/task.go
mortent: Done. Open to ways we can simplify this to avoid nested switch statements, but it isn't too bad.

pkg/apply/taskrunner/task.go
mortent: Updated the comment with an explanation.

pkg/apply/taskrunner/task_test.go
mortent: Just a leftover from when I was trying to debug something. I have reverted the change.

pkg/apply/taskrunner/task.go
mortent: Can we assume we get a StatusUpdate message for every object? I assume the case where this might not hold would be resources that reconcile before the WaitTask is started?

pkg/apply/taskrunner/task.go
karlkfi: you could extract the inner switch to a function, but only if it makes it easier to read with a good name.

pkg/apply/taskrunner/task.go
karlkfi: I guess the better question is: Why are we storing these in the task context if nothing is using `FailedReconciles()`?
Without usage, it's hard to know what should be in the set...

pkg/apply/taskrunner/task.go
karlkfi: If it reconciled before the WaitTask started, we won't get a StatusUpdate for it unless the generation or status changes. But skipped and resolved objects aren't added to `pending` in `Start` so that shouldn't be a problem.

I think we can change it to  do `len(pending) == 0` instead. And that might make the `conditionMet` function obsolete, since we only need to check one resource at a time now.

pkg/apply/taskrunner/task.go
```diff
@@ -198,6 +213,12 @@ func (w *WaitTask) skipped(taskContext *TaskContext, id object.ObjMetadata) bool
 	return false
 }
 
+// resourceStalled returns true if the resource is stalled.
+func (w *WaitTask) resourceStalled(taskContext *TaskContext, id object.ObjMetadata) bool {
```
karlkfi, line 217: `stalledByID` would be more consistent with `reconciledByID`

pkg/apply/event/event.go
karlkfi: Oh, I see kstatus is reporting FailedStatus only when it finds the Stalled = True condition. That's weird.

Is anything actually writing the Stalled condition? Or is this all just dead code?

IIRC, the request from Morgante was specifically around detecting failure of KCC resources, which I don't think use the Stalled condition.

I'd feel better about changing the API here if we knew what that was gonna look like with a real use case, even if it means using injected status check code.

pkg/apply/taskrunner/task.go
mortent: Agree with that. Adding information to the TaskContext isn't strictly needed, so I have removed it from this PR.

pkg/apply/taskrunner/task.go
mortent: Yeah, I think leaving it the way it is might be better for now. Having the logic in one place makes it easier to see what is going on.

pkg/apply/taskrunner/task.go
mortent: Updated the check to use `len(pending) == 0`. I think we can also do some cleanup of the `conditionMet` function and other functionality used by it. But I'll do that in a separate cleanup PR to keep this one small.

pkg/apply/event/event.go
mortent: So kstatus should report `FailedStatus` for situations where a resource is determined to be unlikely to reconcile. One way this could happen is that the controller itself sets the `Stalled` condition to `true`. Another would be that the kstatus library itself determines that a resource is unlikely to reconcile based on other information. An [example](https://github.com/kubernetes-sigs/cli-utils/blob/45a0b5b7e6b292c0a59c899aa50a3662a0c1cf7d/pkg/kstatus/status/core.go#L186) is when the Deployment controller sets the `ProgressDeadlineExceeded` condition. Up until now, we haven't actually used this information for anything, we have treated `Failed` resources the same as `InProgress`. This just make sure we don't end up waiting for the timeout to expire when we are only waiting for `Failed` resources.

As for supporting Config Connector resources, that would not be something that is handled in this library. But I want to make a separate change that allow users to the library to provide custom rules for computing status. So users of the library can build tools that can handle resource types with specific rules, for example Gatekeeper template/policies, Config Connector resources or any other types. Ref https://github.com/kubernetes-sigs/cli-utils/pull/383

mortent:
@karlkfi Updated the PR with the following changes:
 * Updated the `WaitEvent` to use operation `ReconcileFailed` rather than `ReconcileStalled`. Also updated the functions and variable names accordingly.
 * Make sure that resources can transition from `Failed` back to `InProgress`. Added a test for this.

ash2k:
I'm sorry, I will not be able to review PRs for some time.

pkg/apply/taskrunner/task.go
mortent: Removed this from the PR. We can add it back when it is needed.

pkg/apply/taskrunner/task.go
```diff
@@ -243,8 +277,9 @@ func (w *WaitTask) StatusUpdate(taskContext *TaskContext, id object.ObjMetadata)
 		}
 	}
 
-	// if all conditions are met, complete the wait task
-	if conditionMet(taskContext, w.pending, w.Condition) {
+	// If we no longer have any pending resources, the WaitTask
+	// can be completed.
+	if len(w.pending) == 0 {
 		// all reconciled - clear pending and exit
 		w.pending = object.ObjMetadataSet{}
```
haiyanmeng, line 284: Since `len(w.pending) == 0`, we can remove this line.

mortent: That is true. I have removed that line.

Issue #484, opened 2021-11-16
Feature Request: Validation Error Handling


karlkfi:
There's two schools of thought when handling validation:
1. Skip objects that fail validation and apply everything else
2. Exit early if any objects fail validation

## Option 1: Centralized Deployment Automation

Option 1 is desirable when the scope of the applied resources spans multiple domains (ex: multiple unrelated Deployments).

Option 1 is common with users that have a monorepo style deployment strategy, where multiple applications/systems share deployment automation. This is common when the organization follows "Ops Team" style ownership of deployment.

Pros:
- It's less work to automate in one place and have one team mange it.
- Makes it easier to coordinate deployment between application/system dependencies. 

Cons:
- If everything is applied all together, any one failure can block other unrelated applications/systems from updating or even break them with partial deployment.
- Up front validation would block applying of unrelated applications/systems.
- All applications/systems are deployed with the same client user account, requiring effective root permissions.

## Option 2: Distributed Deployment Automation

Option 2 is desirable when the scope of the applied resources maps to a single domain (ex: a single Deployment with it's immediate dependencies, ConfigMap, Secret, Service, Ingress, etc).

Option 2 is common with users that have a multi-repo or microservice style deployment strategy, where each application/system has its own deployment automation (even if they share common tooling). This is common when the organization follows "DevOps" style ownership of deployment.

Pros:
- Each repo/microservice acts as its own failure domain. If deployment fails, it should only affect that system and its downstream dependents.
- Up front validation can block applying, preventing broken deployments.
- Each applications/systems can be deployed with the different client user account, allowing least-privilege deployment security.

Cons:
- Requires more tooling and training to have multiple teams managing their own deployment automation.
- Makes it harder to coordinate deployment between application/system dependencies in different repos with different deployment automation. 

## Conclusion

Since both systems are viable and expected, cli-utils probably needs to support both.

## Proposal

One way to support both would be to add `ValidationEvents` and a new applier/destroyer option to `ExitOnValidationError` (or `RequireValidation`). This would continue to execute validation up front (and there's more we could add, like validating feature annotations), but instead of always sending an ErrorEvent and exiting, it would send a ValidationEvent for each failure and then only send an ErrorEvent and exit if the ExitOnValidationError flag was set.

[...]

Issue #455, opened 2021-10-28
Bug: dependency waiting blocks too aggressively


karlkfi:
# Problems
The applier/destroyer pipelines add Apply/Prune tasks with Wait tasks, waiting for each apply/prune group to become Current (after apply) or NotFound (after prune). This behavior coupled with complex resource sets with multiple dependency branches has the following impact:
1. apply/prune can be blocked for some object even if their dependencies are ready
2. blocking waits for the slowest reconciliation in the previous apply phase, even if it doesn't time out
3. because wait timeout currently causes the pipeline to terminate, any object that cause wait timeout blocks all objects in subsequent phases from being applied/pruned

## Example 1
For example, here's an infra dependency chain with branches (just deploying two GKEs with a CC cluster):
- namespace
  - rbac
  - GKE cluster1
    - node pool1 (depends-on cluster1)
  - GKE cluster2
    - node pool2 (depends-on cluster2)

If any cluster fails to apply (ex: error in config), then both node pools are blocked.

## Example 2
Another example is just using the same apply for multiple namespaces or CRDs and namespace:
- CRD
- namespace1
  - deployment1
- namespace2
  - deployment2

If any CRD or any namespace fails to apply (ex: blocked by policy webhook or config error), then all the deployments and everything in the namespaces are blocked.

# Possible solutions

## Continue on wait timeout
- This helps with problem 1 and 3, but not problem 2, and has the consequence of making failure take longer, because all resources will be applied, even if we know their dependency isn't ready.

## Dependency filter
- This would help remediate the side effect of "continue on wait timeout" by skipping apply/prune for objects with unreconciled dependencies
- This would need to be used on both apply and prune
- This would need to share logic with the graph sorter, which currently handles identifying dependencies (depends-on, apply-time-mutation, CRDs, and namespaces)

## Parallel apply 
- Applying objects in parallel during each Apply Task (and deleting in parallel for prune tasks) would speed up e2e apply time significantly, helping to mitigate dependency cost, but not actually solving either problem 1 or problem 2

## Async (graph) apply
- Building a branching pipeline, instead of flattening into a single synchronous pipeline would help isolate dependencies to only block things that depend on them.
- This is probably the best solution, but also the most complex and risky.
- This would probably require changing how events are tested/used. Consuming them in a single for loop might not be the best strategy any more; async listeners might be better.
- This would make both Dependency filter and Parallel apply obsolete.
[...]

PR #488, opened 2021-11-30
feat: Add ValidationPolicy & ValidationEvent

karlkfi:
- Add ValidationPolicy:
  ExitEarly (default) - exit before apply/prune if any objects are invalid
  SkipInvalid - apply/prune valid objects & skip invalid ones
- Add ValidationEvent to be sent once for every invalid object or set of
  objects, when the SkipInvalid policy is selected. For ExitEarly,
  return an error for reverse compatibility.
- Add validation.Collector to simplify aggregating validation errors
  from multiple sources and extracting invalid object IDs.
- Add invalid objects to the TestContext so they can be retained in
  the inventory (only if already present). This primarily applies to
  invalid annotations and dependencies. Objects without name or kind
  should never be added to the inventory.
- Update Solver to use validation.Collector and filter invalid objects.
- Add e2e test for invalid objects.
- Update Printers to handle ValidationEvent
- Add ExternalDependencyError & InvalidAnnotationError to make it easier
  to handle and introspect validation errors.

fixes: https://github.com/kubernetes-sigs/cli-utils/issues/484


pkg/object/validator/error.go
```diff
@@ -0,0 +1,44 @@
...
+	var b strings.Builder
+	_, _ = fmt.Fprintf(&b, "%d objects failed validation\n", len(mve.Errors))
+	for _, e := range mve.Errors {
+		b.WriteString(e.Error())
+	}
+	return b.String()
+}
+
+// ValidationError captures errors resulting from validation of a resources.
+type ValidationError struct {
```
haiyanmeng, line 30: `a resources` -> `a resource`

pkg/object/validator/validate.go
```diff
@@ -0,0 +1,186 @@
...
+	if scope == meta.RESTScopeNamespace && ns == "" {
+		return field.Required(field.NewPath("metadata", "namespace"), "namespace is required")
+	}
+	if scope == meta.RESTScopeRoot && ns != "" {
+		return field.Invalid(field.NewPath("metadata", "namespace"), ns, "namespace must be empty")
+	}
+	return nil
+}
+
+func (v *Validator) validateDependsOnAnnotation(obj *unstructured.Unstructured, ids object.ObjMetadataSet) []error {
```
haiyanmeng, line 99: Should we validate [CyclicDependencyError](https://github.com/kubernetes-sigs/cli-utils/blob/45a0b5b7e6b292c0a59c899aa50a3662a0c1cf7d/pkg/object/graph/graph.go#L139) here too?

pkg/object/validator/validate.go
```diff
@@ -0,0 +1,186 @@
...
+
+	deps, err := dependson.ReadAnnotation(obj)
+	if err != nil {
+		return []error{err}
+	}
+
+	id := UnstructuredToObjMetadata(obj)
+	var errors []error
+
+	// duplicates are not allowed
```
haiyanmeng, line 112: Why duplicates are not allowed here?

pkg/apply/applier.go
```diff
@@ -188,6 +199,10 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
 
 		// Build list of prune validation filters.
 		pruneFilters := []filter.ValidationFilter{
+			filter.InvalidObjectFilter{
```
haiyanmeng, line 202: Does the order of these filters matter?

pkg/apply/applier.go
```diff
@@ -176,7 +182,12 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
...
 		// Build list of apply validation filters.
-		applyFilters := []filter.ValidationFilter{}
+		applyFilters := []filter.ValidationFilter{
+			filter.InvalidObjectFilter{
+				InvalidIds: invalidObjects,
+				Policy:     options.ValidationPolicy,
+			},
+		}
 		if options.InventoryPolicy != inventory.AdoptAll {
 			applyFilters = append(applyFilters, filter.InventoryPolicyApplyFilter{
```
haiyanmeng, line 192: Does the order of these filters matter?

pkg/object/validator/policy.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2021 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package validator
+
+//go:generate stringer -type=ValidationPolicy
+type ValidationPolicy int
```
mortent, line 7: So it seems like `SkipInvalid` is the approach that best fits our overall approach to errors here. It essentially follows the "continue-on-error" direction, but also avoids applying resources that we know there are problems with. I think there will be some issues that will prevent us from continuing (name/namespace errors comes to mind. Also, circular dependencies isn't currently caught by the validation and it is not clear how we would skip resources in that situation).
Could we simplify by making this the behavior until we have evidence that something else is needed?


I'm not convinced there is much need for the `ApplyAll` policy here, as it seems to suggest we'll be applying resources that we know have problems and effectively ignoring any annotations that can't be parsed.

I think `ExitEarly` do have some good use-cases, but I'm uncertain if this is the right way to do it. For most of the validation, we want to let users shift it left and detect these issues in their CI/CD pipelines rather than when they are ready to apply. So it seems like we want this to be available outside of a pre-step to actually applying it. Also, we already provide a dry-run flag, which should detect these issues as well as many additional validation errors.

pkg/object/graph/depends.go
```diff
@@ -83,20 +83,25 @@ func ReverseSortObjs(objs object.UnstructuredSet) ([]object.UnstructuredSet, err
...
-				// TODO: fail task if parse errors?
+				// Log & skip - Should have been caught by validation
 				klog.V(3).Infof("failed to add edges from: %s: %s", id, err)
-				return
+				continue
 			}
 			for _, sub := range subs {
-				// TODO: fail task if it's not in the inventory?
 				dep := sub.SourceRef.ObjMetadata()
+				if !ids.Contains(dep) {
```
mortent, line 100: Could we handle the validation here where it is needed, rather than doing all validation early? I realize doing it all early is beneficial for detecting all errors early and then report them all in an early exit, but it seems like some validation errors will only be caught later. Circular dependencies would be one of them.

pkg/apply/applier.go
```diff
@@ -140,15 +142,19 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
...
-		if err := (&object.Validator{
-			Mapper: mapper,
-		}).Validate(objects); err != nil {
-			handleError(eventChannel, err)
-			return
+		// Validate objects
+		invalidObjects, err := validateObjects(objects, mapper, eventChannel)
+		if err != nil {
+			klog.V(4).Infof("calculated %d invalid objs", len(invalidObjects))
+			if options.ValidationPolicy == validator.ExitEarly {
```
mortent, line 149: It seems like errors in name and/or namespace would be somewhat different than other validation errors. They will have to be detected before we update the inventory. Skipping those means we would make sure they are not added to the inventory, but also make sure we don't end up accidentally pruning any resources that have been skipped.

pkg/object/validator/validate.go
karlkfi: In order to find a cyclic error, we'd have to build the graph first. It seems pretty inefficient to do that twice. I'm not sure if we want to move graph building forward in front of validation... and building the graph might require valid inputs...

I guess we could make the graph SortObj emit events, but it doesn't do that now so it would require some significant changes...

pkg/object/validator/validate.go
karlkfi: If the user put two of the same dependencies in the same depends-on annotation, it's probably a mistake, right?

pkg/apply/applier.go
karlkfi: They get executed in this order. Filtering invalid objects first prevents the other filters from erroring due to an invalid object.

pkg/apply/applier.go
karlkfi: They get executed in this order. Filtering invalid objects first prevents the other filters from erroring due to an invalid object.

pkg/object/validator/policy.go
karlkfi: > So it seems like SkipInvalid is the approach that best fits our overall approach to errors here.

For now, yes. Tho in the future I'd like to add change groups so we can continue on error by default, but skip whole change groups if any of their objects are invalid. With that behavior, if you don't specify any change groups, all resources are considered part of the same change group, meaning that the default behavior would be equivalent to ExitEarly. Then if you want SkipInvalid behavior you just have to define multiple independent change groups. Or you could opt into SkipInvalid for all objects by default, equivalent to giving every object its own change group.

> I think there will be some issues that will prevent us from continuing

Yes. It will likely error in multiple places. I'll need to add tests for invalid input and make sure it handles them as desired.

> Also, circular dependencies isn't currently caught by the validation and it is not clear how we would skip resources in that situation.

Agreed. It's not obvious what the user's desired behavior would be in the case of circular dependencies either. I think for now we can just throw an error and exit and figure out how to skip it later, if needed.

> Could we simplify by making [SkipInvalid] the behavior until we have evidence that something else is needed?

@seans3 suggested using SkipInvalid as the default behavior. I could be convinced to do it that way, but I explained above why I think ExitEarly is a more forwards compatible default. 

IME building continuous deployment pipelines, I have always in the past preferred exiting early and stopping deployment if there's any problem that needs human input. Validation errors are almost always user mistakes, and partial deployment can easily cause production outage. So if given the option as a user, I would choose ExitEarly behavior. 

That said, I know some users have asked for continue-on-error, and especially after you've already started applying, it's not obvious if continuing or stopping would cause more brokenness. But I suspect that if we presented these same users with the change group option I suggested, they would prefer that over just continue-on-error behavior.

> I think ExitEarly do have some good use-cases, but I'm uncertain if this is the right way to do it.

This is the current behavior. Changing it would violate the principle of least surprise for users. Plus, Config Sync already does it's own validation earlier and would exit early anyway. It just doesn't validate all the same things, like the dependency annotations.

> For most of the validation, we want to let users shift it left and detect these issues in their CI/CD pipelines rather than when they are ready to apply. So it seems like we want this to be available outside of a pre-step to actually applying it. Also, we already provide a dry-run flag, which should detect these issues as well as many additional validation errors.

IME, dry-run is only really useful in a few cases:
1. When applying manually and you want to be see a diff of the changes to make sure hydration didn't have any unexpected side effects.
2. When there are multiple clients modifying the same files and you want to make sure you're not reverting someone else's changes accidentally.
3. When applying with automation and you want a record of the diff to be stored for debugging later or the diff needs manual approval.

There's not really anything to run between dry-run and apply that makes it useful to do a dry-run in CI/CD otherwise, is there? And in all cases, you'd still want to validate again in the apply, in case they didn't run dry-run, right? 

In any case, this validation code is the same code being used for both the apply and the dry-run, and I'm pretty sure we want the options and defaults to be the same for both, right?

pkg/apply/applier.go
karlkfi: AFAIK, skipped objects aren't added or removed from the inventory. If they were previously in the inventory, I think they're retained, right?

So is there a problem?

pkg/object/graph/depends.go
karlkfi: I think I can make SortObjs return a list of validation errors, without exiting on each one. Then we can move the SortObjs call from the solver to the applier/destroyer and send the ValidationEvents from there.

seans3 (commented):
I believe this approach to validation is misguided for a couple reasons:

* We should emphasize `--dry-run` as the primary validation mechanism. Instead of attempting to figure out problems ahead of time, `--dry-run` runs as much of the entire apply algorithm as possible, surfacing errors throughout the algorithm. This type of validation will almost always be better than pre-run client-side validation. It looks like the validations added here for `depends-on` and `apply-time mutation` are already implemented as errors in `--dry-run`. I do not see much utility for these validations other than better error messages.
* Exiting early will be especially problematic, since this makes significant assumptions about what the library client wants and differs dramatically from the underlying `kubectl apply` which continues applying when encountering an apply error. Exiting early will probably require manually calling the final inventory set task to ensure the prune/inventory algorithm continues to work.
* If a library clients want stronger validation (or domain-specific validation), they can always implement it themselves before calling the library apply. This library has to work for all clients, and exiting early makes strong assumptions.

I believe the goal of client-side validation should be much more modest, such as generating a more informative error messages. If the history of Kubernetes CLI's is any guide, over-ambitious client-side validation will be jettisoned eventually for dry-run and server-side validation. But as long as there is a validation option other than `Exit Early`, I will not object to this PR.

karlkfi:
@seans3 The current existing behavior already exits early if the Validator errors. I don't understand why you think that's misguided. The code is littered with hundreds of places that assume that minimal validation has been done. It's not going to be a small change to fix that to skip invalid objects (which seems to be what you want).

What this PR primarily does is adds the ability to skip those invalid objects. 

I also added more explicit depends-on & apply-time-mutation errors, but that's because these are client-side annotations and the server won't validate them. So if they're invalid and the applier continues without skipping them (the current behavior), then that's behaviorally incorrect. We should never completely ignore invalid depends-on & apply-time-mutation errors. We should either skip those invalid objects or exit early. Either way, the user needs to be notified. That's why I added the ValidationError, to make sure the user always gets notified, regardless of whether they chose SkipInvalid or ExitEarly.

Now that I've said that tho, I think the ApplyAll policy is probably only useful when not using depends-on or apply-time-mutation. Otherwise the behavior when they're invalid would be obviously incorrect. ApplyAll would basically be opting out of client-side validation and ignoring errors in client-side features. That's basically what you're championing when you're saying the applier should be more like kubectl and do less validation. The big difference is that kubectl doesn't have any client-side-only features like depends-on or apply-time-mutation.

pkg/object/validator/validate.go
karlkfi: Alright, I've moved all the dependency and sorting errors to SortObjs and moved that forward in the applier so we can use the errors as validation errors, and then modified SortObjs to return a MultiError so we get more than one at a time.

pkg/object/validator/validate.go
karlkfi: I've moved this to SortObjs and comments on the why,.

pkg/apply/applier.go
karlkfi: I've added invalid objects to the TaskContext now. Invalid objects will only be added to the inventory IFF they already existed in the inventory. So that can only really happen for annotations that were changed after being applied, and not for missing name/kind/namespace.

karlkfi:
I've made significant changes to the handling of dependency validation. Please re-review.

- Move dependency validation from Validator to SortObjs.
- Move SortObjs call from solver to applier/destroyer.
  This allows the resulting errors to be treated as validation errors.
- Modify SortObjs to return a MultiError so that all validation errors
  can be sent and invalid objects can be skipped or cause early exit.
- Add invalid objects to the TestContext so they can be retained in
  the inventory (only if already present). This primarily applies to
  invalid annotations and dependencies. Objects without name or kind
  should never be added to the inventory.
- Remove invalid objects from the applySet/pruneSet, rather than
  filtering and skipping them in the apply/prune/wait tasks. This
  reduces the number of events that are redundant now with validation
  events.
- Handle CyclicDependencyError as a validation error. It applies to
  multiple objects, instead of just one, like ValidationError.
- Handle validation of dependency errors in destroyer as well as
  the applier.
- Replace many -OrDie object commands that were panicing when an
  invalid object was found. Add ValidateObjMetadata to do the
  validation that the -OrDie functions were doing.
- Replace MultiValidationError with a more generic MultiError.
- Simplify ValidationError to optionally wrap MultiError.
- Modify ValidationError output to be easier to read.
- Add e2e test for invalid object handling.

I still have more tests to add.

pkg/apply/event/event.go
```diff
@@ -198,11 +204,11 @@ func (age ActionGroupEvent) String() string {
 type ApplyEventOperation int
 
 const (
-	ApplyUnspecified ApplyEventOperation = iota
-	ServersideApplied
-	Created
-	Unchanged
-	Configured
+	ApplyUnspecified  ApplyEventOperation = iota
+	ServersideApplied                     // TODO: rename AppliedServerside
```
karlkfi, line 208: Most of the other event operations are prefixed by something consistent.

Do we want to change these to use Applied/Apply as prefix?

It would make them easier to auto-complete. We can still print them as something without the prefix, because the generator allows specifying the string value as an inline comment.

pkg/apply/applier.go
```diff
@@ -156,6 +159,46 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
...
+			if options.ValidationPolicy == validation.ExitEarly {
+				handleError(eventChannel, multierror.New(validationErrors))
+				return
+			}
+		}
+		// Extract unique IDs from ValidationErrors
+		invalidIds := invalidIds(validationErrors)
+		klog.V(4).Infof("invalid objects: %d", len(invalidIds))
+		// Don't apply or prune invalid objects
+		// TODO: what about dependencies?
```
karlkfi, line 191: This will break waiting for dependencies if the dependency is invalid, but only if SkipInvalid is enabled. Since ExitEarly is default for reverse compatibility, this is not technically a regression. The intended fix will be a new feature to filter objects whose dependencies aren't reconciled. The dependency filter is discussed in https://github.com/kubernetes-sigs/cli-utils/issues/455

pkg/apply/applier.go
```diff
@@ -156,6 +159,46 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
...
+		applySets = removeObjects(applySets, invalidIds)
+		pruneSets = removeObjects(pruneSets, invalidIds)
```
karlkfi, line 193: This new pre-filtering might make the InvalidObjectFilter obsolete. If the objects never get added to the apply/prune tasks, they don't need to be filtered then.

pkg/apply/event/event.go
karlkfi: This should probably be punted to a different PR. It will technically break reverse compatibility by changing an existing event.

pkg/object/objmetadata_set.go
```diff
@@ -130,34 +130,27 @@ func (setA ObjMetadataSet) Diff(setB ObjMetadataSet) ObjMetadataSet {
...
+// Order may or may not remain consistent.
+func (setA ObjMetadataSet) Unique() ObjMetadataSet {
+	return ObjMetadataSetFromMap(setA.ToMap())
+}
+
 // Hash the objects in the set by serializing, sorting, concatonating, and
 // hashing the result with the 32-bit FNV-1a algorithm.
-func (setA ObjMetadataSet) Hash() (string, error) {
+// Used to quickly identify the set of resources in the inventory object.
+func (setA ObjMetadataSet) Hash() string {
```
karlkfi, line 142: This hash change can probably be extracted to a different PR. It's needed for this PR, but can be made standalone.

pkg/object/unstructured.go
```diff
@@ -18,49 +18,26 @@ var (
...
-func UnstructuredToObjMeta(obj *unstructured.Unstructured) (ObjMetadata, error) {
-	return CreateObjMetadata(obj.GetNamespace(), obj.GetName(),
-		obj.GroupVersionKind().GroupKind())
 }
 
 // UnstructuredToObjMetaOrDie extracts the identifying information from an
 // Unstructured object and returns it as Objmetadata. If the values doesn't
 // pass validation, the function will panic.
+// TODO: Deprecate in favor of handling the error to impl continue on error.
 func UnstructuredToObjMetaOrDie(obj *unstructured.Unstructured) ObjMetadata {
```
karlkfi, line 34: Even with the changes in this PR, there are still 127 usages of UnstructuredToObjMetaOrDie in this repo. 

A lot of them are in tests and can probably be replaced by UnstructuredToObjMetadata, but the rest are in the code and could cause a panic if executed on a code path during a SkipInvalid apply/destroy.

karlkfi: I'm not sure if it's a good idea to try to replace them all in this PR or not.

pkg/apply/applier.go
karlkfi: removed InvalidObjectFilter

pkg/apply/event/event.go
karlkfi: Punted to https://github.com/kubernetes-sigs/cli-utils/issues/502

karlkfi:
Extracted replacement of -OrDie (panic) functions to https://github.com/kubernetes-sigs/cli-utils/pull/505

This should vastly reduce the number of files changed in this PR, once rebased.

pkg/object/unstructured.go
karlkfi: This refactor was moved to https://github.com/kubernetes-sigs/cli-utils/pull/505

karlkfi:
This has been rebased on top of https://github.com/kubernetes-sigs/cli-utils/pull/505 and https://github.com/kubernetes-sigs/cli-utils/pull/506

pkg/object/validation/error.go
```diff
@@ -0,0 +1,65 @@
...
+package validation
+
+import (
+	"fmt"
+	"strings"
+
+	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+type Error interface {
```
mortent, line 13: It would be useful to explain the reason for this interface. It looks like `verror` is the only implementation, so why do we need the interface?

mortent (commented):
I've done a first pass. The direction of allowing users to continue even if a subset of the resources have issues is aligned with the direction we are going.

I know some code has already been moved out of this PR, but I'm still finding it hard to do a good review since there is a lot going on. Is there a way to move more out into a separate PR? In particular, the changes to the `pkg/object/graph` package. Also maybe some of the additional validation error changes could be added in a separate PR prior to adding the ValidationEvent changes.

pkg/object/validation/error.go
```diff
@@ -0,0 +1,65 @@
...
+
+func NewError(cause error, ids ...object.ObjMetadata) Error {
+	return &verror{
+		ids:   object.ObjMetadataSet(ids),
+		cause: cause,
+	}
+}
+
+// verror wraps an error with the object or objects it applies to.
+type verror struct {
```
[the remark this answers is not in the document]
karlkfi: oops. the ValidationError struct is obsolete. It's replaced by the validation.Error interface, but got dragged along in a rebase. I'll delete it.

pkg/object/validation/error.go
karlkfi: I can probably get rid of the interface. I thought it would be useful, but all it ends up doing is avoiding a `*` when casting.

Technically the Unwrap function is a new Go standard, but they didn't make a public interface for it when they added it to the logger. So it doesn't actually need a parent interface to compose it.

pkg/object/validation/validate.go
```diff
@@ -1,17 +1,18 @@
...
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"sigs.k8s.io/cli-utils/pkg/multierror"
+	"sigs.k8s.io/cli-utils/pkg/object"
 )
 
 // MultiValidationError captures validation errors for multiple resources.
```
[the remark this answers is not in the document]
karlkfi: yes, by the Validation func.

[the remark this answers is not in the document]
karlkfi: nope. rebase error. replaced by MultiError

pkg/apply/applier.go
```diff
@@ -141,22 +144,66 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
...
+
+		// Send each validation error as a ValidationEvent
+		vErrors = multierror.Unwrap(vErrors...)
+		klog.V(4).Infof("validation errors: %d", len(vErrors))
+		if len(vErrors) > 0 {
+			for _, err := range vErrors {
+				handleValidationError(eventChannel, err)
+			}
+			// Exit early, if configured to do so
+			if options.ValidationPolicy == validation.ExitEarly {
```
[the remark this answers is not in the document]
karlkfi: Technically, no. But it makes the new default ExitEarly mode reverse compatible without needing the client to watch for ValidationEvents.

pkg/apply/solver/solver_test.go
```diff
@@ -104,19 +104,20 @@ metadata:
 
 func TestTaskQueueBuilder_AppendApplyWaitTasks(t *testing.T) {
```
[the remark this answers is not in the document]
karlkfi: Probably not, except as example of how to use the Solver when dealing with dependencies. These are what the inputs will look like now that sorting has been externalized.

pkg/apply/destroyer.go
```diff
@@ -136,9 +142,48 @@ func (d *Destroyer) Run(ctx context.Context, inv inventory.InventoryInfo, option
 				InvPolicy: options.InventoryPolicy,
 			},
 		}
+
+		// Sort deletes into groups based on reverse dependency ordering
```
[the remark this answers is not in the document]
karlkfi: They're called deletes in the destroyer and prunes in the applier. I wish they were all just called deletes...

I suppose I could try wrapping this duplicate chunk in a function.

karlkfi:
Extracted the MultiError changes to https://github.com/kubernetes-sigs/cli-utils/pull/508

pkg/apply/applier.go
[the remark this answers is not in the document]
karlkfi: Occurs to me that this could make for redundant printed output, since the printers print the validation events and error events.

pkg/apply/destroyer.go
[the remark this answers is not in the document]
karlkfi: I found a way to abstract some of the code into a `validateAndSort` function. It sounds weird, but sorting can produce validation errors.

pkg/object/validator/policy.go
karlkfi: I've removed ApplyAll for now. It's not actually possible to apply resources without a kind or name. So the feature wouldn't make much sense.

I've left ExitEarly as the default, because that's the current behavior. We can change this later, if we want.

This PR primarily adds SkipInvalid and makes it opt-in.

pkg/apply/applier.go
```diff
@@ -141,22 +144,66 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
...
 
+		// Decide which objects to apply and which to prune
 		applyObjs, pruneObjs, err := a.prepareObjects(invInfo, objects, options)
 		if err != nil {
 			handleError(eventChannel, err)
 			return
 		}
 		klog.V(4).Infof("calculated %d apply objs; %d prune objs", len(applyObjs), len(pruneObjs))
 
+		// Sort applies into groups based on dependency ordering
```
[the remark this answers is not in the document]
karlkfi: I pulled out some of the logic to a `validateAndSort` function.

> The original design was that more of of the logic for determining which resources should be applied/pruned and the order was in the Solver type.

Correct. This PR changes that, which makes the solver less of a solver. Another alternative would be to make the solver return a list of validation errors. Would that be preferable?

pkg/apply/applier_test.go
```diff
@@ -1179,15 +1181,257 @@ func TestApplier(t *testing.T) {
...
+					EventType: event.ValidationType,
+					ValidationEvent: &testutil.ExpValidationEvent{
+						Identifiers: object.ObjMetadataSet{
+							object.UnstructuredToObjMetadata(
+								testutil.Unstructured(t, resources["deployment"], JSONPathSetter{
+									"$.kind", "",
+								}),
+							),
+						},
+						Error: testutil.EqualErrorType(validation.NewError(nil)), // TODO: be more specific
```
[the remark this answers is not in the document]
karlkfi: Fixed.

pkg/apply/applier.go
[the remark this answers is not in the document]
karlkfi: I've changed the behavior to not send ValidationEvents when configured to ExitEarly.

pkg/apply/applier.go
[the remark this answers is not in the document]
karlkfi: Ok, I moved the sorting back into the Solver. It's slightly cleaner and fewer changes. To make it work, i had to inject InvalidIds into the Solver, because SortObjs needs the full input set, and then invalid objects need to be filtered afterwards, before the Task is created.

One interesting side effect of this is that the InvAddTask WILL include "invalid" objects that fail during sorting (depends-on, apply-time-mutation, and cycle errors), because the InvAddTask is built before sorting happens. This is actually exactly the same behavior as the current code. So it's not a regression. If we want to change this behavior, we can make the InvAddTask do filtering using the InvalidIds in the TaskContext.

pkg/apply/destroyer.go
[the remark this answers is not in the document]
karlkfi: Duplication should be reduced now that the sorting is moved back into the Solver.

pkg/apply/solver/solver_test.go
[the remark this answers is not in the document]
karlkfi: These are no longer duplicates with sorting moved back into the Solver. So I reverted this file and added some error details.

karlkfi:
Ok, I did a big refactor to move the sorting bakc into the Solver. And got the tests finally passing again.

TODO: 
- extract the taskrunner refactor to inject the TaskContext
- extract the Hash change to not return an error
- extract the Sorting fixes
- add test case for InvSetTasks filtering invalid objects

karlkfi:
/unhold

All TODOs finished. Ready for another review pass.

pkg/apply/task/inv_set_task.go
```diff
@@ -105,6 +105,12 @@ func (i *InvSetTask) Start(taskContext *taskrunner.TaskContext) {
 		klog.V(4).Infof("remove from inventory %d abandoned objects", len(abandonedObjects))
 		invObjs = invObjs.Diff(abandonedObjects)
 
+		// If an object is invalid and was previously stored in the inventory,
+		// then keep it in the inventory so it can be applied/pruned next time.
+		invalidObjects := i.PrevInventory.Intersection(taskContext.InvalidObjects())
```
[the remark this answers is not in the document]
karlkfi: Added

pkg/apply/solver/solver.go
```diff
@@ -90,23 +95,18 @@ type Options struct {
...
-	}, nil
+	return &TaskQueue{tasks: t.tasks}, multierror.Wrap(t.errs...)
 }
 
 // AppendInvAddTask appends an inventory add task to the task queue.
 // Returns a pointer to the Builder to chain function calls.
 func (t *TaskQueueBuilder) AppendInvAddTask(inv inventory.InventoryInfo, applyObjs object.UnstructuredSet,
 	dryRun common.DryRunStrategy) *TaskQueueBuilder {
-	klog.V(2).Infoln("adding inventory add task")
+	applyObjs = diffObjsByID(applyObjs, t.InvalidIds) // skip invalid objects
```
karlkfi, line 108: Should we replace this pre-filtering with the original InvalidObjectFilter instead?

mortent (approved):
I think this looks good. I see that we have had some discussions on this PR about when validation should happen, but the primary feature here is that we can continue applying resources in many more situations, which I think is a direction we all agree on. Allowing early exit seems reasonable to support, not least because that is the current behavior.

/approve

mortent:
I'll remove the approval just to make sure @seans3 also has a chance to look at it before it merges.
/approve cancel

pkg/apply/solver/solver.go
karlkfi: Extracted filtering to validation.Collector

pkg/apply/applier.go
```diff
@@ -322,3 +359,37 @@ func localNamespaces(localInv inventory.InventoryInfo, localObjs []object.ObjMet
...
+		}
+	}
+}
+
+// extractInvalidIds extracts IDs from a list of validation errors, ingoring
+// other error types.
+func extractInvalidIds(errs []error) object.ObjMetadataSet {
+	var invalidIds object.ObjMetadataSet
+	for _, err := range errs {
+		if tErr, ok := err.(*validation.Error); ok {
```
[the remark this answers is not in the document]
karlkfi: Why is that better than type comparison?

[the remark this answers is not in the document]
karlkfi: Oh, i see. errors. As will recursively Unwrap errors until it finds the specified type. That makes sense, in case any of the validation errors are wrapped. Unfortunately, it wont work if there's a MultiError in the stack. But in this case, I already unwrapped those before adding them to the Collector (this code is in the Collector now).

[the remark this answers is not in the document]
karlkfi: done

seans3 (changes requested):
I do not believe exiting early should be the default behavior. Even if this behavior exists now, it is incorrect and should be fixed to continue on error. Exiting early is also not consistent with the `kubectl apply` behavior which continues on error.

karlkfi:
> I do not believe exiting early should be the default behavior. Even if this behavior exists now, it is incorrect and should be fixed to continue on error. Exiting early is also not consistent with the kubectl apply behavior which continues on error.

It's likely that ConfigSync and kpt will both initially use ExitEarly when adopting this new code. They may later switch to either hard coding SkipInvalid OR piping the option through to the user. Does that affect your default suggestion @seans3 ?

karlkfi:
FWIW, I do not consider ExitEarly to a be a bug. Its a feature which arises from the desire to group related changes. Of course, Kubernetes doesn't have atomic applies or groups, but you can at least avoid applying a Deployment if its Secret, ServiceAccount, or ConfigMap is invalid. This behavior is highly desired when you take the time to granularly group your objects.

On the other hand, SkipInvalid is attractive when you haven't or don't intend to group your objects. This comes up regularly when multiple teams share a single config repo or delegate config management to a central SRE/DevOps/Ops team.

I'm ok with either being default. I don't think it really matters. There are pros and cons on both sides.
- ExitEarly is reverse compatible, which is desirable for existing kpt users because they're releasing beta versions and breaking changes aren't being released as major versions yet.
- SkipInvalid might be more desirable to new kpt users who are more familiar with kubectl.
- Any wrapping client can select their own default... So as long as we communicate breaking changes, I'm not against them.

I'm biasing slightly towards reverse compatibility here, but if anyone else wants to chime in with a 3rd vote I don't mind making the change.

mortent:
I would probably prefer that we keep the current behavior here (so ExitEarly is the default), but I don't feel strongly about it.

haiyanmeng:
For Config Sync users, the default should be ExitEarly, which is closer to the existing validation behavior in Config Sync.

seans3 (approved):
/lgtm

PR #489, opened 2021-12-01
Allow status readers to be reused between invocations

mortent:
Instead of having the `StatusPoller` create a new instance of every `StatusReader` each time it is invoked, this creates the `StatusReader` instances when the `StatusPoller` is created. In order to do this, the `StatusReader` interface is changed so the `ClusterReader` is passed in every time.


mortent:
@makkes This makes some changes to how users can provide custom StatusReaders to the poller. You provided this functionality, so I want to make sure it doesn't cause any issues for you use-cases.

makkes (approved):
I'm happy to see that factory func go away, it always felt a little complicated to me. :-) Great change and thanks for the ping @mortent.

seans3 (approved):
This looks good to me. I'll merge it within the next day if there is no other feedback by then.

PR #491, opened 2021-12-07
Fix missing action value in table output

mortent:
The table output currently doesn't show any value in the `Action` column. The reason is that we use the value of the ResourceAction to determine whether a resource will be applied or pruned/deleted. With the introduction of the `WaitEvent`, we end up setting the ResourceAction of all resource to `Wait`.


seans3 (approved):
/lgtm

PR #494, opened 2021-12-09
Validate the value of the output flag

mortent:
Make sure the value provided for the `output` flag corresponds to a supported printer


seans3 (approved):
/lgtm

PR #495, opened 2021-12-15
More flexible resource type to StatusReader mapping

mortent:
Currently all StatusReaders must be provided for a specific GroupKind. This is awkward when all resources in a group follow the same pattern and therefore can leverage a single StatusReader.
This change allows each StatusReader to decide whether it can compute status for a resource with a specific GroupKind.


seans3 (approved):
/lgtm

PR #496, opened 2021-12-16
Remove StatusPoller factory and support custom status readers

mortent:
This moves the code from creating a `StatusPoller` from a `util.Factory` into the polling package rather than keeping it separate.
It also allows for providing custom status readers when creating the poller.


seans3 (approved):
/lgtm

PR #497, opened 2021-12-18
Update dependencies to Kubernetes v1.23.1

ash2k:


cmd/main.go
```diff
@@ -10,15 +10,14 @@ import (
 
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/component-base/cli"
 	"k8s.io/kubectl/pkg/cmd/util"
-	"k8s.io/kubectl/pkg/util/logs"
```
ash2k, line 14: This package is gone now, see https://github.com/kubernetes/kubernetes/commit/21d1bcd6b8498370832fa76f680d3de56837bc83

karlkfi:
@haiyanmeng is Config Sync v1.11 going to be able to take v1.23 libs? I know you didn't want to take v1.22 for Config Sync v1.10.

I'm not sure how up to date we want to be. It seems to affect every downstream client.

ash2k:
I'm bumping libraries to pick up https://github.com/kubernetes/kubernetes/pull/105623 to be able to stop using the reflection hack to reset rest mapper - see https://github.com/kubernetes-sigs/cli-utils/issues/421.

If a downstream project cannot update to a newer version of a library, they are not forced to. However, if another consumer needs a fix in the library, their only option is to fork and fix, which is unfortunate as it creates work for everybody to maintain their forks. There needs to be a balance, of course, but to me it seems strange to slow things down because of a particular consumer. I wonder what are the difficulties with bumping to a newer version?

ash2k:
@karlkfi Could you LGTM if it's fine to merge?

pkg/object/infos.go
```diff
@@ -59,7 +59,7 @@ func UnstructuredToInfo(obj *unstructured.Unstructured) (*resource.Info, error)
 		// kyaml adds both annotations for the time being, so we need to remove them both
 		// before apply.
 		delete(annos, kioutil.PathAnnotation)
-		delete(annos, kioutil.LegacyPathAnnotation)
+		delete(annos, kioutil.LegacyPathAnnotation) //nolint:staticcheck
```
karlkfi, line 62: Why are these new nolints needed?

karlkfi (changes requested):
Please amend the commit messages to have fix or feat prefixes. I want to try to start following https://www.conventionalcommits.org/en/v1.0.0/#summary to make change logs easier to produce. If there's a breaking change, add a BREAKING CHANGE line at the end with detail.

I have a feeling we should mark both the go upgrade and the k8s lib upgrades as breaking changes, since they probably require client changes.

pkg/object/infos.go
ash2k: Build started failing because of these after Go 1.16 -> 1.17 bump in https://github.com/kubernetes/test-infra/pull/24709. See https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cli-utils/497/cli-utils-presubmit-master/1482230404432793600#1:build-log.txt%3A276

karlkfi (approved):
The breaking change message seems a little redundant, but we don't have automation to generate the changelog yet and can manually modify it afterwards anyway, as long as we can find it.

PR #498, opened 2021-12-22
Don't print status for all events for reach WaitEvent

mortent:
This is a leftover from previous behavior in the printers where we only printed status events during WaitTasks. Currently we instead print status events directly whenever they happen, which means that this is no longer needed and leads to printing of unnecessary status events.


ash2k (approved):
/lgtm

Issue #443, opened 2021-10-19
Consider removing advanced command line features from project


mortent:
This repo contains cli-commands which includes some rather complex output formats. I think going forward we want this project to focus primarily on providing a library for actuation, rather than a cli tool. Having some simple cli commands are useful for development, but I think we should consider dropping support for the table format output and make the `kapply` cli tool just a thin wrapper around the applier/destroyer.

PR #499, opened 2021-12-26
Allow customization of the dry-run output

mortent:
This allows users to customize the output added to the event output for preview/dry-run

### DIFF
--- a/cmd/apply/cmdapply.go
+++ b/cmd/apply/cmdapply.go
@@ -19,6 +19,7 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/inventory"
 	"sigs.k8s.io/cli-utils/pkg/manifestreader"
 	"sigs.k8s.io/cli-utils/pkg/printers"
+	"sigs.k8s.io/cli-utils/pkg/printers/printer"
 )
 
 func GetApplyRunner(factory cmdutil.Factory, invFactory inventory.InventoryClientFactory,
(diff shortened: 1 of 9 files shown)


ash2k:
Hm, I'm not sure. I think "printer" is a very good/powerful abstraction already and we don't need another indirection/abstraction that is injected into a printer to tweak one little bit of output. I.e. usefulness/new API surface ratio is very low in my opinion. I think instead it'd be good to make it easier to build new printers by making parts of existing ones more reusable (somehow).

I have an example where this doesn't necessarily work. Suppose I have a printer that prints events in a very compact binary format. It doesn't want a string for the strategy, it already represents it as a number. However, now it is kind of forced to do something with that string. This tells me that the new interface tries to split a single concern into two pieces in an unfortunate way that doesn't work well.

mortent:
@ash2k I do mostly agree with you concerns here.

The current implementation has some challenges though:
* We print the dry-run/preview information for the `events` output format, but not for the `json` or `table` formats.
* It uses the term `preview` rather than the more common term of `dry-run` (used in kubectl,...).
* It is pretty verbose, as the output always just repeats the value provided and is always the same value for every line.

I agree that we should try to make it easier to write new printers, although I'm not convinced whether we should have the complex printers in the repo (https://github.com/kubernetes-sigs/cli-utils/issues/443).

How about we just make it simple, and remove the dry-run/preview value from the events printer? It simplifies the output and removes the need for any changes like this PR.

ash2k:
@mortent 

> same value for every line

Yeah, that just creates noise.

> I'm not convinced whether we should have the complex printers in the repo

As long as it's not something a lot of people will definitely use, I agree.

> How about we just make it simple, and remove the dry-run/preview value from the events printer?

I don't mind that, as long as nobody else objects. 👍

PR #500, opened 2022-01-05
feat: Return a specific error type from the printers when resources fail to apply/delete/reconcile

mortent:
Instead of returning a generic error from the printer when one or more resource fails to apply/prune/delete or reconcile, it will return a specific type that includes the counts. This allows clients to more easily distinguish these type of errors from more fatal errors.
An example where this is useful, is to be able to use a different exit codes in a cli when resources fail to apply/prune/delete or reconcile, rather than the operation failing due to other reasons.

### DIFF
--- a/pkg/print/common/errors.go
+++ b/pkg/print/common/errors.go
@@ -1,0 +1,45 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package common
+
+import (
+	"fmt"
+
+	"sigs.k8s.io/cli-utils/pkg/print/stats"
+)
+
+// ResultErrorFromStats takes a stats object and returns either a ResultError or
+// nil depending on whether the stats reports that resources failed apply/prune/delete
+// or reconciliation.
+func ResultErrorFromStats(s stats.Stats) error {
+	if s.FailedActuationSum() > 0 || s.FailedReconciliationSum() > 0 {
+		return &ResultError{
+			Stats: s,
+		}
+	}
+	return nil
+}
+
+// ResultError is returned from printers when the apply/destroy operations completed, but one or
+// more resources either failed apply/prune/delete, or failed to reconcile.
+type ResultError struct {
+	Stats stats.Stats
+}
+
+func (a *ResultError) Error() string {
+	switch {
+	case a.Stats.FailedActuationSum() > 0 && a.Stats.FailedReconciliationSum() > 0:
+		return fmt.Sprintf("%d resources failed, %d resources failed to reconcile before timeout",
+			a.Stats.FailedActuationSum(), a.Stats.FailedReconciliationSum())
+	case a.Stats.FailedActuationSum() > 0:
+		return fmt.Sprintf("%d resources failed", a.Stats.FailedActuationSum())
+	case a.Stats.FailedReconciliationSum() > 0:
+		return fmt.Sprintf("%d resources failed to reconcile before timeout",
+			a.Stats.FailedReconciliationSum())
+	default:
+		// Should not happen as this error is only used when at least one resource
+		// either failed to apply/prune/delete or reconcile.
+		return "unknown error"
+	}
+}
--- a/pkg/print/list/base_test.go
+++ b/pkg/print/list/base_test.go
@@ -1,0 +1,74 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package list
+
+import (
+	"testing"
+
+	"sigs.k8s.io/cli-utils/pkg/apply/event"
+	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/cli-utils/pkg/print/stats"
+	"sigs.k8s.io/cli-utils/pkg/printers/printer"
+	printertesting "sigs.k8s.io/cli-utils/pkg/printers/testutil"
+)
+
+func TestPrint(t *testing.T) {
+	printertesting.PrintResultErrorTest(t, func() printer.Printer {
+		return &BaseListPrinter{
+			FormatterFactory: func(previewStrategy common.DryRunStrategy) Formatter {
+				return newCountingFormatter()
+			},
+		}
+	})
+}
+
+func newCountingFormatter() *countingFormatter {
+	return &countingFormatter{}
+}
+
+type countingFormatter struct {
+	applyEvents      []event.ApplyEvent
+	statusEvents     []event.StatusEvent
+	pruneEvents      []event.PruneEvent
+	deleteEvents     []event.DeleteEvent
+	waitEvents       []event.WaitEvent
+	errorEvent       event.ErrorEvent
+	actionGroupEvent []event.ActionGroupEvent
+}
+
+func (c *countingFormatter) FormatApplyEvent(e event.ApplyEvent) error {
+	c.applyEvents = append(c.applyEvents, e)
+	return nil
+}
+
+func (c *countingFormatter) FormatStatusEvent(e event.StatusEvent) error {
+	c.statusEvents = append(c.statusEvents, e)
+	return nil
+}
+
+func (c *countingFormatter) FormatPruneEvent(e event.PruneEvent) error {
+	c.pruneEvents = append(c.pruneEvents, e)
+	return nil
+}
+
+func (c *countingFormatter) FormatDeleteEvent(e event.DeleteEvent) error {
+	c.deleteEvents = append(c.deleteEvents, e)
+	return nil
+}
+
+func (c *countingFormatter) FormatWaitEvent(e event.WaitEvent) error {
+	c.waitEvents = append(c.waitEvents, e)
+	return nil
+}
+
+func (c *countingFormatter) FormatErrorEvent(e event.ErrorEvent) error {
+	c.errorEvent = e
+	return nil
+}
+
+func (c *countingFormatter) FormatActionGroupEvent(e event.ActionGroupEvent, _ []event.ActionGroup, _ stats.Stats,
+	_ Collector) error {
+	c.actionGroupEvent = append(c.actionGroupEvent, e)
+	return nil
+}
--- a/pkg/printers/events/formatter.go
+++ b/pkg/printers/events/formatter.go
@@ -13,6 +13,7 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/cli-utils/pkg/object"
 	"sigs.k8s.io/cli-utils/pkg/print/list"
+	"sigs.k8s.io/cli-utils/pkg/print/stats"
 )
 
 func NewFormatter(ioStreams genericclioptions.IOStreams,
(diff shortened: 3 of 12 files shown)


pkg/print/common/errors.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2021 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package common
+
+import "fmt"
+
+// ResultError is returned from printers when the apply/destroy operations completed, but one or
+// more resources either failed apply/prune/delete, or failed to reconcile.
+type ResultError struct {
```
karlkfi, line 10: I don't love the name `common.ResultError` (too generic) but I don't really have a better suggestion.

karlkfi:
You only added this to the list printer. Should we use it in the other printers?

mortent:
We should probably add it to the table printer as well, but it should be part of a larger refactor to make sure we can mostly use the same logic for this in both printers. I would like to do this in a separate PR.

pkg/print/common/errors.go
mortent: Agree. I have tried to come up with something better, but haven't found anything yet. Maybe something like `ActuationError` would be better?

karlkfi: That would make the `failedActuateSum` variable name confusing. Does actuation include waiting for reconciliation or not?

pkg/print/common/errors.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2021 The Kubernetes Authors.
```
haiyanmeng, line 1: 2021 -> 2022

pkg/print/common/errors.go
```diff
@@ -0,0 +1,34 @@
...
+	ReconcileTimeout int
+	ReconcileFailed  int
+}
+
+func (a *ResultError) Error() string {
+	failedActuateSum := a.ApplyFailed + a.PruneFailed + a.DeleteFailed
+	failedReconcileSum := a.ReconcileTimeout + a.ReconcileFailed
+	switch {
+	case failedActuateSum > 0 && failedReconcileSum > 0:
+		return fmt.Sprintf("%d resources failed, %d resources failed to reconcile before timeout",
```
haiyanmeng, line 24: `%d resources failed, %d resources failed to reconcile before timeout` -> `%d resources failed to actuate, %d resources failed to reconcile before timeout`

pkg/print/common/errors.go
```diff
@@ -0,0 +1,34 @@
...
+			failedActuateSum, failedReconcileSum)
+	case failedActuateSum > 0:
+		return fmt.Sprintf("%d resources failed", failedActuateSum)
```
haiyanmeng, line 27: `failed` -> `failed to actuate`

pkg/print/common/errors.go
```diff
@@ -0,0 +1,34 @@
...
+	case failedReconcileSum > 0:
+		return fmt.Sprintf("%d resources failed to reconcile before timeout",
+			failedReconcileSum)
+	default:
+		return "actuation and reconcile failed"
```
haiyanmeng, line 32: The `default` case should mean that all the resources are reconciled, right?

pkg/print/list/base_test.go
```diff
@@ -0,0 +1,241 @@
+// Copyright 2021 The Kubernetes Authors.
```
haiyanmeng, line 1: 2021 -> 2022

karlkfi:
It occurs to me that this isn't a "print error". The printer itself didn't really fail. It's really just an aggregate summary of the events.

What if instead of baking this into the BaseListPrinter we add a print wrapper that collects the stats into an object passed in by the user?

For example:

```
type StatsPrinter struct {
	Printer Printer
	Stats EventStats
}

type EventStats struct {
	ApplyFailed  int
	PruneFailed  int
	DeleteFailed int

	ReconcileTimeout int
	ReconcileFailed  int
}

var _ Printer = &StatsPrinter{}

func (sp *StatsPrinter) Print(ch <-chan event.Event, previewStrategy common.DryRunStrategy, printStatus bool) error {
	outerCh = ch
	innerCh := make(chan event.Event)
	go func() {
		for e := range outerCh {
			# add stats to sp.Stats
			innerCh <- e
		}
	}()
	return sp.Printer.Print(innerCh, previewStrategy, printStatus)
}
```

Then the user can optionally wrap their printer impl in a `StatsPrinter` to collect the stats into the `EventStats` object.

```
printer := &StatsPrinter{ Printer: myCustomPrinter }
err := printer.Print(ch, previewStrategy, printStatus)
if (err != nil) {
	return err
}
stats := printer.Stats
# print stats as-needed
```

pkg/print/common/errors.go
karlkfi: This would be an output change to kpt. Without this, the change is reverse compatible.

I'm not sure we want to expose the "actuate" or "actuation" terms to users yet. We'll want to be very deliberate about it and what it means if/when we do.

pkg/print/common/errors.go
karlkfi: The default case for an error is still an error. Tho I don't know what it would mean in this context... seems invalid/unreachable.

karlkfi:
A wrapping printer like above should be usable for all printer types.

mortent:
@karlkfi That is an interesting idea. I'll try that out in a separate PR.

With that implementation, presumably we would no longer consider resources failing actuation or reconciliation as an error in the `BasePrinter` and `TablePrinter` right? If users want to know the result, they would just wrap the printer with the StatsPrinter.

karlkfi:
> With that implementation, presumably we would no longer consider resources failing actuation or reconciliation as an error in the BasePrinter and TablePrinter right? If users want to know the result, they would just wrap the printer with the StatsPrinter.

Correct. It would be a behavioral change, but not an API change.

karlkfi:
Another alternate way to do this would be to move it into the pipieline and have the baserunner do the aggregation and return an ErrorEvent at the end.

mortent:
@karlkfi Created a POC for this: https://github.com/kubernetes-sigs/cli-utils/pull/503

I think we should keep it out of the pipeline if we can. It is already pretty complicated.

pkg/print/common/errors.go
mortent: I was thinking yes in this case, but I agree that it could be confusing.

pkg/print/common/errors.go
mortent: I also don't think we should change the language here.

mortent:
We have looked at alternative solutions to address this, but we haven't come up with an alternative that is clearly better than this approach. Since this is fix with the least impact that addresses the problem, I think we should take this approach. We can revisit it later if we find a better approach.

pkg/print/common/errors.go
```diff
@@ -0,0 +1,36 @@
...
+	ApplyFailed  int
+	PruneFailed  int
+	DeleteFailed int
+
+	ReconcileTimeout int
+	ReconcileFailed  int
+}
+
+func (a *ResultError) Error() string {
+	failedActuateSum := a.ApplyFailed + a.PruneFailed + a.DeleteFailed
```
karlkfi, line 20: Are clients going to need these sums? Should they be a method on the type?

mortent: The idea was to not add additional methods for now. We can always add more methods if needed.

mortent:
@karlkfi I have updated the PR so it updates all printers to return the `ResultError`.

pkg/printers/events/formatter.go
```diff
@@ -129,19 +128,23 @@ func (ef *formatter) FormatActionGroupEvent(
...
+		ps := s.PruneStats
 		ef.print("%d resource(s) pruned, %d skipped, %d failed", ps.Pruned, ps.Skipped, ps.Failed)
 	}
 
 	if age.Action == event.DeleteAction &&
 		age.Type == event.Finished &&
 		list.IsLastActionGroup(age, ags) {
-		ef.print("%d resource(s) deleted, %d skipped", ds.Deleted, ds.Skipped)
+		ds := s.DeleteStats
+		ef.print("%d resource(s) deleted, %d skipped, %d failed", ds.Deleted, ds.Skipped, ds.Failed)
```
karlkfi, line 139: do we want to say "failed to delete"? "failed" by itself feels a little ambiguous. same with failed to reconcile below.

pkg/printers/table/collector.go
```diff
@@ -211,6 +216,9 @@ func (r *ResourceStateCollector) processApplyEvent(e event.ApplyEvent) {
 		klog.V(4).Infof("%s apply event not found in ResourceInfos; no processing", identifier)
 		return
 	}
+	if e.Error != nil {
+		previous.Error = e.Error
```
karlkfi, line 220: does this mean we only show the first error?
are there cases of multiple errors after validation?

pkg/printers/table/collector.go
```diff
@@ -285,6 +296,36 @@ func (r *ResourceStateCollector) LatestState() *ResourceState {
...
+// as a stats.Stats object.
+func (r *ResourceStateCollector) Stats() stats.Stats {
+	var s stats.Stats
+	for _, res := range r.resourceInfos {
+		switch res.ResourceAction {
+		case event.ApplyAction:
+			if res.Error != nil {
+				s.ApplyStats.IncFailed()
+			} else {
+				s.ApplyStats.Inc(res.ApplyOpResult)
```
karlkfi, line 309: are the operations not set when it's an error?

pkg/printers/testing/common.go
```diff
@@ -0,0 +1,205 @@
...
+	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/cli-utils/pkg/object"
+	printcommon "sigs.k8s.io/cli-utils/pkg/print/common"
+	"sigs.k8s.io/cli-utils/pkg/print/stats"
+	"sigs.k8s.io/cli-utils/pkg/printers/printer"
+)
+
+type PrinterFactoryFunc func() printer.Printer
+
+func PrintResultErrorTest(t *testing.T, f PrinterFactoryFunc) {
```
karlkfi, line 23: I see PrintResultErrorTest is used by the list and table, but I don't see one for the json printer.

pkg/printers/testing/common.go
```diff
@@ -0,0 +1,205 @@
+// Copyright 2022 The Kubernetes Authors.
```
karlkfi, line 1: pkg/printers/testing should probably be pkg/printers/testutil to match the other few testutil packages

karlkfi (changes requested):
thanks for adding the Result error to the other printers. just a few nits and questions left.

pkg/printers/table/collector.go
mortent: We currently only attempt once, so we should only get one event pertaining to apply/prune of a specific resource. If we were to attempt more than once, I think it would make sense to use the error from the latest attempt.

pkg/printers/testing/common.go
mortent: It runs the tests for the `table` and the base printer. Both the `events` and the `json` printer uses the base printer (just uses different formatters), so they are both being tested.

pkg/printers/table/collector.go
mortent: They should be set to the default value for the operation type. I have updated the code so we handle this in the `Inc` function rather than in a separate `else` block.

pkg/printers/testing/common.go
karlkfi: gotcha

pkg/printers/table/collector.go
karlkfi: huh... looks like i actually made an operation for ReconcileFailed, but the other event operations don't have them. I wonder if we should add them in anther PR, just to be consistent. Then we could just use the operation for stats collection.

pkg/printers/table/collector.go
karlkfi: eh. ok. looks weird, but it's not a big deal.

pkg/printers/table/collector.go
mortent: Hmm.. I'm not sure what is the best approach here. Having a separate `Apply/Prune/DeleteFailed` operation means we potentially have two place to look. So is there a way those operations can fail without the event containing an error (I don't think so..) and is there any situation where we might want to include an error in the event, even if the operation is something else than `Failed`?
I don't feel very strongly either way.

pkg/printers/events/formatter.go
mortent: fixed.

karlkfi (approved):
/lgtm

PR #501, opened 2022-01-05
fix: Remove preview/dry-run type from events output

mortent:
The reasons for this change:
* The output is just repeating the preview/dry-run strategy that the user provided, and the value is the same for every line. Thus, it ends up being very verbose.
* We don't currently print it consistently for all printers. It is only used in the output for the events output format.
* We currently uses the term `preview`, rather than the more common term `dry-run` used in kubectl. Making is possible to customize the term here seems a bit overkill (https://github.com/kubernetes-sigs/cli-utils/pull/499).

Removing it seems like a good solution to keep the printers simple, but still allow users to build their own printers if they want to.

### DIFF
--- a/cmd/preview/cmdpreview.go
+++ b/cmd/preview/cmdpreview.go
@@ -163,6 +163,15 @@ func (r *PreviewRunner) RunE(cmd *cobra.Command, args []string) error {
 		})
 	}
 
+	// Print the preview strategy unless the output format is json.
+	if r.output != printers.JSONPrinter {
+		if drs.ServerDryRun() {
+			fmt.Println("Preview strategy: server")
+		} else {
+			fmt.Println("Preview strategy: client")
+		}
+	}
+
 	// The printer will print updates from the channel. It will block
 	// until the channel is closed.
 	printer := printers.GetPrinter(r.output, r.ioStreams)
--- a/examples/alphaTestExamples/helloapp.md
+++ b/examples/alphaTestExamples/helloapp.md
@@ -174,11 +174,11 @@ Run preview to check which commands will be executed
 ```
 kapply preview $BASE | tee $OUTPUT/status
 
-expectedOutputLine "3 resource(s) applied. 3 created, 0 unchanged, 0 configured, 0 failed (preview)"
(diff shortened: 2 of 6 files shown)


mortent:
/hold
Make sure everyone who might want to provide input here has a chance to do so.

karlkfi:
Why don't we just change it to "dry-run" everywhere?

"Preview" never made sense to me anyway. It doesn't actually print the diff, which is what I actually want out of a preview.

mortent:
> Why don't we just change it to "dry-run" everywhere?

I did consider it, but because of the other reasons listed in the description, I think it is better to not print it at all. Printing out the same value for every line seems unnecessary. And the value is just what the user provided (server-side vs client-side dry-run/preview).

There is some history behind the `preview` term, in particular the `kapply` binary has a separate `preview` command separate from `apply`. I agree with your comment that is currently doesn't do what I would expect from a preview and in fact just does a dry-run.

karlkfi:
How about we instead print a line at the beginning that says something like `Dry-run mode enabled.`?

mortent:
> How about we instead print a line at the beginning that says something like `Dry-run mode enabled.`?

Yeah, we could do that, but I don't think it should be part of the printer. If so, we should print this from the cobra command before we give control to the printer.

karlkfi:
> Yeah, we could do that, but I don't think it should be part of the printer. If so, we should print this from the cobra command before we give control to the printer.

sgtm

karlkfi:
> we should print this from the cobra command before we give control to the printer.

Want to make another PR for this?

ash2k:
LGTM

mortent:
Updated the PR to print the preview strategy on a separate line before the printer output:
* It still uses the `preview` rather than the `dry-run` terminology. I think that makes sense as long as we have a separate preview command.
* It doesn't print it if the output format is json. This is to avoid non-json output that would cause problems for users parsing the output. I don't think it is worth including a separate line with the preview strategy in json format, but open to input here.

karlkfi:
/lgtm

Are you going to modify kpt too to print its own dry-run line?

karlkfi (approved):

karlkfi (approved):
/lgtm

PR #503, opened 2022-01-06
[WIP] Separate out collection of stats from the printer

mortent:
This is a POC for how we can separate collection of stats from an apply/destroy operation from the printers. It splits out the code used for this within the `BaseListPrinter` and make the `Stats` type available for other use-cases. 
It makes it available in the `stats.Collector`, which can be used as in this example:
```golang
eventChannel := a.Run(ctx, inv, objs, apply.Options{})

statsCollector := stats.NewCollector()
printer := printers.GetPrinter(r.output, r.ioStreams)
if err := printer.Print(statsCollector.Run(eventChannel), common.DryRunNone, r.printStatusEvents); err != nil {
  // handle err
}

s := statsCollector.Stats
```

### DIFF
--- a/pkg/print/list/base.go
+++ b/pkg/print/list/base.go
@@ -4,11 +4,10 @@
 package list
 
 import (
-	"fmt"
-
 	"sigs.k8s.io/cli-utils/pkg/apply/event"
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/cli-utils/pkg/object"
+	"sigs.k8s.io/cli-utils/pkg/stats"
 )
 
 type Formatter interface {
@@ -21,10 +20,7 @@ type Formatter interface {
 	FormatActionGroupEvent(
 		age event.ActionGroupEvent,
 		ags []event.ActionGroup,
-		as *ApplyStats,
-		ps *PruneStats,
-		ds *DeleteStats,
-		ws *WaitStats,
+		stats stats.Stats,
 		c Collector,
 	) error
 }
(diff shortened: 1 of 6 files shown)


pkg/stats/stats.go
```diff
@@ -0,0 +1,161 @@
...
+
+type Collector struct {
+	Stats Stats
+}
+
+func NewCollector() *Collector {
+	return &Collector{}
+}
+
+func (n *Collector) Run(eventChannel <-chan event.Event) <-chan event.Event {
```
karlkfi, line 20: How about "Collect" as the func name?

pkg/print/list/base.go
```diff
@@ -149,26 +54,20 @@ func (sc *StatusCollector) LatestStatus() map[object.ObjMetadata]event.StatusEve
...
-	pruneStats := &PruneStats{}
-	deleteStats := &DeleteStats{}
-	waitStats := &WaitStats{}
+	var statsCollector stats.Stats
 	statusCollector := &StatusCollector{
 		latestStatus: make(map[object.ObjMetadata]event.StatusEvent),
 	}
 	formatter := b.FormatterFactory(previewStrategy)
 	for e := range ch {
+		statsCollector.Handle(e)
```
karlkfi, line 63: Why collect these if we're not gonna return them to the caller?
Doesn't that mean the user has to wrap their own and collect twice?

pkg/stats/stats.go
mortent: Yeah, that is a better name. I have updated the PR.

pkg/print/list/base.go
mortent: Yeah, this means it will be collected twice if the users wrap it. 

So in this case we do need the stats in the printer itself, since we do print some of the stats as part of the output. We can't handle that from outside the printer without making significant changes to the output.

mortent: So I think we have two options here:
* Make it the responsibility of the printers to collect stats and return it to the caller. This is useful for printers that needs to use the same stats anyway, but less useful for printers that doesn't (i.e. `table`). The change in https://github.com/kubernetes-sigs/cli-utils/pull/500 returns this in an error since that can be done without changing the `printer` interface, but we could potentially returns the stats as a separate return value.
* Make collecting stats separate from the printing process. The separation of concerns is somewhat compelling, but it does make for some double work if the implementation of a printer itself requires the stats. This is currently the case for the `json` and the `events` printer. We could revisit the output format such that the summaries are only printed at the end, and thus could be handled outside of the print code itself. But other printer implementations might still want to include information that requires stats.

I think I prefer the latter, as I agree that collecting the stats doesn't seem like something that should be the responsibility of the printer. I don't think collecting it twice would be a significant overhead.

PR #504, opened 2022-01-06
json printer should only print stats after last action group per type

mortent:
The json printer currently outputs stats after each action group. This is not consistent with the `events` printer and also not correct since the stats are not reset after each group. This updates the `json` printer to behave the same as the `events` printer.


karlkfi (approved):
/lgtm

This is fine for a first pass to make these more consistent. But please follow up with a PR to move the summary printing to the end of the Print function, to avoid the hacky IsLastActionGroup check.

PR #506, opened 2022-01-07
fix: Handle more validation errors as field errors

karlkfi:
- Using field.Error allows more errors to be wrapped as a
  ValidationError, instead of interrupting validation and exiting
  early.
- Add explicit validation for kind (clearer error).
- Move NestedField from testutil to object, so it can be used at
  runtime.


mortent (approved):
/approve

karlkfi:
@ash2k can you take a look at this? It's extracted from https://github.com/kubernetes-sigs/cli-utils/pull/505 which you already approved.

pkg/object/field.go
```diff
@@ -0,0 +1,108 @@
...
+		Type:     field.ErrorTypeNotFound,
+		Field:    FieldPath(fieldPath),
+		BadValue: value,
+		Detail:   "",
+	}
+}
+
+// Simplistic jsonpath formatter, just for NestedField errors.
+func FieldPath(fieldPath []interface{}) string {
+	path := ""
```
ash2k, line 81: nit: could use `strings.Builder` here for concatenation.

ash2k (approved):
/lgtm

pkg/object/field.go
karlkfi: good catch: https://github.com/kubernetes-sigs/cli-utils/pull/507

PR #507, opened 2022-01-08
fix: Clean up object.FieldPath

karlkfi:
- Use string.Builder to be slightly more efficient
- Format invalid types with Go syntax, wrapped with brackets
- Test that invalid types don't panic


ash2k (approved):
/lgtm

mortent (approved):
/approve

PR #508, opened 2022-01-11
feat: MultiError for invalid annotations

karlkfi:
- Move Validator to pkg/object/validation
- Replace ValidationError with validation.Error
- Replace MultiValidationError with generic MultiError
- Update Validator & SortObjs to use MultiError
- Add ResourceReferenceFromObjMetadata
- Rename NewResourceReference -> ResourceReferenceFromUnstructured
- Delete duplicate ResourceReference.ObjMetadata()
- Modify some error messages for consistency and clarity

BREAKING CHANGE: apply-time-mutation namespace required for namespace-scoped resources


pkg/object/dependson/strings.go
```diff
@@ -114,8 +114,8 @@ func ParseObjMetadata(objStr string) (object.ObjMetadata, error) {
 	fields := strings.Split(objStr, fieldSeparator)
 
 	if len(fields) != numFieldsClusterScoped && len(fields) != numFieldsNamespacedScoped {
-		return obj, fmt.Errorf("too many fields (expected %d or %d): %q",
-			numFieldsClusterScoped, numFieldsNamespacedScoped, objStr)
+		return obj, fmt.Errorf("expected %d or %d fields, found %d: %q",
```
mortent, line 117: Is this a very helpful error message? Maybe more helpful to just print the value we tried to parse and then say that it couldn't be recognized as neither a cluster-scoped or namespace-scoped reference.

pkg/object/graph/depends.go
```diff
@@ -80,73 +96,140 @@ func ReverseSortObjs(objs object.UnstructuredSet) ([]object.UnstructuredSet, err
...
-			if err != nil {
-				// TODO: fail task if parse errors?
-				klog.V(3).Infof("failed to add edges from: %s: %s", id, err)
-				return
+	}
+}
+
+// addApplyTimeMutationEdges updates the graph with edges from objects
+// with an explicit "apply-time-mutation" annotation.
+// The objs and ids must match in order and length (optimization).
```
mortent, line 109: Is this to avoid having to translate from unstructured to ObjMetadata many times?

pkg/object/graph/graph.go
```diff
@@ -43,6 +45,17 @@ func (g *Graph) AddVertex(v object.ObjMetadata) {
 	}
 }
 
+// getVertices returns an unsorted set of unique vertices in the graph.
```
mortent, line 48: Nit: The g should be capitalized.

mortent (commented):
Just a few small comments, but looks pretty good.

pkg/object/dependson/strings.go
karlkfi: Mostly i changed it because the message was wrong for the test i wrote that had no underscores. There were too few segments.

Both before and after the value itself is being appended.  For example, see https://github.com/kubernetes-sigs/cli-utils/pull/508/files#diff-84e8cf5b1fe1c9df4d3447f4414ddbcec8f820f7dbc0ee3cb56aaed1754f4d59R734-R737

pkg/object/graph/depends.go
karlkfi: Yes. It's just a minor optimization and only affects private functions.

I considered passing a map of id->obj, but then you lose the ordering, which makes testing painful or requires sorting again, which is less efficient.

That said, the sets are all being sorted at the end of SortObjs, so the end result would be mostly the same. The gotcha is that if the sort only sorts by type, then the output will be different if the input is different (ex: two objects of the same type in a different order).

pkg/object/graph/graph.go
karlkfi: good catch.

pkg/multierror/multierror.go
```diff
@@ -0,0 +1,91 @@
+// Copyright 2021 The Kubernetes Authors.
```
haiyanmeng, line 1: 2021 -> 2022

pkg/object/mutation/types.go
```diff
@@ -105,6 +105,16 @@ func NewResourceReference(obj *unstructured.Unstructured) ResourceReference {
 	}
 }
 
+// NewResourceReference returns the object as a ResourceReference
```
haiyanmeng, line 108: NewResourceReference -> ResourceReferenceFromObjMetadata

pkg/object/validation/error.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2021 The Kubernetes Authors.
```
haiyanmeng, line 1: 2021 -> 2022

pkg/object/validation/validate.go
```diff
@@ -0,0 +1,95 @@
+// Copyright 2021 The Kubernetes Authors.
```
haiyanmeng, line 1: 2021 -> 2022

pkg/object/validation/validate_test.go
```diff
@@ -0,0 +1,274 @@
+// Copyright 2021 The Kubernetes Authors.
```
haiyanmeng, line 1: 2021 -> 2022

PR #510, opened 2022-01-19
fix: Update ObjMetadataSet.Hash to not error

karlkfi:
Hash32.Write never returns an error: https://pkg.go.dev/hash#pkg-types

Hash functions should never return errors in general, because they're commonly used in String() and Equals() funcs, which don't error.


mortent (approved):
/approve

PR #511, opened 2022-01-19
fix: Merge task runner impls into taskStatusRunner

karlkfi:
- Delete taskRunner & baseRunner
- Add short strings for ResourceAction enums
- Add detail to task errors
- Modify taskStatusRunner.Run to take TaskContext instead of
  eventChannel and resouceCache.
- Update runner tests to use fakePoller

BREAKING CHANGE: Changed taskStatusRunner.Run function signature


seans3 (approved):
/approve

PR #512, opened 2022-01-19
fix: Make CyclicDependencyError edges stable

karlkfi:
- Add sorting for Edges & Verticies
- Improve tests that contain CyclicDependencyError


mortent (approved):
/approve

PR #513, opened 2022-01-20
fix: Improve solver tests

karlkfi:
- Make FakeRESTMapper comparable
- Add Comparer for WaitTask
- Add Asserter to allow pre-configuring comparison options for Equal/NotEqual
- Update solver tests to use Asserter for task list comparison (more actionable errors)

### DIFF
--- a/pkg/apply/solver/solver_test.go
+++ b/pkg/apply/solver/solver_test.go
@@ -4,10 +4,11 @@
 package solver
 
 import (
-	"reflect"
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/stretchr/testify/assert"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/cli-utils/pkg/apply/filter"
@@ -105,6 +106,12 @@ metadata:
 )
 
 func TestTaskQueueBuilder_AppendApplyWaitTasks(t *testing.T) {
+	// Use a custom Asserter to customize the comparison options
+	asserter := testutil.NewAsserter(
+		cmpopts.EquateErrors(),
+		waitTaskComparer(),
+	)
+
 	testCases := map[string]struct {
 		applyObjs     []*unstructured.Unstructured
 		options       Options
@@ -113,7 +120,7 @@ func TestTaskQueueBuilder_AppendApplyWaitTasks(t *testing.T) {
 	}{
 		"no resources, no tasks": {
 			applyObjs:     []*unstructured.Unstructured{},
-			expectedTasks: []taskrunner.Task{},
+			expectedTasks: nil,
 		},
 		"single resource, one apply task, one wait task": {
 			applyObjs: []*unstructured.Unstructured{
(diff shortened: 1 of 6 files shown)


pkg/apply/taskrunner/task.go
```diff
@@ -77,8 +78,25 @@ type WaitTask struct {
 	mu sync.RWMutex
 }
 
+// Equal returns true if the public fields are equal (private fields ignored).
+// Implements the "(T) Equal(T) bool" interface for cmp.Equal:
+// https://pkg.go.dev/github.com/google/go-cmp/cmp#Equal
+func (w *WaitTask) Equal(other *WaitTask) bool {
```
ash2k, line 84: I'm fine with adding such method to testing code, but I think it'd be better to use a [comparer](https://pkg.go.dev/github.com/google/go-cmp/cmp#Comparer) for `WaitTask`. Ideally production code shouldn't have any scaffolding for testing that otherwise would not be there.

karlkfi: huh, i didn't know comparers existed.

The problem with the AssertEqual function interface is that it ends with a vararg, making it hard to add optional cmpopts to the EqualMatcher it wraps... I could use the EqualMatcher directly, but that makes it several more lines of code to use. I considered having a global variable to set default cmpopts for the Equalmatcher, but that just seems sloppy.

If you can think of a clean way to make the options optional, I'm all ears.

karlkfi: Alright, I came up with a way to handle options with an optional parent Asserter struct... 

It's a tad confusing to look at, but allows the existing global functions to delegate to a global DefaultAsserter. Then if you need to use custom options you have to construct your own Asserter and use its methods instead of the global functions.

karlkfi:
@ash2k I think I solved your concern, please take a look.

pkg/testutil/asserter.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Equal fails the test if the actual value does not deeply equal the
+// expected value. Prints a diff on failure.
+func (a *Asserter) Equal(t *testing.T, expected, actual interface{}, msgAndArgs ...interface{}) {
+	t.Helper() // print the caller's file:line, instead of this func, on failure
+	matcher := a.EqualMatcher(expected)
+	match, err := matcher.Match(actual)
+	if err != nil {
+		t.Errorf("errored testing equality: %s", err)
```
ash2k, line 47: We don't need to proceed with checking `match` if there was an error, right?
```suggestion
		t.Errorf("errored testing equality: %v", err)
		return
```

pkg/testutil/asserter.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// NotEqual fails the test if the actual value deeply equals the
+// expected value. Prints a diff on failure.
+func (a *Asserter) NotEqual(t *testing.T, expected, actual interface{}, msgAndArgs ...interface{}) {
+	t.Helper() // print the caller's file:line, instead of this func, on failure
+	matcher := a.EqualMatcher(expected)
+	match, err := matcher.Match(actual)
+	if err != nil {
+		t.Errorf("errored testing equality: %s", err)
```
ash2k, line 68:
```suggestion
		t.Errorf("errored testing equality: %v", err)
		return
```

ash2k (commented):
I like this approach! Left a few minor suggestions.

karlkfi:
Resolved and rebased

pkg/apply/solver/solver_test.go
```diff
@@ -720,72 +729,49 @@ func TestTaskQueueBuilder_AppendPruneWaitTasks(t *testing.T) {
...
+			return y == nil
+		}
+		if y == nil {
+			return false
+		}
+		return x.TaskName == y.TaskName &&
+			x.Ids.Hash() == y.Ids.Hash() && // exact order match
+			x.Condition == y.Condition &&
+			x.Timeout == y.Timeout &&
+			cmp.Equal(x.Mapper, y.Mapper)
```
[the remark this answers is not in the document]
karlkfi: We don't really have to, but it's in there and now comparable, so it at least confirms that the same mapper was injected into multiple tasks.

mortent (approved):
/approve

ash2k (approved):
/lgtm

PR #514, opened 2022-01-20
capture the apply status in the inventory object

Liujingfang1:
Capture the apply status for individual resources after each apply.
If the inventory object supports status field, it is updated
after setting the inventory list at then of one apply process.


pkg/inventory/inventory-client.go
```diff
@@ -513,3 +541,46 @@ func (cic *ClusterInventoryClient) toInfo(obj *unstructured.Unstructured) (*reso
 func (cic *ClusterInventoryClient) helperFromInfo(info *resource.Info) *resource.Helper {
 	return resource.NewHelper(info.Client, info.Mapping)
 }
+
+func (cic *ClusterInventoryClient) updateStatus(obj *unstructured.Unstructured, dryRun common.DryRunStrategy) error {
```
mortent, line 545: This doesn't seem right to me. Similar to how we put the decision on how/where the inventory list is added to a specific resource type in an type-specific implementation of the `Inventory` interface, this functionality should also belong in the type-specific implementation.

mortent (commented):
So I think providing the applystatus as input to the `Replace` function in the `InventoryInterface` is a reasonable change. Being able to capture the result of an apply operation into a KRM resource seems generally pretty useful.

I think the challenge here is that putting this into a specific place in the `Status` object of a resource seems very specific. We should encapsulate this into the resource-specific implementation of the `Inventory` interface. I realize that the implementation of the InventoryClient currently makes that difficult, as it doesn't allow updates to the Status object. I think we can make some improvements to the current code that will both allow that if needed and clean things up a bit.

pkg/inventory/inventory-client.go
Liujingfang1: The `Inventory` interface handles where and how the resource list is stored, it doesn't handle how the inventory object is applied to the cluster.
The `InventoryClient` handles how it is applied to the cluster. `updateStatus` is part of applying the inventory object to the cluster.

pkg/inventory/inventory-client.go
```diff
@@ -113,24 +124,32 @@ func (cic *ClusterInventoryClient) Merge(localInv InventoryInfo, objs object.Obj
...
 		klog.V(4).Infof("num objects to prune: %d", len(pruneIds))
 		klog.V(4).Infof("num merged objects to store in inventory: %d", len(unionObjs))
 		wrappedInv := cic.InventoryFactoryFunc(clusterInv)
-		if err = wrappedInv.Store(unionObjs); err != nil {
+		if err = wrappedInv.Store(unionObjs, applyStatus); err != nil {
 			return pruneIds, err
 		}
+		clusterInv, err := wrappedInv.GetObject()
+		if objs.Equal(clusterObjs) {
+			klog.V(4).Infof("applied objects same as cluster inventory: do nothing")
```
karlkfi, line 147: This log isn't accurate any more if the status is still being updated.

pkg/inventory/inventory-client.go
```diff
@@ -113,24 +124,32 @@ func (cic *ClusterInventoryClient) Merge(localInv InventoryInfo, objs object.Obj
...
+		applyStatus := getApplyStatus(pruneIds, unionObjs)
 		klog.V(4).Infof("num objects to prune: %d", len(pruneIds))
 		klog.V(4).Infof("num merged objects to store in inventory: %d", len(unionObjs))
 		wrappedInv := cic.InventoryFactoryFunc(clusterInv)
-		if err = wrappedInv.Store(unionObjs); err != nil {
+		if err = wrappedInv.Store(unionObjs, applyStatus); err != nil {
 			return pruneIds, err
 		}
+		clusterInv, err := wrappedInv.GetObject()
+		if objs.Equal(clusterObjs) {
```
karlkfi, line 146: this if statement can probably be inverted to wrap the code below so there's only one call to updateStatus at the end

pkg/inventory/inventory-client.go
```diff
@@ -158,30 +180,36 @@ func (cic *ClusterInventoryClient) Replace(localInv InventoryInfo, objs object.O
...
 	clusterInv, err := cic.GetClusterInventoryInfo(localInv, dryRun)
 	if err != nil {
 		return fmt.Errorf("failed to read inventory from cluster: %w", err)
 	}
-	clusterInv, err = cic.replaceInventory(clusterInv, objs)
+	clusterInv, err = cic.replaceInventory(clusterInv, objs, applyStatus)
 	if err != nil {
 		return err
 	}
+	if objs.Equal(clusterObjs) {
```
karlkfi, line 191: this if statement can probably be inverted to wrap the code below so there's only one call to updateStatus at the end

pkg/inventory/inventory-client.go
```diff
@@ -158,30 +180,36 @@ func (cic *ClusterInventoryClient) Replace(localInv InventoryInfo, objs object.O
...
+		klog.V(4).Infof("applied objects same as cluster inventory: do nothing")
```
karlkfi, line 192: This log isn't accurate any more if the status is still being updated.

pkg/inventory/inventory-client.go
karlkfi: The problem this is solving is that updating status must be a separate API call because the status is a subresource.

Unfortunately, the current abstraction has the InventoryClient in cli-utils making the API calls, which means it's not possible for the Inventory or InventoryInfo implementation to decide whether one or two API calls is needed.

This is the kludge I was talking about. To work around it would require more significant changes to these interfaces, like pushing the `applyInventoryObj` impl out of the InventoryClient and into something that can be injected.

pkg/kstatus/status/apply_status.go
```diff
@@ -0,0 +1,41 @@
+package status
+
+const (
```
karlkfi, line 3: Recommend replacing all these strings with enums. You can use go-stringer to impl the String() methods and then optionally provide short string values in comments.

Ex: https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/apply/event/event.go#L106

karlkfi (changes requested):
Suggested some changes in comments. I don't love how this works, but I don't have a better suggestion for how to make it cleaner. I'll wait for sign off from @mortent and @seans3

karlkfi:
Please amend the commit messages to have fix or feat prefixes. I want to try to start following https://www.conventionalcommits.org/en/v1.0.0/#summary to make change logs easier to produce. If there's a breaking change, add a BREAKING CHANGE line at the end with detail.

pkg/inventory/inventory-client.go
Liujingfang1: Removed this log as inverting the if statement.

pkg/inventory/inventory-client.go

pkg/inventory/inventory-client.go
Liujingfang1: Done.

pkg/kstatus/status/apply_status.go
Liujingfang1: Done.

pkg/inventory/inventory-client.go
Liujingfang1: Done.

seans3 (changes requested):
This appears to change the behavior of the pruning/inventory algorithm. Currently, we only store in the inventory object references to objects which have successfully been applied. While there are nuances, conceptually the inventory object should only contain references to objects which currently exist in the cluster (these are the only object we need to prune). This code appears to store every object which was attempted to be applied with its status, including objects that have failed and do not exist in the cluster. This feels like a significant change. I'm not sure this works.

Also, how would migration work? What about inventory objects which do not currently have the object status, and would have a different set of references that the newer inventory objects?

Finally, I'm not sure why we need this significant change. If this is code that is only necessary for one `cli-utils` client, then I don't think it belongs in the here. I do not think this status field is needed to implement the pruning functionality, since it already works. What is the benefit for all the `cli-utils` clients?

pkg/apply/task/inv_set_task.go
```diff
@@ -118,3 +122,23 @@ func (i *InvSetTask) Cancel(_ *taskrunner.TaskContext) {}
...
 // StatusUpdate is not supported by the InvSetTask.
 func (i *InvSetTask) StatusUpdate(_ *taskrunner.TaskContext, _ object.ObjMetadata) {}
+
+func (i InvSetTask) getApplyStatus(taskContext *taskrunner.TaskContext) map[object.ObjMetadata]status.ApplyStatus {
+	applyStatus := map[object.ObjMetadata]status.ApplyStatus{}
+	for _, obj := range taskContext.SuccessfulApplies() {
+		applyStatus[obj] = status.ApplySucceeded
+	}
+	for _, obj := range i.PrevInventory.Intersection(taskContext.FailedApplies()) {
+		applyStatus[obj] = status.ApplyFailed
```
seans3, line 132: We should only store references to objects which exist in the cluster. We store objects which failed if they had previously succeeded.

pkg/inventory/inventory-client.go
Liujingfang1: As discussed offline. I add this method as the minimum change using the current interface. Later, we should improve the interface for the inventory object. Then applying an inventory object is handled as part of the inventory interface.

pkg/apply/task/inv_set_task.go
Liujingfang1: The inventory set is not affected. For the ConfigMap inventory, the passed in apply/reconcile status is not used for now. For the inventory object in Kpt, the apply/reconcile status is stored in `.status`.

Liujingfang1:
@mortent @karlkfi @seans3 Please take another look at the change. Let me know if there is any other concern that you have.

pkg/apply/task/inv_set_task.go
```diff
@@ -124,3 +128,36 @@ func (i *InvSetTask) Cancel(_ *taskrunner.TaskContext) {}
...
+		applyStatus[obj] = status.ApplyReconcileStatus{
+			ApplyStatus: status.ApplySkipped,
+		}
+	}
+	for _, obj := range i.PrevInventory.Intersection(taskContext.FailedDeletes()) {
+		applyStatus[obj] = status.ApplyReconcileStatus{
+			ApplyStatus: status.PruneFailed,
+		}
+	}
+	for _, obj := range i.PrevInventory.Intersection(taskContext.SkippedDeletes()) {
```
mortent, line 155: Does this cover all the apply statuses?

pkg/inventory/inventory-client.go
```diff
@@ -33,7 +34,7 @@ type InventoryClient interface {
 	Merge(inv InventoryInfo, objs object.ObjMetadataSet, dryRun common.DryRunStrategy) (object.ObjMetadataSet, error)
 	// Replace replaces the set of objects stored in the inventory
 	// object with the passed set of objects, or an error if one occurs.
-	Replace(inv InventoryInfo, objs object.ObjMetadataSet, dryRun common.DryRunStrategy) error
+	Replace(inv InventoryInfo, objs object.ObjMetadataSet, dryRun common.DryRunStrategy, applyStatus map[object.ObjMetadata]status.ApplyReconcileStatus) error
```
mortent, line 37: Is the set of keys in the `applyStatus` map always the same as the set of `objs` here?

mortent (commented):
Two small comments, but I think this looks good. I'll let @seans3 and @karlkfi also take a look.

pkg/inventory/inventorycm.go
```diff
@@ -94,7 +95,7 @@ func (icm *InventoryConfigMap) Load() (object.ObjMetadataSet, error) {
 // Store is an Inventory interface function implemented to store
 // the object metadata in the wrapped ConfigMap. Actual storing
 // happens in "GetObject".
-func (icm *InventoryConfigMap) Store(objMetas object.ObjMetadataSet) error {
+func (icm *InventoryConfigMap) Store(objMetas object.ObjMetadataSet, _ map[object.ObjMetadata]status.ApplyReconcileStatus) error {
```
seans3, line 98: The fact that we don't even store the apply status in the default `ConfigMap` object illustrates that this functionality might not belong in the `cli-utils`. Are we sure we want to do this?

pkg/apply/task/inv_set_task.go
```diff
@@ -124,3 +128,36 @@ func (i *InvSetTask) Cancel(_ *taskrunner.TaskContext) {}
 
 // StatusUpdate is not supported by the InvSetTask.
 func (i *InvSetTask) StatusUpdate(_ *taskrunner.TaskContext, _ object.ObjMetadata) {}
+
+func (i InvSetTask) getApplyReoncileStatus(taskContext *taskrunner.TaskContext) map[object.ObjMetadata]status.ApplyReconcileStatus {
```
seans3, line 132: This function will probably not make sense to any future code reader. We should add a significant comment to this function describing which objects are being returned (e.g. objects we've computed in the cluster which belong in the inventory).

pkg/apply/task/inv_set_task.go
Liujingfang1: Yes, it covers all the apply statuses. The reconcile status are to be added later.

pkg/inventory/inventory-client.go
Liujingfang1: I think it should be the same as the `objs` set, but I haven't verified it.

pkg/inventory/inventorycm.go
Liujingfang1: We need to improve the interface for the inventory client and the inventory object later: getting the inventory object, updating the inventory object locally or in the cluster will be covered by the interface. When we do that, the `Store` function may not be needed. Changing this function here is to make minimal change on the existing code before we refactor the interfaces.

pkg/apply/task/inv_set_task.go
Liujingfang1: Added some comments for this function.

seans3 (approved):
/lgtm

karlkfi:
I thought this was gonna wait until I got the reconcile status in the TaskContext? I'm working on that PR now. I guess it will require an update to this...

Liujingfang1:
I think it was auto merged after the `/lgtm`.

PR #515, opened 2022-01-22
fix: Use the dynamic client in the InventoryClient

mortent:
This updates the InventoryClient implementation to use the dynamic client, rather than the builder and info functionality. It means we don't have to use the `resource.Info` types, which removes the need for the `InfoHelper` in the InventoryClient. The dynamic client is also easier to work with and comes with a `FakeDynamicClient` implementation that is easier to use than stubbing out the REST api through the TestFactory.

### DIFF
--- a/pkg/inventory/inventory-client.go
+++ b/pkg/inventory/inventory-client.go
@@ -4,17 +4,17 @@
 package inventory
 
 import (
+	"context"
 	"fmt"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/cli-runtime/pkg/resource"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/klog/v2"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/util"
-	"k8s.io/kubectl/pkg/validation"
-	"sigs.k8s.io/cli-utils/pkg/apply/info"
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/cli-utils/pkg/object"
 )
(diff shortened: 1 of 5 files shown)


pkg/inventory/inventory-client.go
```diff
@@ -411,19 +397,16 @@ func (cic *ClusterInventoryClient) applyInventoryObj(obj *unstructured.Unstructu
...
-	invInfo, err := cic.toInfo(obj)
-	if err != nil {
-		return err
-	}
-	helper := cic.helperFromInfo(invInfo)
-	klog.V(4).Infof("replacing inventory object: %s/%s", invInfo.Namespace, invInfo.Name)
-	var overwrite = true
-	replacedObj, err := helper.Replace(invInfo.Namespace, invInfo.Name, overwrite, invInfo.Object)
+
+	mapping, err := cic.mapper.RESTMapping(obj.GroupVersionKind().GroupKind())
```
ash2k, line 401: Here and in other places where `RESTMapping()` is being called I think we should get the mapping for the version of the object, not the default version.
```suggestion
	gvk := obj.GroupVersionKind()
	mapping, err := cic.mapper.RESTMapping(gvk.GroupKind(), gvk.Version())
```

ash2k (changes requested):

pkg/inventory/inventory-client.go
```diff
@@ -280,22 +274,19 @@ func (cic *ClusterInventoryClient) getClusterInventoryObjsByLabel(inv InventoryI
...
+		LabelSelector: labelSelector,
+	})
 	if err != nil {
 		return nil, err
 	}
-	return object.InfosToUnstructureds(retrievedInventoryInfos), nil
+	var invList []*unstructured.Unstructured
+	for i := range uList.Items {
+		u := uList.Items[i]
+		invList = append(invList, &u)
```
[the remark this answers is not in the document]
mortent: Fixed.

pkg/inventory/inventory-client.go
mortent: Agree. I have updated the PR.

seans3 (approved):
This looks good to me. I'll let others have a look. If there are no objections we'll merge this in two days.

ash2k (approved):
/lgtm

seans3 (approved):
/lgtm

PR #516, opened 2022-01-23
fix: Remove support for multiple inventory objects with the same inventory id

mortent:
This removes support for having multiple inventory objects with the same inventory id. Having multiple inventory objects with the same id was part of the original design, but we switched to using a single inventory object well over 12 months ago. We kept support for reading and merging multiple inventories, but I think it is time we consider this situation an error. It removes some complex code from the InventoryClient and also cleans up the `InventoryClient` interface by no longer updating the live state in functions that seems like they should be read only. This also allows us to remove the `dry-run` flag from those functions.


mortent:
@seans3 Can you take a look at this one. I want to make sure you agree that enough time has passed so we can clean this up.

seans3 (approved):
/lgtm
/approve

PR #517, opened 2022-01-23
fix: remove reflection hack

ash2k:
Reset RESTMapper using the [new API](https://github.com/kubernetes/kubernetes/pull/105623) rather than using reflection.

Fixes https://github.com/kubernetes-sigs/cli-utils/issues/421.


seans3 (approved):
/lgtm
/approve

PR #519, opened 2022-01-28
Expose ClusterReader options in the API

mortent:
This change allows clients to provide custom implementations of the `ClusterReader` interface for polling. The StatusPoller.Poll function currently allows users to choose between two different implementations using the `UseCache` option. With this change, the ClusterReader implementation is instead provided when the `StatusPoller` is created and allows for any implementation of the interface. This allows users to provide custom implementations for ApplyDestroy by providing a custom instance of the StatusPoller (this is currently a bit awkward, but should be addressed by https://github.com/kubernetes-sigs/cli-utils/pull/485)


seans3 (approved):
/lgtm

PR #520, opened 2022-01-30
Avoid goroutine being blocked forever in WaitTask

mortent:
This addresses an issue where cancelling a WaitTask will cause the goroutine to block forever. With this change, we use a context instead of a timer so that the blocking goroutine will be released when the context is cancelled.

This has already been fixed as part of a larger refactor on the master branch.


karlkfi (approved):
/lgtm

Nice job finding the memory leak!

PR #523, opened 2022-02-02
[WIP] fix: skip if dependencies not reconciled

karlkfi:
This is a hack to demonstrate a problem and the uglines of solving it to generate conversation about how to solve it better.

Changes:
- Split graph.SortObjs into AddDependencies + Sort + Filter + Reverse (for delete)
- Move AddDependencies into Applier/Destroyer to include both applyObjs & pruneObjs.
- Add DepenedencyFilter to filter apply/delete using the Graph & TaskContext
  Getting the graph into a Filter is messy...
  Do we need to check all statuses or just reconciled?
- Moved AdoptAll behavior into InventoryPolicyApplyFilter to simplify building filters in the Applier
- Renamed ResourceReference to ObjectReference and moed from mutation -> reference pkg
- Fix field NotFound error requiring the BadValue (always empty)
- Added ObjMetadataSet.Copy
- Added dependents annotation (currently unused)
- Refactored depends-on to use ObjectReference
- Change Graph.Sort to be non-destructive.
  This allows sorting the same graph for ApplyTasks and PruneTasks.
  Should we just cache the result instead?
- Added e2e test for DependencyFilter
- Added e2e test for LocalNamespaceFilter
- Clarified error in SkipInvalid e2e test

Open Questions:
- How do we build a DependencyFilter without needing to inject the Graph?
  Filters and Mutators only take one object at a time.
  I tried adding a "dependent" annotation to all objects refered to by a dependency, but couldn't find a clean way to inject it without also injecting the graph.
- If we pull the graph out of the solver, is it really a solver? Should it be replaced?
- Do we want depends-on to allow optional api version?
  If not, should we remove it from apply-time-mutation?
  They should ideally share code, but it can't easily be ObjMetadata, because ObjMetadata.String is a different format used by the inventory ConfigMap.

Major Fixed Bugs:
1. https://github.com/kubernetes-sigs/cli-utils/issues/526
2. https://github.com/kubernetes-sigs/cli-utils/issues/527
3. https://github.com/kubernetes-sigs/cli-utils/issues/528

TODO:
- https://github.com/kubernetes-sigs/cli-utils/pull/531
- https://github.com/kubernetes-sigs/cli-utils/pull/539
- https://github.com/kubernetes-sigs/cli-utils/pull/549
- Extract ObjectReference refactor
- https://github.com/kubernetes-sigs/cli-utils/pull/541
- Extract LocalNamespaceFilter e2e test
- Extract SkipInvalid e2e test fix
- Rebase DependencyFilter
- Punt on fixing dependency between applyobjs & pruneObjs


pkg/apply/applier.go
```diff
@@ -159,6 +160,21 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
 		}
 		klog.V(4).Infof("calculated %d apply objs; %d prune objs", len(applyObjs), len(pruneObjs))
 
+		// Add all objects to a dependency graph
```
karlkfi, line 163: This chunk makes me cringe. 

I wanted to just use `objects` but that doesn't include the objects in the inventory that need pruning. So I had to merge `applyObjs` and `pruneObjs` without modifying either set.

Then the AddDependencies functions can only be executed once, otherwise it can't resolve all the dependencies. In order to make AddDependencies executable multiple times it would need to be pulled out to its own class with a Build or Finalize function to validate if there are any external dependencies...

pkg/apply/applier.go
```diff
@@ -180,14 +197,17 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.InventoryInfo, obje
 			InventoryPolicy:        options.InventoryPolicy,
 		}
 		// Build list of apply validation filters.
-		applyFilters := []filter.ValidationFilter{}
-		if options.InventoryPolicy != inventory.AdoptAll {
-			applyFilters = append(applyFilters, filter.InventoryPolicyApplyFilter{
+		applyFilters := []filter.ValidationFilter{
```
karlkfi, line 200: This change is a net win, IMO. It simply moves the if statement into the InventoryPolicyApplyFilter.

pkg/apply/filter/dependency-filter.go
```diff
@@ -0,0 +1,94 @@
...
+// Filter returns true if the specified object should be skipped because at
+// least one of its dependencies is Not Found or Not Reconciled.
+func (dnrf DependencyFilter) Filter(obj *unstructured.Unstructured) (bool, string, error) {
+	id := object.UnstructuredToObjMetadata(obj)
+
+	switch dnrf.Action {
+	case ApplyAction:
+		edgesFrom := dnrf.TaskContext.Graph.EdgesFrom(id)
+		for _, depID := range edgesFrom {
+			if dnrf.TaskContext.IsInvalidObject(depID) {
```
karlkfi, line 45: I have no idea if we really need to check all these statuses... Maybe we only need to check `IsSuccessfulReconcile`? But the error message would then be misleading saying it didn't reconcile when it was never applied...

pkg/apply/filter/dependency-filter.go
```diff
@@ -0,0 +1,94 @@
...
+				return true, fmt.Sprintf("dependency invalid: %q", depID), nil
+			}
+			if dnrf.TaskContext.IsSkippedApply(depID) {
+				return true, fmt.Sprintf("dependency apply skipped: %q", depID), nil
+			}
+			if dnrf.TaskContext.IsFailedApply(depID) || !dnrf.TaskContext.IsSuccessfulApply(depID) {
```
karlkfi, line 51: It feels redundant to have both `IsFailedApply` and `IsSuccessfulApply` but they're not always one or the other... There's some weird code in the applier that might cause neither to be true.

karlkfi: Same is true for delete and reconcile status caches in TaskContext...
Maybe we just need to simplify into one status for each: apply, delete, reconcile?
Or maybe just apply & reconciles, with both meanings including apply and delete?

pkg/apply/prune/prune.go
```diff
@@ -99,6 +99,16 @@ func (p *Pruner) Prune(
 	for _, obj := range objs {
 		id := object.UnstructuredToObjMetadata(obj)
 		klog.V(5).Infof("evaluating prune filters (object: %q)", id)
+
+		// UID will change if the object is deleted and re-created.
+		uid := obj.GetUID()
+		if uid == "" {
```
karlkfi, line 105: I'm pretty sure we always GET the object before DELETE. So the UID should always be populated, right?

pkg/apply/prune/prune.go
```diff
@@ -146,6 +156,11 @@ func (p *Pruner) Prune(
 		if !opts.DryRunStrategy.ClientOrServerDryRun() {
 			klog.V(4).Infof("deleting object (object: %q)", id)
 			err := p.deleteObject(id, metav1.DeleteOptions{
+				// Only delete the resource if it hasn't already been deleted
+				// and recreated since the last GET. Otherwise error.
+				Preconditions: &metav1.Preconditions{
```
karlkfi, line 161: Without this precondition another client might delete and re-create the object before we have a chance to delete it. This change will cause that scenario to error with a failed apply. This should be ok because it will succeed on retry. It's kind of like how a client-side apply might fail from conflict if it was changed between GET and PATCH.

pkg/apply/solver/solver.go
```diff
@@ -209,9 +210,13 @@ func (t *TaskQueueBuilder) AppendPruneTask(pruneObjs object.UnstructuredSet,
 // (like CRD's). Returns a pointer to the Builder to chain function calls.
 func (t *TaskQueueBuilder) AppendApplyWaitTasks(applyObjs object.UnstructuredSet,
 	applyFilters []filter.ValidationFilter, applyMutators []mutator.Interface, o Options) *TaskQueueBuilder {
-	// Use the "depends-on" annotation to create a graph, ands sort the
-	// objects to apply into sets using a topological sort.
-	applySets, err := graph.SortObjs(applyObjs)
+	// Add dependencies from annotations
```
karlkfi, line 213: AddDependencies moved to the Applier/Destroyer. 
I don't like it there, but I'm not sure where a better place to put it is that has both `applyObjs` and `pruneObjs`.

pkg/apply/task/apply_task.go
```diff
@@ -178,6 +178,7 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
 				taskContext.AddFailedApply(id)
 			} else if info.Object != nil {
 				acc, err := meta.Accessor(info.Object)
+				// TODO: why is apply not successful if Accessor errors?
```
karlkfi, line 181: I don't know what the right way to handle `err != nil` here, but if it happens now `AddSuccessfulApply` won't get called and neither will `AddFailedApply`...

pkg/apply/taskrunner/context.go
```diff
@@ -242,3 +346,11 @@ type applyInfo struct {
 	// uid captures the uid of the resource that has been applied.
 	uid types.UID
 }
+
+// deleteInfo captures information about resources that have been
+// deleted. This is captured in the TaskContext so other tasks
+// running later might use this information.
+type deleteInfo struct {
+	// uid captures the uid of the resource that has been applied.
+	uid types.UID
```
karlkfi, line 355: Unfortunately `Generation` is not available from the DELETE response. If the object delete was delayed (deleteTimestamp) then the http call returns a non-KRM status object, not the object itself. And the dynamic client doesn't return anything but an error. So if you want the object when it does return you need to use the RESTClient directly...

This needs to be changed in upstream K8s, but it's not gonna happen any time soon.

pkg/object/dependents/annotation.go
```diff
@@ -0,0 +1,75 @@
+// Copyright 2021 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package dependents
```
karlkfi, line 4: This package/annotation is currently unused.

karlkfi:
/hold

being broken into multiple PRs

seans3 (commented):
Does this entire problem go away if we assume that all resources referenced by annotations (in sorting and apply-time mutation) are in the current apply set?

NOTE: This is what we had all initially agreed to, and it makes sense that our MVP for sorting would only support sorting/apply-time mutation on objects that are applied at the same time.

karlkfi:
> Does this entire problem go away if we assume that all resources referenced by annotations (in sorting and apply-time mutation) are in the current apply set?

No, we currently already assume this.

The problem is that the initial depends-on impl only ever relied on sorting to handle ordering. So the bug is that if A depends on B and A apply fails or times out, then B is still applied. So we need a filter to check if A was actually applied an reconciled before applying B.

This would be much simpler to implement is we hadn't flattened the dependency graph into phases. If we had a graph of tasks instead of a queue we could just prune a whole branch whenever an apply/delete/reconcile failed.

pkg/apply/applier.go
karlkfi: extracted to https://github.com/kubernetes-sigs/cli-utils/pull/541

karlkfi:
Replaced by https://github.com/kubernetes-sigs/cli-utils/pull/555

PR #530, opened 2022-02-02
capture the apply/reconcile status in the inventory object

Liujingfang1:
Capture the apply status for individual resources after each apply.
If the inventory object supports status field, it is updated
after setting the inventory list at then of one apply process.

BREAKING CHANGE: Update the inventory client and inventory interfaces
to pass the apply/reconcile status.

### DIFF
--- a/pkg/apply/task/inv_set_task.go
+++ b/pkg/apply/task/inv_set_task.go
@@ -114,8 +114,11 @@ func (i *InvSetTask) Start(taskContext *taskrunner.TaskContext) {
 		klog.V(4).Infof("keep in inventory %d invalid objects", len(invalidObjects))
 		invObjs = invObjs.Union(invalidObjects)
 
+		klog.V(4).Infof("get the apply status for %d objects", len(invObjs))
+		objStatus := taskContext.InventoryManager().Inventory().Status.Objects
+
 		klog.V(4).Infof("set inventory %d total objects", len(invObjs))
-		err := i.InvClient.Replace(i.InvInfo, invObjs, i.DryRun)
+		err := i.InvClient.Replace(i.InvInfo, invObjs, objStatus, i.DryRun)
 
 		klog.V(2).Infof("inventory set task completing (name: %q)", i.Name())
 		taskContext.TaskChannel() <- taskrunner.TaskResult{Err: err}
--- a/pkg/inventory/fake-inventory-client.go
+++ b/pkg/inventory/fake-inventory-client.go
@@ -6,14 +6,16 @@ package inventory
 import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"sigs.k8s.io/cli-utils/pkg/apis/actuation"
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/cli-utils/pkg/object"
 )
 
 // FakeClient is a testing implementation of the Client interface.
 type FakeClient struct {
-	Objs object.ObjMetadataSet
-	Err  error
+	Objs   object.ObjMetadataSet
+	Status []actuation.ObjectStatus
+	Err    error
 }
 
 var (
@@ -57,12 +59,13 @@ func (fic *FakeClient) Merge(_ Info, objs object.ObjMetadataSet, _ common.DryRun
 
 // Replace the stored cluster inventory objs with the passed obj, or an
 // error if one is set up.
-
-func (fic *FakeClient) Replace(_ Info, objs object.ObjMetadataSet, _ common.DryRunStrategy) error {
+func (fic *FakeClient) Replace(_ Info, objs object.ObjMetadataSet, status []actuation.ObjectStatus,
+	_ common.DryRunStrategy) error {
 	if fic.Err != nil {
 		return fic.Err
 	}
 	fic.Objs = objs
+	fic.Status = status
 	return nil
 }
 
--- a/pkg/inventory/inventory-client.go
+++ b/pkg/inventory/inventory-client.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/klog/v2"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/util"
+	"sigs.k8s.io/cli-utils/pkg/apis/actuation"
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/cli-utils/pkg/object"
 )
@@ -34,7 +35,7 @@ type Client interface {
 	Merge(inv Info, objs object.ObjMetadataSet, dryRun common.DryRunStrategy) (object.ObjMetadataSet, error)
 	// Replace replaces the set of objects stored in the inventory
 	// object with the passed set of objects, or an error if one occurs.
-	Replace(inv Info, objs object.ObjMetadataSet, dryRun common.DryRunStrategy) error
+	Replace(inv Info, objs object.ObjMetadataSet, status []actuation.ObjectStatus, dryRun common.DryRunStrategy) error
 	// DeleteInventoryObj deletes the passed inventory object from the APIServer.
 	DeleteInventoryObj(inv Info, dryRun common.DryRunStrategy) error
 	// ApplyInventoryNamespace applies the Namespace that the inventory object should be in.
@@ -100,8 +101,9 @@ func (cic *ClusterClient) Merge(localInv Info, objs object.ObjMetadataSet, dryRu
 	}
 	if clusterInv == nil {
 		// Wrap inventory object and store the inventory in it.
+		status := getObjStatus(nil, objs)
 		inv := cic.InventoryFactoryFunc(invObj)
-		if err := inv.Store(objs); err != nil {
+		if err := inv.Store(objs, status); err != nil {
 			return nil, err
 		}
 		invInfo, err := inv.GetObject()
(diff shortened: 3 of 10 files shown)


Liujingfang1:
@seans3 @mortent @karlkfi 
/hold

pkg/kstatus/status/apply_status.go
```diff
@@ -0,0 +1,38 @@
...
+	ReconcilePending   ReconcileStatus = iota // ReconcilePending
+	ReconcileSucceeded                        // ReconcileSucceeded
+	ReconcileSkipped                          // ReconcileSkipped
+	ReconcileTimeout                          // ReconcileTimeout
+	ReconcileFailed                           // ReconcileFailed
+)
+
+// ApplyReconcileStatus captures the apply and reconcile
+// status for a resource.
+type ApplyReconcileStatus struct {
```
karlkfi, line 35: Why add this to kstatus?
The status enums already exist in the inventory package.
https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/inventory/types.go

Also, I thought we decided to use the 3 field API (Strategy, Actuation, Reconcile), instead of just 2 fields.

pkg/inventory/inventory.go
```diff
@@ -31,7 +32,7 @@ type Inventory interface {
 	// Load retrieves the set of object metadata from the inventory object
 	Load() (object.ObjMetadataSet, error)
 	// Store the set of object metadata in the inventory object
-	Store(objs object.ObjMetadataSet) error
+	Store(objs object.ObjMetadataSet, applyStatus map[object.ObjMetadata]status.ApplyReconcileStatus) error
```
karlkfi, line 35: I'm trying to do a refactor to make the client/storage use the whole Inventory as input/output, but this map is probably fine for now. 

That said, do we want it to be an ObjectStatus list instead? 
https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/inventory/types.go#L60
Or is it important to keep it using ObjMetadata keys for now?

Liujingfang1: We can use the ObjectStatus list here. Converting it to a map is an implementation detail inside the `Store` function.

pkg/kstatus/status/apply_status.go
Liujingfang1: Updated this PR. Please take another look.

pkg/inventory/inventory.go
Liujingfang1: Updated PR. I used the `[]ObjStatus` here.

pkg/apply/task/inv_set_task.go
```diff
@@ -127,3 +130,12 @@ func (i *InvSetTask) Cancel(_ *taskrunner.TaskContext) {}
 
 // StatusUpdate is not supported by the InvSetTask.
 func (i *InvSetTask) StatusUpdate(_ *taskrunner.TaskContext, _ object.ObjMetadata) {}
+
+// getApplyReoncileStatus captures the apply status and reconcile status
+// for each individual resource in the inventory list.
+// The apply/reconcile status is then passed
+// to the inventory client and stored in the inventory object
+// through the Inventory interface.
+func (i InvSetTask) getApplyReoncileStatus(taskContext *taskrunner.TaskContext) []inventory.ObjectStatus {
```
karlkfi, line 139: I think you can just inline this one-line method.

pkg/inventory/inventory-client.go
```diff
@@ -133,14 +141,18 @@ func (cic *ClusterInventoryClient) Merge(localInv InventoryInfo, objs object.Obj
...
+		}
 	}
 
 	return pruneIds, nil
 }
 
 // Replace stores the passed objects in the cluster inventory object, or
 // an error if one occurred.
-func (cic *ClusterInventoryClient) Replace(localInv InventoryInfo, objs object.ObjMetadataSet, dryRun common.DryRunStrategy) error {
+func (cic *ClusterInventoryClient) Replace(localInv InventoryInfo, objs object.ObjMetadataSet, dryRun common.DryRunStrategy,
```
karlkfi, line 154: move applyStatus before dryRun

pkg/inventory/inventorycm.go
```diff
@@ -94,7 +94,7 @@ func (icm *InventoryConfigMap) Load() (object.ObjMetadataSet, error) {
 // Store is an Inventory interface function implemented to store
 // the object metadata in the wrapped ConfigMap. Actual storing
 // happens in "GetObject".
-func (icm *InventoryConfigMap) Store(objMetas object.ObjMetadataSet) error {
+func (icm *InventoryConfigMap) Store(objMetas object.ObjMetadataSet, _ []ObjectStatus) error {
```
karlkfi, line 97: We might want to serialize the status list into a string for the InventoryConfigMap. The current impl just has map keys, but we can uses the map values for status. Maybe serialize to a json string?

@seans3 what do you think?

pkg/apply/task/inv_set_task.go
Liujingfang1: Done

pkg/inventory/inventory-client.go

pkg/inventory/inventorycm.go
Liujingfang1: For the ConfigMap, I set the value of each key to the format of `{"actuation":XXX,"reconcile":XXX,"strategy":XXX}`. Let me know if this looks good to you.

Liujingfang1:
@karlkfi @mortent @seans3 This PR is ready for another review.

pkg/inventory/fake-inventory-client.go
```diff
@@ -57,8 +57,8 @@ func (fic *FakeInventoryClient) Merge(_ InventoryInfo, objs object.ObjMetadataSe
 
 // Replace the stored cluster inventory objs with the passed obj, or an
 // error if one is set up.
-
-func (fic *FakeInventoryClient) Replace(_ InventoryInfo, objs object.ObjMetadataSet, _ common.DryRunStrategy) error {
+func (fic *FakeInventoryClient) Replace(_ InventoryInfo, objs object.ObjMetadataSet, _ []ObjectStatus,
```
karlkfi, line 60: The fact that this passes tests worries me. It seems to indicate we're missing tests.
We don't ever read the status back out again, so I suspect the gap is just in verifying that the status was stored.

pkg/inventory/inventory-client.go
```diff
@@ -507,3 +511,28 @@ func (cic *ClusterInventoryClient) hasSubResource(groupVersion, resource, subres
 	}
 	return false, nil
 }
+
+// getObjStatus returns the list of object status
+// at the beginning of an apply process.
+func getObjStatus(pruneIds, unionIds []object.ObjMetadata) []ObjectStatus {
```
karlkfi, line 517: Why do we need to generate ObjectStatus in the InventoryClient at all? Shouldn't it always be provided by the caller?

pkg/inventory/inventory-client_test.go
```diff
@@ -291,11 +291,11 @@ func TestReplace(t *testing.T) {
 	// Client and server dry-run do not throw errors.
 	invClient, err := NewInventoryClient(tf, WrapInventoryObj, InvInfoToConfigMap)
 	require.NoError(t, err)
-	err = invClient.Replace(copyInventory(), object.ObjMetadataSet{}, common.DryRunClient)
+	err = invClient.Replace(copyInventory(), object.ObjMetadataSet{}, nil, common.DryRunClient)
```
karlkfi, line 294: These tests are just ignoring the new input. Please add new tests that exercise the new status field.

pkg/inventory/inventory_test.go
```diff
@@ -423,7 +423,7 @@ func copyInventory() InventoryInfo {
 
 func storeObjsInInventory(info InventoryInfo, objs object.ObjMetadataSet) *unstructured.Unstructured {
```
karlkfi, line 424: This function is only used in `inventory-client_test.go`. I think you can move it into that file. Then it probably needs to be modified to take the status too, and those tests modified to also store the status.

pkg/inventory/fake-inventory-client.go
Liujingfang1: Added unit test for the `Replace` call.

pkg/inventory/inventory-client_test.go
Liujingfang1: done

pkg/inventory/inventory_test.go
Liujingfang1: done

pkg/inventory/inventory-client.go
Liujingfang1: This is for the beginning part of the apply process. In the `Merge` function, it computes the union ids and ids to be pruned. So the `ObjectStatus` is not provided by the caller. We need to assign the ObjectStatus for different ids.

Liujingfang1:
@karlkfi I updated some unit tests to cover the status included in the inventory object.

pkg/inventory/inventory-client.go
```diff
@@ -493,3 +498,28 @@ func (cic *ClusterClient) hasSubResource(groupVersion, resource, subresource str
 	}
 	return false, nil
 }
+
+// getObjStatus returns the list of object status
+// at the beginning of an apply process.
+func getObjStatus(pruneIds, unionIds []object.ObjMetadata) []actuation.ObjectStatus {
```
karlkfi, line 504: I realized why this is being generated here. It's because we haven't updated InvAddTask to generate the initial status. I think we probably want to do that in a followup PR. The solution is complicated because we don't actually need a Merge method in the first place. The InvAddTask can be injected with the applyIdDs and pruneIDs which were generated from the merge of the local objects and Inventory objects. So all it needs to do is a Replace.

Liujingfang1: Yes, we can move the logic of computing the diff and object status into invAddTask. Then we can use the Replace function instead of the Merge function. 
We can do it in a separate PR.

pkg/inventory/inventorycm.go
```diff
@@ -115,10 +119,31 @@ func (icm *ConfigMap) GetObject() (*unstructured.Unstructured, error) {
...
+	objStatusMap := map[object.ObjMetadata]actuation.ObjectStatus{}
+	for _, status := range objStatus {
+		objStatusMap[ObjMetadataFromObjectReference(status.ObjectReference)] = status
+	}
 	for _, objMetadata := range objMetas {
-		objMap[objMetadata.String()] = ""
+		if status, found := objStatusMap[objMetadata]; found {
+			objMap[objMetadata.String()] = stringFrom(status)
+		} else {
+			objMap[objMetadata.String()] = ""
```
karlkfi, line 132: I think we might want to return an error here if there are objects in the spec that aren't in the status.

Liujingfang1: Done

karlkfi (approved):
/lgtm

This API will probably change again in the near future, but this is a fine intermediate change.

PR #531, opened 2022-02-02
chore: Add Inventory to TaskContext

karlkfi:
- Add a new Inventory KRM object for storing the spec and status
  of the inventory objects in memory.
- Improve reconcile, apply, & delete status tracking in the TaskContext/Inventory to cover all possible statuses
- Move most of the convenience methods from the TaskContext into a
  new inventory.Manager.
- Fix a minor bug where object UID might have drifted (delete & recreate) between GET and DELETE.

This is "Part 1" of a multi-PR change to simplify how the inventory is managed. 

The new Inventory scheme may or may not change moving forward, but it is only used in-memory for now, so it won't be a breaking change. For now, spec and metadata are unused. In the future, these should allow replacement of the InventoryInfo and InventoryClient to simplify the interface between cli-utils and clients that use custom inventory storage. 

The new Inventory doesn't technically need to be KRM, because it's primarily for in-memory usage, but having annotations should simplify serialization, logging, and test comparison.


pkg/inventory/types.go
```diff
@@ -0,0 +1,132 @@
...
+// InventoryStatus is the status of the current/last-known inventory state.
+type InventoryStatus struct {
+	Objects []ObjectStatus `json:"objects,omitempty"`
+}
+
+// ObjectReference is a reference to a KRM resource by name and kind.
+//
+// Kubernetes only stores one API Version for each Kind at any given time,
+// so version is not used when referencing objects.
+type ObjectReference struct {
```
seans3, line 35: This `ObjectReference` looks like a duplidate of `ObjMetadata`. Why do we need it?

seans3 (commented):
This `InventoryManager` is a clear improvement over the existing code. Small nit on the necessity of `ObjectReference`

pkg/inventory/types.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// ObjectStatus is a snapshot of the actuation and reconciliation status of a
+// referenced object.
+type ObjectStatus struct {
+	ObjectReference `json:",inline"`
+
+	Strategy  ActuationStrategy `json:"strategy,omitempty"`
+	Actuation ActuationStatus   `json:"actuation,omitempty"`
+	Reconcile ReconcileStatus   `json:"reconcile,omitempty"`
```
Liujingfang1, line 65: Please add some comments for these fields.

pkg/apply/prune/prune.go
```diff
@@ -99,6 +99,16 @@ func (p *Pruner) Prune(
 	for _, obj := range objs {
 		id := object.UnstructuredToObjMetadata(obj)
 		klog.V(5).Infof("evaluating prune filters (object: %q)", id)
+
+		// UID will change if the object is deleted and re-created.
+		uid := obj.GetUID()
+		if uid == "" {
+			err := object.NotFound([]interface{}{"metadata", "uid"}, "")
+			taskContext.SendEvent(eventFactory.CreateFailedEvent(id, err))
+			taskContext.InventoryManager().AddFailedDelete(id)
```
Liujingfang1, line 108: When the object is deleted or re-created by 3rd party, we think that the prune operation is successful.

karlkfi:
@seans3 for posterity I'll address your inline comment here:

You're right, ObjectReference is similar to ObjMetadata. We had a similar convo when I added mutation.ResourceReference.

First off, I regret naming ResourceReference with "resource" because it actually references an object (and instance of an API resource). So I want to replace ResourceReference with ObjectReference, but didn't want to refactor the mutation package in this PR (to keep size down).

Secondly, after discussion with @mortent, we came to the conclusion that the way ResourceReference & ObjMetadata use their String format for external facing interfaces is a problem. The String output should ideally only be used for logs and maybe errors. Tho in our case, errors are being passed through the printer to use user, so those all need to be consistently formatted, and they're not today. ObjMetadata.String is being used in the InventoryConfigMap serialization. And ResourceReference.String is being used in the ApplyTimeMutation serialization. They both should be modified to use specific encoders/decoders depending on situation. But I didn't want to try to make that change before first renaming ResourceReference to ObjectReference and moving it out of the mutation package.

Thirdly, ObjectReference uses flat top level fields with custom annotations which allow for json/yaml serialization, which I can't add to ObjMetadata because it uses the schema.GroupKind from apimachiney. That's also why K8s types don't use GroupKind in their type structs. For types structs used for APIs, it's best to have your own copy of the struct to annotate and decouple from other internal usages.

I'm hoping that in the future we can just replace ObjMetadata with ObjectReference, but I'm not committed to that yet, because it's a huge change. But it will probably be required if we want to make all the errors and printers consistently use the newer more readable depends-on style formatting (slash delimiters with optional namespace, instead of underscores with double-underscore escaping) without breaking reverse compatibility with existing ConfigMap inventories.

pkg/inventory/types.go
karlkfi: See PR comment: https://github.com/kubernetes-sigs/cli-utils/pull/531#issuecomment-1028511879

pkg/apply/prune/prune.go
karlkfi: If the object is deleted _before the applier runs_ then yes, you are correct.

However, if the object is deleted after the initial GET and before the DELETE, the current code would error with a 404 NotFound error.

And if the object were deleted and recreated after the initial GET and before the DELETE, the current code would delete the new object, thinking it was the old object, even if it didn't have the same inventory-id annotation.

This change I added causes the DELETE to error if the UID has changed at all since the GET was made, which fixes both of the above use cases by erroring with a clear error that the precondition was not satisfied because the UID has changed. This makes the behavior more like a server-side-apply, which errors if there's irreconcilable drift. The user can then simply retry the applier to re-do the GET & DELETE calls.

karlkfi: This particular line just errors if the UID is empty, which should never happen, because we always pull it from the server at the beginning of the applier when deciding what to prune.

seans3 (approved):
I'll merge this shortly if there are no other concerns.

pkg/apply/prune/prune.go
Liujingfang1: The cases that you described sounds good to me.

For the case that the object is deleted by the applier and recreated the 3rd party, it is after or during the applier run.
In this case, the UID doesn't change from GET to DELETE, but will change during the Wait period.
Could you point me to where it it handle?

pkg/inventory/types.go
karlkfi: Done.

pkg/apply/prune/prune.go
karlkfi: This PR doesn't solve that bug yet. I just needed the UID to put into the `Inventory.ObjectStatus.UID`, so I figured I'd fix a delete bug I saw while i was at it. I'll make another PR to fix the bug we caught with the Config Sync e2e test.

karlkfi: Now that the UID is in the TaskContext tho, it unblocks using it in the WaitTask.

Liujingfang1: Makes sense. Thank you for explaining this.

Liujingfang1:
The change looks good to me.

PR #532, opened 2022-02-03
add updateStatus function when applying the inventory object

Liujingfang1:
This change splits part of the change from
https://github.com/kubernetes-sigs/cli-utils/pull/530.

The change for capturing the apply/reconcile status will be after @karlkfi's change on the inventory interface.


Liujingfang1:
@mortent @karlkfi @seans3

karlkfi (commented):
Is there a way to look up whether the status is actually a subresource? Maybe from mapping?

We might want to exit early from updateStatus if there's no subresource (not just no "status" field).

Liujingfang1:
> Is there a way to look up whether the status is actually a subresource? Maybe from mapping?

I haven't figured out an easy way to look up if the object has a subresource. I tried the mapping from https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/api/meta/interfaces.go#L128-L131, but it doesn't contain the subresources.

Liujingfang1:
@karlkfi We can use the dynamic client to get the subresource for the object. If it returns an error, it means the substatus doesn't exist. I'll update the PR to use this method.

pkg/inventory/inventory-client.go
```diff
@@ -421,3 +431,42 @@ func (cic *ClusterInventoryClient) ApplyInventoryNamespace(obj *unstructured.Uns
...
+		return err
+	}
+	resource := cic.dc.Resource(mapping.Resource).Namespace(obj.GetNamespace())
+	meta := metav1.TypeMeta{
+		Kind:       obj.GetKind(),
+		APIVersion: obj.GetAPIVersion(),
+	}
+	liveObj, err := resource.Get(context.TODO(), obj.GetName(), metav1.GetOptions{TypeMeta: meta}, "status")
+	if err != nil {
+		klog.V(4).Infof("skip updating inventory subresource status for %v: %v", meta, err)
```
karlkfi, line 457: I think these errors include NotFound.
What error are we expecting exactly if it's not a subresource?

pkg/inventory/inventory-client.go
```diff
@@ -421,3 +431,42 @@ func (cic *ClusterInventoryClient) ApplyInventoryNamespace(obj *unstructured.Uns
...
+	err = unstructured.SetNestedMap(liveObj.UnstructuredContent(), status, "status")
+	if err != nil {
+		klog.V(4).Infof("skip updating inventory subresource status for %v: %v", meta, err)
+		return nil
+	}
+	_, err = resource.UpdateStatus(context.TODO(), liveObj, metav1.UpdateOptions{TypeMeta: meta})
+	if err != nil {
+		klog.V(4).Infof("failed to update inventory status for %v: %v", meta, err)
+	}
+	// Don't exit the apply process if failed to update the inventory object status.
```
karlkfi, line 470: why not?

pkg/inventory/inventory-client.go
Liujingfang1: This error is the `NOTFOUND` error since the subresource can't be found.

Since the updateStatus is updated after applying the inventory object, we shouldn't get the `NOTFOUND` error because of the inventory object itself doesn't exist.

pkg/inventory/inventory-client.go
Liujingfang1: This is to make sure that if applying the subresource fails due to the CRD schema changes(like a previous version of inventory object), it doesn't break the apply process.

pkg/inventory/inventory-client.go
```diff
@@ -421,3 +431,40 @@ func (cic *ClusterInventoryClient) ApplyInventoryNamespace(obj *unstructured.Uns
...
+		Kind:       obj.GetKind(),
+		APIVersion: obj.GetAPIVersion(),
+	}
+	liveObj, err := resource.Get(context.TODO(), obj.GetName(), metav1.GetOptions{TypeMeta: meta}, "status")
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(4).Infof("skip updating inventory subresource status for %v: %v", meta, err)
+			return nil
+		}
+		return err
```
karlkfi, line 461: fmt.Errorf("failed to get inventory status from cluster: %w", err)

pkg/inventory/inventory-client.go
```diff
@@ -421,3 +431,40 @@ func (cic *ClusterInventoryClient) ApplyInventoryNamespace(obj *unstructured.Uns
...
+	}
+
+	err = unstructured.SetNestedMap(liveObj.UnstructuredContent(), status, "status")
+	if err != nil {
+		return err
+	}
+	_, err = resource.UpdateStatus(context.TODO(), liveObj, metav1.UpdateOptions{TypeMeta: meta})
+	return err
```
karlkfi, line 469: fmt.Errorf("failed to write updated inventory status to cluster: %w", err)

pkg/inventory/inventory-client.go
```diff
@@ -421,3 +431,40 @@ func (cic *ClusterInventoryClient) ApplyInventoryNamespace(obj *unstructured.Uns
...
+		}
+		return fmt.Errorf("failed to get inventory status from cluster: %w", err)
+	}
+
+	err = unstructured.SetNestedMap(liveObj.UnstructuredContent(), status, "status")
+	if err != nil {
+		return err
+	}
+	_, err = resource.UpdateStatus(context.TODO(), liveObj, metav1.UpdateOptions{TypeMeta: meta})
+	return fmt.Errorf("failed to write updated inventory status to cluster: %w", err)
```
karlkfi, line 469: sorry my initial comment was wrong here. we also need a nil check, because it doesn't always error.

pkg/inventory/inventory-client.go
Liujingfang1: done

pkg/inventory/inventory-client.go
Liujingfang1: done

pkg/inventory/inventory-client.go
Liujingfang1: done

karlkfi (approved):
/lgtm

PR #533, opened 2022-02-06
Avoid reporting context cancellation or deadline expiration as resource error

mortent:
This updates the kstatus framework to better handle `context.Canceled` and `context.DeadlineExceeded`. Currently, if the context is canceled or the deadline expires while calling the `Get` or `List` functions on the controller-runtime client, a context error is returned, but the status readers or the polling functionality doesn't handle this properly. The error is returned for every resource and the polling doesn't immediately stop.
With this change, if the error returned from the ClusterReader is either `context.Canceled` or `context.DeadlineExceeded`, the status reader will return those as a separate error instead of including it in the `*event.ResourceStatus`. This is then handled appropriately by the poller to immediately stop work and return.
This should address the issue referenced in https://github.com/fluxcd/pkg/blob/main/ssa/manager_wait.go#L89

BREAKING CHANGE: The `ReadStatus` and `ReadStatusForObject` functions on the `StatusReader` interface updated to return an error in addition to `*event.ResourceStatus`


pkg/kstatus/polling/clusterreader/caching_reader.go
```diff
@@ -239,7 +239,12 @@ func (c *CachingClusterReader) Sync(ctx context.Context) error {
 		list.SetGroupVersionKind(mapping.GroupVersionKind)
 		err = c.reader.List(ctx, &list, listOptions...)
 		if err != nil {
-			// We continue even if there is an error. Whenever any pollers
+			// If the context was cancelled, we just stop the work and return
+			// the error.
+			if err == context.Canceled || err == context.DeadlineExceeded {
```
ash2k, line 244: Are we interested in supporting wrapped errors? i.e. `errors.Is(err, context.Canceled)` instead of equality. https://go.dev/blog/go1.13-errors

ash2k (commented):
Just one non-blocking question, otherwise looks great! I like the change.

pkg/kstatus/polling/clusterreader/caching_reader.go
mortent: Yeah, I agree that is better. I have updated the PR.

mortent:
/test cli-utils-presubmit-master

It seems like the TestApplier/apply_resource_with_existing_object_belonging_to_different_inventory test is flaking at times. I have created an issue for this and we can try to address that separately: https://github.com/kubernetes-sigs/cli-utils/issues/534

pkg/kstatus/polling/statusreaders/common.go
```diff
@@ -173,3 +160,52 @@ func toSelector(resource *unstructured.Unstructured, path ...string) (labels.Sel
...
+	}, nil
+}
+
+// errIdentifierToResourceStatus construct the appropriate ResourceStatus
+// object based on an error and the identifier for a resource.
+func errIdentifierToResourceStatus(err error, identifier object.ObjMetadata) (*event.ResourceStatus, error) {
+	// If the error is from the context, we don't attach that to the ResourceStatus,
+	// but just return it directly so the caller can decide how to handle this
+	// situation.
+	if err == context.Canceled || err == context.DeadlineExceeded {
```
ash2k, line 196:
```suggestion
	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```

ash2k (commented):
One more spot to change :)

pkg/kstatus/polling/statusreaders/common.go
mortent: Thanks for catching it. Fixed.

pkg/kstatus/polling/clusterreader/caching_reader.go
```diff
@@ -5,10 +5,11 @@ package clusterreader
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
```
karlkfi, line 12: we use `apierrors` as the alias elsewhere

pkg/kstatus/polling/engine/engine.go
```diff
@@ -204,17 +214,25 @@ func (r *statusPollerRunner) syncAndPoll() error {
...
 
 // pollStatusForAllResources iterates over all the resources in the set and delegates
 // to the appropriate engine to compute the status.
-func (r *statusPollerRunner) pollStatusForAllResources() {
+func (r *statusPollerRunner) pollStatusForAllResources() error {
 	for _, id := range r.identifiers {
+		// Check if the context has been cancelled on every iteration.
+		select {
+		case <-r.ctx.Done():
+			return r.ctx.Err()
```
karlkfi, line 227: Why not just return nil here? Then Run doesn't need to catch it and ignore it.

pkg/kstatus/polling/statusreaders/common.go
```diff
@@ -6,10 +6,11 @@ package statusreaders
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"sort"
 
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
```
karlkfi, line 13: apierrors

pkg/kstatus/polling/engine/engine.go
mortent: The idea was that since context errors from the `ReadStatus` are passed back, it would be consistent if we returned them here too. That way we can catch them all in `Run`.

pkg/kstatus/polling/clusterreader/caching_reader.go
mortent: Fixed.

karlkfi (approved):
/lgtm

PR #535, opened 2022-02-07
[WIP] Removes inventory object from applier Run signature

seans3:
* Removes the inventory object from the applier `Run` function signature.
* Supports `cli-utils` clients that do not want to use the pruning functionality.
* Moves the identification of the inventory object from outside `Run` to inside.
* In a future PR, the inventory add and set tasks will become dependent on if an inventory object is detected among the passed objects to apply.

BREAKING CHANGE: the `applier.Run()` function signature is changed to remove the inventory object. The `destroyer.Run()` function signature is changed to take all objects to delete instead of the inventory object.

### DIFF
--- a/cmd/preview/cmdpreview.go
+++ b/cmd/preview/cmdpreview.go
@@ -120,12 +120,6 @@ func (r *PreviewRunner) RunE(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	invObj, objs, err := inventory.SplitUnstructureds(objs)
-	if err != nil {
-		return err
-	}
-	inv := inventory.WrapInventoryInfoObj(invObj)
-
 	invClient, err := r.invFactory.NewInventoryClient(r.factory)
 	if err != nil {
 		return err
@@ -148,7 +142,7 @@ func (r *PreviewRunner) RunE(cmd *cobra.Command, args []string) error {
 
 		// Run the applier. It will return a channel where we can receive updates
 		// to keep track of progress and any issues.
-		ch = a.Run(ctx, inv, objs, apply.ApplierOptions{
+		ch = a.Run(ctx, objs, apply.ApplierOptions{
 			EmitStatusEvents:  false,
 			NoPrune:           noPrune,
 			DryRunStrategy:    drs,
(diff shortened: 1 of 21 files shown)


ash2k (commented):
Both idea and code LGTM. Thanks!

cmd/preview/cmdpreview.go
```diff
@@ -148,7 +142,7 @@ func (r *PreviewRunner) RunE(cmd *cobra.Command, args []string) error {
 
 		// Run the applier. It will return a channel where we can receive updates
 		// to keep track of progress and any issues.
-		ch = a.Run(ctx, inv, objs, apply.ApplierOptions{
+		ch = a.Run(ctx, objs, apply.ApplierOptions{
```
mortent, line 145: So how should we think about the information about inventory (being either the existing inventory information or a resource among the other resources with the special annotation)?

Are we providing the KRM resource that will be applied (Though in practice it will be a Replace rather than Apply/Patch) in the cluster, with the framework just adding additional data (i.e. the inventory) to the resource prior to updating the live state?

Or is the inventory resource resource just a place where we specify the key information needed to create, find, update, or delete the resource containing the inventory, but not providing the inventory resource itself?

I've always thought of it as the latter, where the information passed into the `Run` function (either as a KRM resource or as the InventoryInfo) just contains the key pieces of information needed (name, namespace and id), but it was up to the InventoryClient to manage the actual inventory resource in the cluster. I've been uncomfortable with the implementation, exactly because it uses the `InventoryInfo` interface to wrap a KRM resource.

pkg/apply/destroyer.go
```diff
@@ -42,20 +43,22 @@ func NewDestroyer(factory cmdutil.Factory, invClient inventory.InventoryClient)
...
 	return &Destroyer{
-		pruner:       pruner,
-		StatusPoller: statusPoller,
-		factory:      factory,
-		invClient:    invClient,
+		pruner:           pruner,
+		StatusPoller:     statusPoller,
+		factory:          factory,
+		invClient:        invClient,
+		WrapInventoryObj: inventory.WrapInventoryInfoObj,
```
mortent, line 50: This property is private in the Applier and public in the Destroyer. Is there a reason why it isn't private in both?

mortent (commented):
So I don't like the idea of passing the inventory object in with the other resources, since it is special. I find the API to be easier to understand when it is a separate parameter rather than a resource like the others that happens to have a special annotation. I know we have other examples of this, but I still don't think it makes for a good API.

If this change is to support non-inventory use-cases, I would like to see what the complete solution looks like before being too definitive. Is the idea to implement support for this in this PR?
I think there might be a solution to that use-case that involves a special noop implementation of the InventoryClient. I think the idea that has been discussed about simplifying the InventoryClient interface might help here.

PR #538, opened 2022-02-09
chore: Upgrade golint, replace deprecated linters, and fix new lint errors

karlkfi:
This small change turned into a big project to rename a lot of the exposed API. There are no behavioral changes, but a lot of Go API changes, namely to reduce stuttering and avoid returning unexposed types from exposed functions. I've split a lot of the changes into their own commit to help make review easier.


mortent:
This looks good to me. It changes a lot of names, but the majority of them are unlikely to be used by any users. And if anyone depends on them, this change will result in compile time errors, so they should be easy to find.

The commit message should probably mention the important names that are changed as a breaking change. The `EventType` to `Type` change on the events comes to mind.

I do want others to have a chance to take look too.

ash2k:
👍

karlkfi:
Rebased and passing tests!

seans3 (approved):
/lgtm

PR #540, opened 2022-02-11
Use discovery client to check if inventory objects has subresource status

Liujingfang1:
Use the function `DiscoveryClient.ServerResourcesForGroupVersion`
to list all the resources in a group. Then check if the subresource
status is contained in the list.

### DIFF
--- a/pkg/inventory/inventory-client.go
+++ b/pkg/inventory/inventory-client.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/klog/v2"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
@@ -48,11 +49,13 @@ type InventoryClient interface {
 // InventoryClient interface.
 type ClusterInventoryClient struct {
 	dc                    dynamic.Interface
+	discoveryClient       discovery.CachedDiscoveryInterface
 	mapper                meta.RESTMapper
 	InventoryFactoryFunc  StorageFactoryFunc
 	invToUnstructuredFunc InventoryToUnstructuredFunc
 }
 
+var _ InventoryClient = &ClusterInventoryClient{}
 var _ InventoryClient = &ClusterInventoryClient{}
 
 // NewInventoryClient returns a concrete implementation of the
@@ -68,8 +71,13 @@ func NewInventoryClient(factory cmdutil.Factory,
 	if err != nil {
 		return nil, err
 	}
+	discoveryClinet, err := factory.ToDiscoveryClient()
+	if err != nil {
+		return nil, err
+	}
 	clusterInventoryClient := ClusterInventoryClient{
 		dc:                    dc,
+		discoveryClient:       discoveryClinet,
 		mapper:                mapper,
 		InventoryFactoryFunc:  invFunc,
 		invToUnstructuredFunc: invToUnstructuredFunc,
@@ -102,8 +110,14 @@ func (cic *ClusterInventoryClient) Merge(localInv InventoryInfo, objs object.Obj
 			return nil, err
 		}
 		klog.V(4).Infof("creating initial inventory object with %d objects", len(objs))
-		if err := cic.createInventoryObj(invInfo, dryRun); err != nil {
+		var createdObj *unstructured.Unstructured
+		if createdObj, err = cic.createInventoryObj(invInfo, dryRun); err != nil {
 			return nil, err
+		}
+		// TODO(Liujingfang1): Remove the nil check statement after
+		// fixing the client mock in the unit test.
+		if createdObj != nil {
+			invInfo.SetResourceVersion(createdObj.GetResourceVersion())
 		}
 		if err := cic.updateStatus(invInfo, dryRun); err != nil {
 			return nil, err
@@ -135,8 +149,14 @@ func (cic *ClusterInventoryClient) Merge(localInv InventoryInfo, objs object.Obj
 				return pruneIds, err
 			}
 			klog.V(4).Infof("update cluster inventory: %s/%s", clusterInv.GetNamespace(), clusterInv.GetName())
-			if err := cic.applyInventoryObj(clusterInv, dryRun); err != nil {
+			var appliedObj *unstructured.Unstructured
+			if appliedObj, err = cic.applyInventoryObj(clusterInv, dryRun); err != nil {
 				return pruneIds, err
+			}
+			// TODO(Liujingfang1): Remove the nil check statement after
+			// fixing the client mock in the unit test.
+			if appliedObj != nil {
+				clusterInv.SetResourceVersion(appliedObj.GetResourceVersion())
 			}
 		}
 		if err := cic.updateStatus(clusterInv, dryRun); err != nil {
@@ -170,8 +190,14 @@ func (cic *ClusterInventoryClient) Replace(localInv InventoryInfo, objs object.O
 	if !objs.Equal(clusterObjs) {
 		klog.V(4).Infof("replace cluster inventory: %s/%s", clusterInv.GetNamespace(), clusterInv.GetName())
 		klog.V(4).Infof("replace cluster inventory %d objects", len(objs))
-		if err := cic.applyInventoryObj(clusterInv, dryRun); err != nil {
+		var appliedObj *unstructured.Unstructured
+		if appliedObj, err = cic.applyInventoryObj(clusterInv, dryRun); err != nil {
 			return fmt.Errorf("failed to write updated inventory to cluster: %w", err)
+		}
+		// TODO(Liujingfang1): Remove the nil check statement after
+		// fixing the client mock in the unit test.
+		if appliedObj != nil {
+			clusterInv.SetResourceVersion(appliedObj.GetResourceVersion())
 		}
 	}
 	if err := cic.updateStatus(clusterInv, dryRun); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/inventory/inventory-client.go
```diff
@@ -441,24 +448,36 @@ func (cic *ClusterInventoryClient) updateStatus(obj *unstructured.Unstructured,
...
+		return nil
+	}
+
+	resources, err := cic.discoveryClient.ServerResourcesForGroupVersion(obj.GetAPIVersion())
 	if err != nil {
 		return err
 	}
+	hasStatus := false
+	for _, r := range resources.APIResources {
+		if r.Name == mapping.Resource.Resource+"/status" {
```
karlkfi, line 462: Maybe pull it out with the previous ServerResourcesForGroupVersion call to a method called `hasSubResource(mapping Mapping, subresource string)` and use  `name := fmt.Sprintf("%s/%s", resource, subresource)` to prebuild the name to match.

pkg/inventory/inventory-client.go
```diff
@@ -441,24 +448,28 @@ func (cic *ClusterInventoryClient) updateStatus(obj *unstructured.Unstructured,
...
 	if err != nil {
-		if apierrors.IsNotFound(err) {
-			klog.V(4).Infof("skip updating inventory subresource status for %v: %v", meta, err)
-			return nil
-		}
-		return fmt.Errorf("failed to get inventory status from cluster: %w", err)
+		return fmt.Errorf("failed to get inventory object from cluster: %w", err)
 	}
 
 	err = unstructured.SetNestedMap(liveObj.UnstructuredContent(), status, "status")
```
karlkfi, line 475: UnstructuredContent is dangerous to use here. It returns an empty map if Object is nil. So if it then gets populated, the original liveObj will not be updated. 

I think in general you have to use UnstructuredContent for safe [Get]Nested calls but obj.Object for safe SetNested calls.

pkg/inventory/inventory-client.go
```diff
@@ -441,24 +448,28 @@ func (cic *ClusterInventoryClient) updateStatus(obj *unstructured.Unstructured,
...
+	}
+
+	klog.V(4).Infof("update inventory status")
 	resource := cic.dc.Resource(mapping.Resource).Namespace(obj.GetNamespace())
 	meta := metav1.TypeMeta{
 		Kind:       obj.GetKind(),
 		APIVersion: obj.GetAPIVersion(),
 	}
-	liveObj, err := resource.Get(context.TODO(), obj.GetName(), metav1.GetOptions{TypeMeta: meta}, "status")
+	liveObj, err := resource.Get(context.TODO(), obj.GetName(), metav1.GetOptions{TypeMeta: meta})
```
karlkfi, line 470: Can we skip this GET call? Or is it necessary in order to do an update with the latest spec & metadata? I think we have the original desired spec in `obj`, right?

karlkfi: Also, was the "status" subresource dropped here intentionally?

pkg/inventory/inventory-client.go
Liujingfang1: `liveObj` is get from the above `Get` call. When the `Get` call doesn't return an error, the `liveObj` shouldn't be empty

pkg/inventory/inventory-client.go
Liujingfang1: This is necessary since the update needs to be with latest spec and metadata. The `obj` may not have the latest metadata.

Liujingfang1: Yes, the "status" was dropped intentionally. It was originally included to check if the resource has the "status" subresource. Now we use the discovery client to check this. The "status" here can be dropped.

karlkfi: Do subresources have distinct metadata?
If you write metadata to a subresource does it change the parent resource's metadata?

pkg/inventory/inventory-client.go
karlkfi: If we expect it not to be empty, then you can just use liveObj.Object, right?

pkg/inventory/inventory-client.go
karlkfi: kernel team says CRD subresources drop/ignore metadata changes.
So we don't need to get new metadata. The only meta that matters are the ones used to identify the resource.

karlkfi: ConfigMaps don't have subresources, so it doesn't matter for that use case.

karlkfi: The Spec is also being dropped/ignored when you update the status, so we don't need to update that either.

I think the whole GET can be skipped.

pkg/inventory/inventory-client.go
Liujingfang1: updated.

Liujingfang1:
@karlkfi I added a field to capture the resourceVersion of the inventory object from the last Update/Create/Get. So we can skip the `Get` call when updating the status. This PR is ready for another review.

pkg/inventory/inventory-client.go
```diff
@@ -441,33 +464,47 @@ func (cic *ClusterInventoryClient) updateStatus(obj *unstructured.Unstructured,
...
-			klog.V(4).Infof("skip updating inventory subresource status for %v: %v", meta, err)
-			return nil
-		}
-		return fmt.Errorf("failed to get inventory status from cluster: %w", err)
-	}
-
-	err = unstructured.SetNestedMap(liveObj.UnstructuredContent(), status, "status")
-	if err != nil {
+	// set the resourceVersion of the inventory object
+	obj.SetResourceVersion(cic.invResourceVersion)
```
karlkfi, line 487: Is UID not also checked? Usually it's required to perform an update, at least on normal resources.

Kernel team says if subresource update doesn't check the UID it might be a bug.

pkg/inventory/inventory-client.go
```diff
@@ -377,9 +394,15 @@ func (cic *ClusterInventoryClient) createInventoryObj(obj *unstructured.Unstruct
...
-	_, err = cic.dc.Resource(mapping.Resource).Namespace(obj.GetNamespace()).
-		Create(context.TODO(), obj, metav1.CreateOptions{})
-	return err
+	var createdObj *unstructured.Unstructured
+	if createdObj, err = cic.dc.Resource(mapping.Resource).Namespace(obj.GetNamespace()).
+		Create(context.TODO(), obj, metav1.CreateOptions{}); err != nil {
+		return err
+	}
+	if createdObj != nil {
+		cic.invResourceVersion = createdObj.GetResourceVersion()
```
karlkfi, line 403: Why not just return the createdObj and copy the ResourceVersion in the caller? Caching it on the client struct seems unnecessary.

pkg/inventory/inventory-client.go
```diff
@@ -350,9 +361,15 @@ func (cic *ClusterInventoryClient) applyInventoryObj(obj *unstructured.Unstructu
...
-	_, err = cic.dc.Resource(mapping.Resource).Namespace(obj.GetNamespace()).
-		Update(context.TODO(), obj, metav1.UpdateOptions{})
-	return err
+	var updatedObj *unstructured.Unstructured
+	if updatedObj, err = cic.dc.Resource(mapping.Resource).Namespace(obj.GetNamespace()).
+		Update(context.TODO(), obj, metav1.UpdateOptions{}); err != nil {
+		return err
+	}
+	if updatedObj != nil {
+		cic.invResourceVersion = updatedObj.GetResourceVersion()
```
karlkfi, line 370: Why not just return the updatedObj and copy the ResourceVersion in the caller?

pkg/inventory/inventory-client.go
Liujingfang1: How could we check the UID if there is no extra `Get` to get the live object from the cluster? The UID in `liveObj` is from the last Get/Update/Create call.

pkg/inventory/inventory-client.go
Liujingfang1: We can also do it that way. Caching it looks simpler implementation. If we return the createdObj, we need to update multiple places in the code.

karlkfi: It's best not to cache things on the struct if you don't need to. It makes the code thread unsafe. The private methods are not exposed, so changing their signature is fine.

Liujingfang1: done

pkg/inventory/inventory-client.go
Liujingfang1: done

pkg/inventory/inventory-client.go
```diff
@@ -102,9 +110,13 @@ func (cic *ClusterInventoryClient) Merge(localInv InventoryInfo, objs object.Obj
 			return nil, err
 		}
 		klog.V(4).Infof("creating initial inventory object with %d objects", len(objs))
-		if err := cic.createInventoryObj(invInfo, dryRun); err != nil {
+		var createdObj *unstructured.Unstructured
+		if createdObj, err = cic.createInventoryObj(invInfo, dryRun); err != nil {
 			return nil, err
 		}
+		if createdObj != nil {
```
karlkfi, line 117: is this ever nil after success? I think we can assume it's non-nil, right?

karlkfi: same question for apply.

Liujingfang1: In the unit test, some functions that are mocked doesn't match the Get/Update behaviors, but I haven't figured out which part to change.

karlkfi: If the mock doesn't represent reality, we should fix the mock.

I think we can safely assume the object is non-nil if the error is nil, except in the DELETE implementation, where the object is not returned from the DynamicClient, because it is inconsistently returned from the RESTClient underneath.

Liujingfang1: How about adding a TODO for this and fix it afterwards? I can sync up with Morten to learn more clue for this, but I don't think we want to block the change for this.

karlkfi: fair enough.

Liujingfang1: Added the TODO.

pkg/inventory/inventory-client.go
karlkfi: I think we should be getting the UID from Storage.GetObject().

However, it looks like we don't actually need it, because it's not actually required to be the same on update, so it won't error as-is. If we wanted it to error we'd have to specify an UpdateOption.Precondition. But I don't know if it really matters. If something else is recreating inventory asynchronously, it's probably gonna break in other ways too.

pkg/inventory/inventory-client.go
karlkfi: Fixed by https://github.com/kubernetes-sigs/cli-utils/pull/545

PR #542, opened 2022-02-15
chore: Move Inventory object to pkg/apis/actuation

karlkfi:
- Make Inventory satisfy the metav1.Object, runtime.Object, and client.Object interfaces. 
  - This should make the Inventory API easier to use and easily convertible to Unstructured & YAML/JSON.
  - Add deepcopy-gen to generate DeepCopy methods for Inventory structs.
  - Replace custom TypeMeta & ObjectMeta with standard metav1 structs.
  - Use the group name `cli-utils.kubernetes.io` as a placeholder with no version.

BREAKING CHANGE: Move inventory.Inventory to actuation.Inventory (under pkg/apis/)


Makefile
```diff
@@ -35,7 +35,13 @@ install-addlicense:
 install-lint:
 	(which $(GOPATH)/bin/golangci-lint || go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.40.1)
 
-generate: install-stringer
+install-deepcopy-gen:
```
ash2k, line 38: I suggest we move `.PHONY` to each target individually otherwise it's easy to forget about it when creating targets. Doesn't have to be in this PR.

ash2k (approved):
/lgtm

Makefile
```diff
@@ -35,7 +35,13 @@ install-addlicense:
...
+	(which $(GOPATH)/bin/deepcopy-gen || go install k8s.io/code-generator/cmd/deepcopy-gen@v0.23.3)
+
+generate-deepcopy: install-deepcopy-gen
+	hack/run-in-gopath.sh deepcopy-gen --input-dirs ./pkg/apis/... -O zz_generated.deepcopy --go-header-file ./LICENSE_TEMPLATE_GO
```
Liujingfang1, line 42: `run-in-gopath.sh` is used only in `make generate-deepcopy`. Shall we also update other targets?

karlkfi: Nah. Only deepcopy-gen needs it. Everything else works in Go's module mode.
Upstream Kubernetes does the same thing for its generators.

Liujingfang1: Ack

Makefile
karlkfi: looks like we're missing quite a few already. So I'll make a separate PR for that.

karlkfi:
Looks like a flaky example test error timeout.

```
deployment.apps/mysql is NotFound: Resource not found
secret/mysql-pass is NotFound: Resource not found
service/mysql is NotFound: Resource not found
secret/mysql-pass created
service/mysql created
deployment.apps/mysql created
3 resource(s) applied. 3 created, 0 unchanged, 0 configured, 0 failed
secret/mysql-pass reconcile pending
service/mysql reconcile pending
deployment.apps/mysql reconcile pending
deployment.apps/mysql is InProgress: replicas: 0/1
secret/mysql-pass is Current: Resource is always ready
secret/mysql-pass reconciled
service/mysql is Current: Service is ready
service/mysql reconciled
deployment.apps/mysql is InProgress: Available: 0/1
deployment.apps/mysql is InProgress: Available: 0/1
deployment.apps/mysql is InProgress: Available: 0/1
deployment.apps/mysql reconcile timeout
2 resource(s) reconciled, 0 skipped, 0 failed to reconcile, 1 timed out
...
Example Test Failed: MultipleServices.md
```

PR #543, opened 2022-02-15
chore: Remove redundant object sorting

karlkfi:
- The Solver Graph performs sorting.
  So the applier & destroyer don't need to.


mortent (approved):
/approve

PR #544, opened 2022-02-15
chore: Log YAML to V7, instead of V4

karlkfi:
- Rename "resource" to "object" in mutation code.
  Resource is a REST interface. Object is an instance of a Resource.

### DIFF
--- a/pkg/apply/mutator/apply_time_mutator.go
+++ b/pkg/apply/mutator/apply_time_mutator.go
@@ -23,9 +23,9 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/object/mutation"
 )
 
-// ApplyTimeMutator mutates a resource by injecting values specified by the
+// ApplyTimeMutator mutates an object by injecting values specified by the
 // apply-time-mutation annotation.
-// The optional ResourceCache will be used to speed up source resource lookups,
+// The optional ResourceCache will be used to speed up source object lookups,
 // if specified.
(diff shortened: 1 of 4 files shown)


pkg/apply/mutator/apply_time_mutator.go
```diff
@@ -134,25 +136,26 @@ func (atm *ApplyTimeMutator) Mutate(ctx context.Context, obj *unstructured.Unstr
...
 
 		mutated = true
-		reason = fmt.Sprintf("resource contained annotation: %s", mutation.Annotation)
+		reason = fmt.Sprintf("object contained annotation: %s", mutation.Annotation)
 	}
 
 	if mutated {
-		klog.V(4).Infof("mutated target resource (%s):\n%s", targetRef, object.YamlStringer{O: obj})
+		klog.V(4).Infof("mutated target object: %s", targetRef)
+		klog.V(7).Infof("mutated target object YAML:\n%s", object.YamlStringer{O: obj})
```
Liujingfang1, line 151: I'm curious why logging the whole object yaml instead of the fields being mutated.

karlkfi: It's mostly just for debugging when the JSONPath fails to identify the field you wanted.

Liujingfang1: Ack

PR #545, opened 2022-02-15
fix: Always copy ResourceVersion before update

karlkfi:
- Fix FakeDynamicClient to return an object from deletes
- Fix FakeDynamicClient to return DeepCopy to avoid mutation

### DIFF
--- a/pkg/inventory/inventory-client.go
+++ b/pkg/inventory/inventory-client.go
@@ -110,15 +110,12 @@ func (cic *ClusterInventoryClient) Merge(localInv InventoryInfo, objs object.Obj
 			return nil, err
 		}
 		klog.V(4).Infof("creating initial inventory object with %d objects", len(objs))
-		var createdObj *unstructured.Unstructured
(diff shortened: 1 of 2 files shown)


pkg/inventory/inventory-client.go
```diff
@@ -144,20 +141,13 @@ func (cic *ClusterInventoryClient) Merge(localInv InventoryInfo, objs object.Obj
 			return pruneIds, nil
 		}
 		if !objs.Equal(clusterObjs) {
-			clusterInv, err = wrappedInv.GetObject()
```
karlkfi, line 147: Removed because redundant. It was called a few lines up.

karlkfi:
Example test helloapp.md failure from nil pointer. Seems like `createInventoryObj` returns nil for dry-run.

Liujingfang1:
> Example test helloapp.md failure from nil pointer. Seems like createInventoryObj returns nil for dry-run.
Good finding. We can check if it is a dry-run and exist early before triggering `createInventoryObj`.

PR #546, opened 2022-02-15
fix: Always copy ResourceVersion before update

karlkfi:
- Fix FakeDynamicClient to return an object from deletes
- Fix FakeDynamicClient to return DeepCopy to avoid mutation
- Fix InventoryClient to return a DeepCopy of the input when
  dry-run is enabled.

Replaces https://github.com/kubernetes-sigs/cli-utils/pull/545


Liujingfang1:
Thank you for fixing this

/lgtm
/approve

PR #549, opened 2022-02-17
chore: Change solver to decide which tasks to use

karlkfi:
- This redesign of the solver moves complexity about deciding which
  tasks to use from the applier/destroyer into the solver. This makes
  the solver more of a solver and less of a "do what your told"-er.
- One major difference is that the InvAddTask is now skipped if there
  are no objects to be applied. This makes logic sense because the
  inventory doesn't need to be expanded.
- This change unblocks unification of graph dependency resolution
  using both the apploy objects and prune objects, which will help
  distinguish between external dependencies and dependencies being
  accidentally deleted.

### DIFF
--- a/pkg/apply/applier.go
+++ b/pkg/apply/applier.go
@@ -127,25 +127,6 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.Info, objects objec
 
 		// Fetch the queue (channel) of tasks that should be executed.
 		klog.V(4).Infoln("applier building task queue...")
-		taskBuilder := &solver.TaskQueueBuilder{
-			Pruner:        a.pruner,
-			DynamicClient: a.client,
-			OpenAPIGetter: a.openAPIGetter,
-			InfoHelper:    a.infoHelper,
-			Mapper:        a.mapper,
-			InvClient:     a.invClient,
-			Destroy:       false,
-			Collector:     vCollector,
-		}
-		opts := solver.Options{
-			ServerSideOptions:      options.ServerSideOptions,
-			ReconcileTimeout:       options.ReconcileTimeout,
-			Prune:                  !options.NoPrune,
-			DryRunStrategy:         options.DryRunStrategy,
-			PrunePropagationPolicy: options.PrunePropagationPolicy,
-			PruneTimeout:           options.PruneTimeout,
-			InventoryPolicy:        options.InventoryPolicy,
-		}
 		// Build list of apply validation filters.
 		applyFilters := []filter.ValidationFilter{
 			filter.InventoryPolicyApplyFilter{
@@ -155,7 +136,6 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.Info, objects objec
 				InvPolicy: options.InventoryPolicy,
 			},
 		}
-
 		// Build list of prune validation filters.
 		pruneFilters := []filter.ValidationFilter{
 			filter.PreventRemoveFilter{},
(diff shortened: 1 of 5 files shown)


karlkfi:
These example tests seem to be erroring over a non-error...

```
deployment.apps/the-deployment deleted
service/the-service deleted
configmap/the-map2 deleted
3 resource(s) deleted, 0 skipped, 0 failed to delete
Success: output line found
Success: output line found
Success: output line found
Preview strategy: server
Error: output line not found
Expected: deployment.apps/the-deployment deleted]
```

pkg/apply/applier.go
```diff
@@ -176,14 +156,35 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.Info, objects objec
...
 		}
+		taskBuilder := &solver.TaskQueueBuilder{
+			Pruner:        a.pruner,
+			DynamicClient: a.client,
+			OpenAPIGetter: a.openAPIGetter,
+			InfoHelper:    a.infoHelper,
+			Mapper:        a.mapper,
+			InvClient:     a.invClient,
+			Collector:     vCollector,
+			ApplyFilters:  applyFilters,
```
mortent, line 167: Not necessarily something for this PR, but could we also move more the filter setup into the solver?

pkg/apply/applier_test.go
```diff
@@ -674,22 +674,6 @@ func TestApplier(t *testing.T) {
 					EventType: event.InitType,
 					InitEvent: &testutil.ExpInitEvent{},
 				},
-				{
```
mortent, line 677: More generally, do we have any use-cases where the client cares about the inventory-related tasks? Or could we consider them implementation details and avoid sending events for them?

mortent (approved):
Some comments, but they are more about the longer-term direction than this PR. I like the change.

/approve

pkg/apply/applier.go
karlkfi: sounds like a good follow up PR. It will need a bunch of new tests.

pkg/apply/applier_test.go
karlkfi: I think we want to send them because they make web requests, which might take up to 30s to timeout if the apiserver is slow. It's usually a good ideal to tell the caller what you're doing when you hand off control like that. It can also be useful when building a progress bar/% (tho we're not doing this yet).

karlkfi:
Found the cause of the test failure.

The DryRunStrategy was being passed in both the Options and the Builder structs. So it wasn't be propagated consistently. I've removed the one in the Builder and left it only in the Options.

pkg/apply/applier_test.go
karlkfi: TIL the printer don't print those events.... it probably should. 

Plus, it would have helped me find the bug in the tests that was deleting the inventory during a dry-run.

pkg/apply/solver/solver.go
```diff
@@ -86,74 +87,139 @@ func (tq *TaskQueue) ToActionGroups() []event.ActionGroup {
...
+
+	// Use the "depends-on" annotation to create a graph, ands sort the
+	// objects to apply into sets using a topological sort.
+	if len(applyObjs) > 0 {
+		applySets, err := graph.SortObjs(applyObjs)
+		if err != nil {
+			t.Collector.Collect(err)
+		}
+		for _, applySet := range applySets {
+			applySet = t.Collector.FilterInvalidObjects(applySet)
```
Liujingfang1, line 152: One thing that I don't quite understand is why filtering the applySet in addition to the filtering in line 130-131.

pkg/apply/solver/solver.go
```diff
@@ -86,74 +87,139 @@ func (tq *TaskQueue) ToActionGroups() []event.ActionGroup {
...
+
+	if o.Prune && len(pruneObjs) > 0 {
+		// Use the "depends-on" annotation to create a graph, ands sort the
+		// objects to prune into sets using a (reverse) topological sort.
+		pruneSets, err := graph.ReverseSortObjs(pruneObjs)
+		if err != nil {
+			t.Collector.Collect(err)
+		}
+		for _, pruneSet := range pruneSets {
+			pruneSet = t.Collector.FilterInvalidObjects(pruneSet)
```
Liujingfang1, line 175: ditto

pkg/apply/solver/solver.go
```diff
@@ -86,74 +87,139 @@ func (tq *TaskQueue) ToActionGroups() []event.ActionGroup {
...
+		tasks = append(tasks, &task.InvSetTask{
+			TaskName:      "inventory-set-0",
+			InvClient:     t.InvClient,
+			InvInfo:       t.invInfo,
+			PrevInventory: prevInvIds,
+			DryRun:        o.DryRunStrategy,
+		})
+	}
+
+	if o.Destroy {
```
Liujingfang1, line 203: You can change this to 
` else {`

pkg/apply/solver/solver.go
```diff
@@ -86,74 +87,139 @@ func (tq *TaskQueue) ToActionGroups() []event.ActionGroup {
...
+			InvClient: t.InvClient,
+			InvInfo:   t.invInfo,
+			Objects:   applyObjs,
+			DryRun:    o.DryRunStrategy,
+		})
+	}
+
+	// Use the "depends-on" annotation to create a graph, ands sort the
+	// objects to apply into sets using a topological sort.
+	if len(applyObjs) > 0 {
```
seans3, line 146: This **if** statement looks the same as the one directly above; can this statement be removed and the code consolidated into the above block?

seans3 (commented):
This looks good. There is a nit about combining the same **if** statement.

pkg/apply/solver/solver.go
karlkfi: you're right, it's redundant. I'll remove these ones.

pkg/apply/solver/solver.go
karlkfi: I could, but the above Destroy check is temporary. In the near future I want to always run InvSetTask so that the inventory gets updated if any objects were NOT deleted.

pkg/apply/solver/solver.go
karlkfi: i tried to keep the tasks logically separated for readability, but yes, i could "reduce" the code a bit.

pkg/apply/solver/solver.go
Liujingfang1: Ack

pkg/apply/solver/solver.go
karlkfi: turns out this is required, because graphing and sorting can produce additional invalid objects.

pkg/apply/solver/solver.go
karlkfi: done

pkg/apply/solver/solver.go
karlkfi: done

pkg/apply/solver/solver.go
karlkfi: I added comments to explain

Liujingfang1: Ack

Liujingfang1 (commented):
/lgtm

PR #550, opened 2022-02-18
Ensure that Daemonset controller has acted on the Daemonset.

### LINKED ISSUE #548: Bug: kstatus/status returns incorrect "Current" result immediately after DaemonSet apply
We have a use case where we want to create a DaemonSet and wait for its replicas to be ready before moving forward with the rest of the work. I was trying to use the `kstatus/status` library to do that. I thought I would apply the DaemonSet and then continuously check the result returned by `status.Compute()`.

However, immediately after applying the DaemonSet, the `status.Compute()` method returns a `Current` result seemingly because the status is: `status:map[currentNumberScheduled:0 desiredNumberScheduled:0 numberMisscheduled:0 numberReady:0]]}`

**Is this expected? Should we add an exception case to return an `InProgress` result when `desiredNumberScheduled == 0`?**

Here's the code I was using:
```go
package main

import (
	"context"
	"flag"
	"log"
	"os"

	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	serializeryaml "k8s.io/apimachinery/pkg/runtime/serializer/yaml"
	"k8s.io/apimachinery/pkg/types"
	"k8s.io/client-go/discovery"
	"k8s.io/client-go/discovery/cached/memory"
	"k8s.io/client-go/dynamic"
	"k8s.io/client-go/restmapper"
	"k8s.io/client-go/tools/clientcmd"
	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
)

func main() {
	kubeconfig := flag.String("kubeconfig", "", "absolute path to the kubeconfig file")
	manifest := flag.String("manifest", "", "absolute path to the daemonset manifest")
	flag.Parse()

	dsBytes, err := os.ReadFile(*manifest)
	if err != nil {
		log.Fatalf("Can't read %q: %v", *manifest, err)
	}

	obj := &unstructured.Unstructured{}
	_, gvk, err := serializeryaml.NewDecodingSerializer(unstructured.UnstructuredJSONScheme).Decode(dsBytes, nil, obj)
	if err != nil {
		log.Fatalf("Can't decode yaml: %v", err)
	}
	result, err := status.Compute(obj)
	if err != nil {
		log.Fatalf("Error computing status: %v", err)
	}
	log.Printf("Result Before: %v", result)

	config, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
	if err != nil {
		log.Fatalf("Can't create *rest.Config to talk to the cluster: %v", err)
	}

	dynamicClient, err := dynamic.NewForConfig(config)
	if err != nil {
		log.Fatalf("Can't create a dynamic client: %v", err)
	}

	discoveryClient, err := discovery.NewDiscoveryClientForConfig(config)
	if err != nil {
		log.Fatalf("Can't create a discovery client: %v", err)
	}
	mapper := restmapper.NewDeferredDiscoveryRESTMapper(memory.NewMemCacheClient(discoveryClient))

	mapping, err := mapper.RESTMapping(gvk.GroupKind(), gvk.Version)
	if err != nil {
		log.Fatalf("Can't find GVR for GVK: %v", err)
	}

	if newObj, err := dynamicClient.Resource(mapping.Resource).Namespace(obj.GetNamespace()).Patch(context.TODO(), obj.GetName(), types.ApplyPatchType, dsBytes, metav1.PatchOptions{FieldManager: "deployer"}); err != nil {
		log.Fatalf("Failed to apply patch: %v", err)
	} else {
		result, err := status.Compute(newObj)
		if err != nil {
			log.Fatalf("Error computing new status: %v", err)
		}
		log.Printf("Result After: %v", result)
	}
}
```

Here's the output of the run:
```shell
$ ./main --kubeconfig admin.conf --manifest=ds.yaml
2022/02/17 08:37:15 Result Before: &{InProgress Missing .status.desiredNumberScheduled [{Reconciling True NoDesiredNumber Missing .status.desiredNumberScheduled}]}
2022/02/17 08:37:15 Result After: &{Current All replicas scheduled as expected. Replicas: 0 []}
```

Because I had a "bad" DaemonSet that didn't start, if I run the program again, it worked correctly:
```shell
$ ./main --kubeconfig admin.conf --manifest=ds.yaml
2022/02/17 08:37:30 Result Before: &{InProgress Missing .status.desiredNumberScheduled [{Reconciling True NoDesiredNumber Missing .status.desiredNumberScheduled}]}
2022/02/17 08:37:30 Result After: &{InProgress Available: 0/2 [{Reconciling True LessAvailable Available: 0/2}]}
```

rohitagarwal003:
In status.Compute() -> checkGenericProperties() -> checkGeneration(),
we check that metadata.generation is equal to status.observedGeneration
but don't return an InProgress status if either of them is unset.

This may be reasonable for generic resource where we are unsure if these
fields would be set, but for daemonsets, we *know* that these fields get
set by the controller and so we should ensure that.

This fixes #548.

### DIFF
--- a/pkg/kstatus/status/core.go
+++ b/pkg/kstatus/status/core.go
@@ -303,6 +303,16 @@ func replicasetConditions(u *unstructured.Unstructured) (*Result, error) {
 
 // daemonsetConditions return standardized Conditions for DaemonSet
 func daemonsetConditions(u *unstructured.Unstructured) (*Result, error) {
+	// We check that the latest generation is equal to observed generation as
+	// part of checking generic properties but in that case, we are lenient and
+	// skip the check if those fields are unset. For daemonset, we know that if
+	// the daemonset controller has acted on a resource, these fields would not
+	// be unset. So, we ensure that here.
+	res, err := checkGenerationSet(u)
+	if err != nil || res != nil {
+		return res, err
+	}
+
 	obj := u.UnstructuredContent()
 
 	// replicas
@@ -343,6 +353,38 @@ func daemonsetConditions(u *unstructured.Unstructured) (*Result, error) {
 		Message:    fmt.Sprintf("All replicas scheduled as expected. Replicas: %d", desiredNumberScheduled),
 		Conditions: []Condition{},
 	}, nil
+}
+
+// checkGenerationSet checks that the metadata.generation and
+// status.observedGeneration fields are set.
+func checkGenerationSet(u *unstructured.Unstructured) (*Result, error) {
+	_, found, err := unstructured.NestedInt64(u.Object, "metadata", "generation")
+	if err != nil {
+		return nil, fmt.Errorf("looking up metadata.generation from resource: %w", err)
+	}
+	if !found {
+		message := fmt.Sprintf("%s metadata.generation not found", u.GetKind())
+		return &Result{
+			Status:     InProgressStatus,
+			Message:    message,
+			Conditions: []Condition{newReconcilingCondition("NoGeneration", message)},
+		}, nil
+	}
+
+	_, found, err = unstructured.NestedInt64(u.Object, "status", "observedGeneration")
+	if err != nil {
+		return nil, fmt.Errorf("looking up status.observedGeneration from resource: %w", err)
+	}
+	if !found {
+		message := fmt.Sprintf("%s status.observedGeneration not found", u.GetKind())
+		return &Result{
+			Status:     InProgressStatus,
+			Message:    message,
+			Conditions: []Condition{newReconcilingCondition("NoObservedGeneration", message)},
+		}, nil
+	}
+
+	return nil, nil
 }
 
 // pvcConditions return standardized Conditions for PVC
--- a/pkg/kstatus/status/status_compute_test.go
+++ b/pkg/kstatus/status/status_compute_test.go
@@ -503,6 +503,21 @@ status:
    numberReady: 4
    observedGeneration: 1
 `
+var dsDifferentGeneration = `
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+   name: test
+   namespace: qual
+   generation: 2
+status:
+   desiredNumberScheduled: 4
+   currentNumberScheduled: 4
+   updatedNumberScheduled: 4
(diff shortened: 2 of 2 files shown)


pkg/kstatus/status/core.go
```diff
@@ -345,6 +355,38 @@ func daemonsetConditions(u *unstructured.Unstructured) (*Result, error) {
 	}, nil
 }
 
+// checkDaemonsetGeneration checks that the metadata.generation and
+// status.observedGeneration fields are set.
+func checkDaemonsetGeneration(u *unstructured.Unstructured) (*Result, error) {
```
mortent, line 360: Could we find a more generic name here? The same issue is almost certainly true for many other resources as well, including deployments and statefulset.

mortent (commented):
Thanks for fixing this. Just a small comment on the name, but looks good.

karlkfi:
Is this problem actually specific to DaemonSet?
Don't most built-in controllers use `metadata.generation` and `status.observedGeneration`?

rohitagarwal003:
Ping. Let me know how you want to proceed. I can also update the PR to update all resources in this PR if that's preferred.

karlkfi:
I think we will want a ore generic solution, but I don't know the set of resources that have controllers than specify `status.observedGeneration`. It's probably not 100%. So we might need to start with a list of resources we know do use it. Is that something you can compile? It's ok if it's not complete. We can always add more later. but I suspect the list includes many of the core resources and probably common extended ones too.

mortent:
I think many of the commonly used built-in resource types do have `observedGeneration` (in particular `Deployment` and `StatefulSet`), but we can't assume that all types have it. We should add this check to the ones we know follow this pattern, but I'm ok adding that in follow-up PRs.
The longer-term solution here could be to use the openAPI to check if a resource has the `status.observedGeneration` property, but that would be a bigger change. I'm also not sure if we want to go down this path.

/approve

rohitagarwal003:
Thanks @karlkfi, @mortent.

I will work on a follow-up PR. Can you also LGTM this in the meantime. Thanks!

seans3 (approved):
/lgtm

Issue #528, opened 2022-02-02
Bug: Dependencies between applyobjs & pruneObjs not respected (considered external deps)


karlkfi:
If an object being pruned is depended on by a new or existing object being applied, validation fails with an ExternalDependencyError (if ExitError) or continues to delete the dependency (if SkipInvalid).

The desired behavior should be to prevent deletion of the dependency.

Issue #526, opened 2022-02-02
Bug: Dependent objects applied after dependency fails/timeout


karlkfi:
If a dependency fails to apply or times out waiting to reconcile, any objects that require it will still be applied. They should be skipped.

Issue #554, opened 2022-02-25
Bug: InventoryManager.SetObjectStatus doesn't actually update the object status


karlkfi:
Critical bug in `InventoryManager.SetObjectStatus` means that none of the calls to SetObjectStatus actually work. Surprisingly this still passes tests, but it's likely that the status output is wrong in many cases.

 https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/inventory/manager.go#L68-L76

The `objStatus` input variable is being shaddowed by the for loop.

Because of the scoping of the for loop, the bug only applies if the object status was already set. This is probably why it hasn't failed tests. But this will change is we add planned pending actuations before the aply/prune task.

PR #555, opened 2022-02-25
feat: Add dependency filter

karlkfi:
- Pass TaskContext into TaskBuilder.Build
- Combine dependency graph for apply and prune objects.
  This is required to catch dependencies that would have been deleted.
- Replace graph.SortObjs into DependencyGraph + Sort + HydrateSetList
- Replace graph.ReverseSortObjs with ReverseSetList to perform on the
  combined (apply + prune) set list.
- Add planned pending applies and prune to the InventoryManager
  before executing the task queue.
  This allows the DependencyFilter to validate against the planned
  actuation strategy of objects that haven't been applied/pruned yet.
- Add the dependency graph to the TaskContext, for the
  DependencyFilter to use.
  This can be removed in the future if the filters are managed by the
  solver.
- Make Graph.Sort non-destructive, so the graph can be re-used by the
  DependencyFilter.
- Add Graph.EdgesFrom and EdgesTo for the DependencyFilter to use.
  This requires storing the reverse edge list.
- Add an e2e test for the DependencyFilter
- Add an e2e test for the LocalNamespaceFilter

Fixes https://github.com/kubernetes-sigs/cli-utils/issues/526
Fixes https://github.com/kubernetes-sigs/cli-utils/issues/528

Replaces https://github.com/kubernetes-sigs/cli-utils/pull/523


seans3 (commented):
This is the wrong approach. The approach should be three separate efforts to address the three completely different issues. 

https://github.com/kubernetes-sigs/cli-utils/issues/528 is the result of overzealous client-side validation. As I mentioned, we should not be doing client-side validation. The solution is to rollback the overzealous validation.

https://github.com/kubernetes-sigs/cli-utils/issues/554 should be a simple fix with a better test to  prove the fix.

For https://github.com/kubernetes-sigs/cli-utils/issues/526, it is not clear this is a bug as much as an optional feature. If we are solving this, we are not going to do it a couple days before a release without tests.

/hold

mortent (commented):
One small comment, but I think this approach looks fine. As for the concerns raised by @seans3:

For #554 and #526: We need more tests and I think the PR also mentions that those will be added before this is merged. So I think we are all in agreement there. 
I don't think the timing related to a release matters. We have to trust that the combination of code reviews and tests are sufficient to catch most bugs, and when they don't (which will happen from time to time), we have to add better tests. It is unclear to me if waiting would make it safer as we don't have any clients (to my knowledge) who are running unreleased code. We can leverage the tests in https://github.com/GoogleContainerTools/kpt to do an additional check before we make a release. I have done that in the past, but it not something we have automated.

For #526 specifically, I think that is a bug since it seems strange to handle dependencies, but also ignore them in the face of failures. Skipping them seems to align with the general approach of the library, where we continue with as much as possible when something fails, but also make it safe to retry to "make things right" after a failure.

I think we are all in agreement that we want to avoid as much client-side validation as possible. But in this case, we have always restricted dependencies to only include resources within the apply set. This just changes how we report the error.

pkg/apply/filter/dependency-filter.go
```diff
@@ -0,0 +1,113 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/cli-utils/pkg/apis/actuation"
+	"sigs.k8s.io/cli-utils/pkg/apply/taskrunner"
+	"sigs.k8s.io/cli-utils/pkg/object"
+)
+
+// DependencyFilter implements ValidationFilter interface to determine if an
+// object can be applied or deleted based on the status of it's dependencies.
+type DependencyFilter struct {
+	TaskContext *taskrunner.TaskContext
```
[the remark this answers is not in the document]
karlkfi: I agree, but was trying to limit the amount of unrelated changes. That can be done in a separate PR.

karlkfi:
> For https://github.com/kubernetes-sigs/cli-utils/issues/526, it is not clear this is a bug as much as an optional feature. If we are solving this, we are not going to do it a couple days before a release without tests.

IMO we should not launch depends-on in Config Sync without this bug fixed, and it was probably a mistake to launch the feature in kpt with this bug in the first place. Prior to this PR, depends-on was only half implemented (sorted ordering, but no enforcement).

karlkfi:
> https://github.com/kubernetes-sigs/cli-utils/issues/528 is the result of overzealous client-side validation. As I mentioned, we should not be doing client-side validation. The solution is to rollback the overzealous validation.

For CRDs this is probably fine, because there is a service-side error that prevents deletion of CRDs with CRs in use.

For namespaces tho, this is catastrophic, because namespace deletion causes garbage collection of everything else in the namespace.

For all other explicit dependencies there is no server-side dependency enforcement. So the client-side enforcement cannot be rolled back without completely removing the dependency ordering feature.

karlkfi:
> https://github.com/kubernetes-sigs/cli-utils/issues/554 should be a simple fix with a better test to prove the fix.

This is fair. I can extract this to another PR.

karlkfi:
Pulled out fix for #554: https://github.com/kubernetes-sigs/cli-utils/pull/556

karlkfi:
> https://github.com/kubernetes-sigs/cli-utils/issues/528 is the result of overzealous client-side validation. As I mentioned, we should not be doing client-side validation. The solution is to rollback the overzealous validation.

@seans3 Removing the validation for external dependencies won't fix this bug. It would stop the confusing validation error, but it wouldn't fix the behavior. It wouldn't fail server-side, because there's no server-side impl. And without the DependencyFilter, neither the apply nor the prune would be skipped. So the apply would go through and then its dependency would be deleted. And we would lose the validation error for otherwise legitimately external dependencies.

Combining the graph for both applies and prunes allows for detection of implicit dependencies (CRDs & Namespaces) which would have otherwise been missed. And for explicit dependencies (depends-on  & apply-time-mutation), it detects them as "internal" dependencies instead of "external". Then both implicit and explicit deps can be checked by the DependencyFilter. And to keep the behavior otherwise the same, the graph only needs to be sorted once and then filtered down to include either the apply objects or prune objects.

I was planning to add more unit tests, but the e2e tests and applier/destroyer tests already show that the existing behavior is otherwise preserved, and I wanted to make sure the approach was otherwise acceptable before writing more tests.

pkg/apply/filter/dependency-filter.go
```diff
@@ -0,0 +1,113 @@
...
+// least one of its dependencies is Not Found or Not Reconciled.
+func (dnrf DependencyFilter) Filter(obj *unstructured.Unstructured) (bool, string, error) {
+	id := object.UnstructuredToObjMetadata(obj)
+
+	switch dnrf.Strategy {
+	case actuation.ActuationStrategyApply:
+		// For apply, check dependencies (outgoing)
+		relationship := "dependency"
+		for _, depID := range dnrf.TaskContext.Graph().EdgesFrom(id) {
+			filter, reason, err := dnrf.filterByRelationStatus(depID, relationship)
```
Liujingfang1, line 39: Suggest changing "filter" to "skip" for readability

pkg/apply/filter/dependency-filter.go
```diff
@@ -0,0 +1,113 @@
...
+			}
+			if filter {
+				return filter, reason, nil
+			}
+		}
+	case actuation.ActuationStrategyDelete:
+		// For delete, check dependents (incoming)
+		relationship := "dependent"
+		for _, depID := range dnrf.TaskContext.Graph().EdgesTo(id) {
+			filter, reason, err := dnrf.filterByRelationStatus(depID, relationship)
```
Liujingfang1, line 51: ditto

pkg/apply/solver/solver.go
```diff
@@ -119,19 +119,48 @@ func (t *TaskQueueBuilder) WithPruneObjects(pruneObjs object.UnstructuredSet) *T
...
+	if err != nil {
+		t.Collector.Collect(err)
+	}
+	// Store graph for use by DependencyFilter
+	taskContext.SetGraph(g)
+	// Sort objects into phases (apply order).
+	// Cycles will be treated as validation errors.
+	idSetList, err := g.Sort()
+	if err != nil {
+		t.Collector.Collect(err)
```
Liujingfang1, line 150: Does `Collect(err)` send an error event?

pkg/object/graph/graph.go
```diff
@@ -101,25 +108,53 @@ func (g *Graph) Size() int {
...
+	edgesTo, exists := g.reverseEdges[to]
+	if !exists {
+		return nil
+	}
+	c := make(object.ObjMetadataSet, len(edgesTo))
+	copy(c, edgesTo)
+	return c
+}
+
+// Sort returns the ordered set of vertices after a topological sort.
```
Liujingfang1, line 143: Since the graph now contains`edges` and `reverseEdges`, do we need to update the existing unit test for `Sort`?

pkg/apply/solver/solver.go
karlkfi: Nope. It just collects the errors. The applier and destroyer have the code to either send the collected errors as error events or return as an error.

pkg/object/graph/graph.go
karlkfi: Yup. It's on the todo list. I just wanted to get feedback on the approach before writing a bunch of tests that i might have to throw away.

pkg/apply/filter/dependency-filter.go
karlkfi: done

pkg/apply/solver/solver.go
Liujingfang1: Ack.

karlkfi:
/unhold

Added unit tests:
- DependencyFilter.Filter
- solver.Build (apply, prune, apply & prune, inventory status)
- graph.ReverseSetList
- graph.HydrateSetList
- graph.DependencyGraph
- Graph.Dependencies
- Graph.Dependents
- Graph.Sort (non-destructive)

Discussed approach with @seans3 offline. His primary complaint was that the dependency ordering was never intended to be a guarantee, just an expedient to produce fewer errors. However, while this may have been true when dependencies were first added, there are a number of use cases that have been discovered that are not solved by just ordering, which are better solved by an ordering guarantee. Most of these use cases come down to improving UX such that the error is obvious and once fixed in the local config, a re-apply is sufficient to resolve the  problem.

One example worth calling out is that this helps expedite deployment of chains of dependencies that would otherwise be applied all up front and fall into retry-back-off, causing end-to-end reconciliation to take orders of magnitude longer to complete compared to waiting to apply resources until their dependencies are reconciled. This is especially true for infrastructure resources which take a notoriously long time to reconcile (like GKE clusters). Plus, some retry loops (like Jobs) have a finite number of retries configured before failing, after which a reapply of the initial spec is insufficient.

seans3 (approved):
/approve

The tests look good. Thanks for all the quick work. I will leave the LGTM to Jingfang.

Liujingfang1:
Thanks for adding the unit tests.

/lgtm

PR #558, opened 2022-03-01
fix: Make DependencyFilter handle DryRun

karlkfi:
DryRun skips WaitEvents. So the DependencyFilter needs to skip
checking for reconciliation if DryRun is enabled.


seans3 (approved):
/lgtm

karlkfi:
/test cli-utils-presubmit-master 

Test pod never came up... Hoping it's a flake.

PR #560, opened 2022-03-02
fix: Allow empty set of apply objects

karlkfi:
- Add e2e test to ensure the inventory is still correctly created
  and destroyed when the apply object set is empty.

Fixes https://github.com/GoogleContainerTools/kpt/issues/2857

### DIFF
--- a/pkg/apply/applier_test.go
+++ b/pkg/apply/applier_test.go
@@ -641,6 +641,22 @@ func TestApplier(t *testing.T) {
 			},
 			expectedStatusEvents: []testutil.ExpEvent{
 				{
+					EventType: event.ActionGroupType,
+					ActionGroupEvent: &testutil.ExpActionGroupEvent{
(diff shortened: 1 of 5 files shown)


Liujingfang1:
> and destroyed when the apply object set is empty.

Is this destroy operation only for destroyer being called explicitly?
In other words, the inventory object is not destroyed even the object set is empty after after a replace?

karlkfi:
> > and destroyed when the apply object set is empty.
> 
> Is this destroy operation only for destroyer being called explicitly? In other words, the inventory object is not destroyed even the object set is empty after after a replace?

The Applier never deletes the inventory. Only the Destroyer deletes the inventory.

The new test (like most of the others) calls the applier, validates the result, and then calls the destroyer, and validates the result.

The only behavioral change here is that the InvAddtask now (again) executes on every Applier run.

PR #562, opened 2022-03-04
feat: expose skip reason as error

karlkfi:
This change makes filters return an error, since all errors from apply/prune/wait are sent as events and don't terminate the task queue anyway. This exposes the reason why the object was skipped, instead of just skipping it silently.

- Rewrite actuation filters to return an error with the reason for
  skipping.
- Add explicit error types for most skip errors, to make it easier to catch
  and handle them.
- Add Is method to explicit error types to allow use of errors.Is for
  recursive unwrapped matching.
- Rename InventoryPolicyFilter to InventoryPolicyPruneFilter for
  consistency with InventoryPolicyApplyFilter
- Update deletion prevention inventory-id removal to use errors.Is instead
  of matching the filter name.
- Convert error structs to use pointers to allow nil errors and avoid
  copying contents.

BREAKING CHANGE: Skipped actuation events now include an error.
BREAKING CHANGE: DeleteEvent.Reason replaced with an error.
BREAKING CHANGE: Unused InventoryNamespaceInSet error removed.
BREAKING CHANGE: InventoryOverlapError replaced with PolicyPreventedActuationError.
BREAKING CHANGE: NeedAdoptionError replaced with PolicyPreventedActuationError.
BREAKING CHANGE: NoInventoryObjError & MultipleInventoryObjError now use pointers.

New Error Types:
- filter.ApplyPreventedDeletionError (CurrentUIDFilter)
- filter.DependencyPreventedActuationError (DependencyFilter)
- filter.DependencyActuationMismatchError (DependencyFilter)
- inventory.PolicyPreventedActuationError (replaces NeedAdoptionError  & InventoryOverlapError from InventoryPolicyApplyFilter & InventoryPolicyPruneFilter)
- filter.NamespaceInUseError (LocalNamespacesFilter)
- filter.AnnotationPreventedDeletionError (PreventRemoveFilter)
- filter.FatalError (internal use only - for filters to trigger Failed instead of Skipped)

Removed error types:
- inventory.InventoryOverlapError (replaced by PolicyPreventedActuationError)
- inventory.NeedAdoptionError (replaced by PolicyPreventedActuationError)
- inventory.InventoryNamespaceInSet (unused - namespace is being created automatically)

Other existing error types (not used for skips):
- inventory.NoInventoryObjError (used by kpt, but not cli-utils)
- inventory.MultipleInventoryObjError (ValidateNoInventory)
- manifestreader.UnknownTypesError
- error.UnknownTypeError
- error.ApplyRunError
- error.InitializeApplyOptionError
- multierror.MultiError
- object.InvalidAnnotationError
- graph.ExternalDependencyError
- graph.CyclicDependencyError
- validation.Error
- common.ResultError


pkg/apply/applier_test.go
```diff
@@ -875,7 +875,12 @@ func TestApplier(t *testing.T) {
 					ApplyEvent: &testutil.ExpApplyEvent{
 						GroupName:  "apply-0",
 						Identifier: testutil.ToIdentifier(t, resources["deployment"]),
-						Error:      testutil.EqualErrorType(inventory.NewInventoryOverlapError(fmt.Errorf(""))),
+						Operation:  event.Unchanged,
```
karlkfi, line 878: I would like to change this to ApplySkipped in a future event operation refactor. But for now Unchanged == Skipped.

pkg/apply/event/event.go
```diff
@@ -286,15 +284,13 @@ type DeleteEvent struct {
 	Identifier object.ObjMetadata
 	Operation  DeleteEventOperation
 	Object     *unstructured.Unstructured
-	// If delete is skipped, this reason string explains why
-	Reason string
```
karlkfi, line 290: I don't know how DeleteEvent ended up with a `Reason` field while PruneEvent and ApplyEvent don't, but it's "fixed" now!

pkg/apply/filter/current-uids-filter_test.go
```diff
@@ -50,19 +48,8 @@ func TestCurrentUIDFilter(t *testing.T) {
...
-				t.Errorf("CurrentUIDFilter expected filter (%t), got (%t)", tc.filtered, actual)
-			}
-			if tc.filtered && len(reason) == 0 {
-				t.Errorf("CurrentUIDFilter filtered; expected but missing Reason")
-			}
-			if !tc.filtered && len(reason) > 0 {
-				t.Errorf("CurrentUIDFilter not filtered; received unexpected Reason: %s", reason)
-			}
+			err := filter.Filter(obj)
+			testutil.AssertEqual(t, tc.expectedError, err)
```
karlkfi, line 52: Now that we're using typed errors with the `Is` method implemented, we can use a simple AssertEqual and the error comparator will handle equality comparisons, even if the pointer doesn't point to the exact same error instance or even if it's wrapped in other errors.

pkg/apply/filter/dependency-filter_test.go
```diff
@@ -72,10 +67,10 @@ func TestDependencyFilter(t *testing.T) {
 				taskContext.InventoryManager().AddPendingApply(idA)
 				taskContext.AddInvalidObject(idInvalid)
 			},
-			id:               idA,
-			expectedFiltered: true,
-			expectedReason:   fmt.Sprintf("dependency invalid: %q", idInvalid),
-			expectedError:    nil,
+			id: idA,
+			expectedError: testutil.EqualError(
```
karlkfi, line 71: `assert.Equal` can't handle comparing `fmt.Errorf`, because it doesn't implement the `Is` or `Equal` methods. So we have to wrap it with `testutil.EqualError` to use the type and error string for comparison instead.

pkg/print/stats/stats.go
```diff
@@ -32,19 +32,22 @@ func (s *Stats) FailedReconciliationSum() int {
 func (s *Stats) Handle(e event.Event) {
 	switch e.Type {
 	case event.ApplyType:
-		if e.ApplyEvent.Error != nil {
+		if e.ApplyEvent.Error != nil &&
+			e.ApplyEvent.Operation != event.Unchanged {
```
karlkfi, line 36: This feels hacky, but the alternative is to add an ApplyFailed operation, which we can do in a future PR.

pkg/printers/json/formatter_test.go
```diff
@@ -106,6 +106,27 @@ func TestFormatter_FormatApplyEvent(t *testing.T) {
...
+		"resource apply skip error": {
+			previewStrategy: common.DryRunNone,
+			event: event.ApplyEvent{
+				Operation:  event.Unchanged,
+				Identifier: createIdentifier("apps", "Deployment", "", "my-dep"),
+				Error:      errors.New("example error"),
+			},
+			expected: []map[string]interface{}{
+				{
+					"eventType": "resourceApplied",
```
karlkfi, line 118: This "resourceApplied" eventType feels inaccurate, but "resourceFailed" is wrong too. We could add a new "resourceSkipped", but then it's redundant with the operations.

I think we should come back in another PR and clean up the json printer to use the newer strategy/actuation/reconcile pattern we made for the ResourceGroup status interface.

pkg/testutil/object.go
```diff
@@ -16,6 +16,10 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/object/dependson"
 )
 
+// OwningInventoryKey is the annotation key indicating the inventory owning an object.
+// This is a copy of inventory.OwningInventoryKey to avoid dependency cycle.
+const OwningInventoryKey = "config.k8s.io/owning-inventory"
```
karlkfi, line 21: It occurs to me that we could avoid this const copy if we injected it into the mutators that use it. Then we could change them to AddAnnotation and DeleteAnnotation. 

Is it worth doing in this PR or a follow up?

pkg/apply/event/event.go
```diff
@@ -261,15 +261,13 @@ type PruneEvent struct {
 	Identifier object.ObjMetadata
 	Operation  PruneEventOperation
 	Object     *unstructured.Unstructured
-	// If prune is skipped, this reason string explains why
-	Reason string
-	Error  error
+	Error      error
```
Liujingfang1, line 264: This change will break the cases when pruning objects are skipped, but the while case should pass. For example, the lifecycle annotation for preventing deletion. 

I think we need to keep the `Reason` field.

karlkfi: > This change will break the cases when pruning objects are skipped, but the while case should pass.

The skip will be returned as an error, but errors don't stop the task or the task queue. So it should just continue the same way it does now, right?

Callers do need to be modifier to allow for this, like I did in the printer code.

Liujingfang1: Yes, the callers need to to consume the prune events differently. Are the errors related to dependency all with specific error types?

karlkfi: > Are the errors related to dependency all with specific error types?

That's part of this PR. I added the following:
- CurrentUIDFilter:
  - filter.ApplyPreventedDeletionError
- DependencyFilter
  - filter.DependencyPreventedActuationError
  - filter.DependencyActuationMismatchError
- InventoryPolicyApplyFilter & InventoryPolicyPruneFilter:
  - inventory.PolicyPreventedActuationError
- LocalNamespacesFilter:
  - filter.NamespaceInUseError
- PreventRemoveFilter:
  - filter.AnnotationPreventedDeletionError

pkg/apply/event/event.go
karlkfi: Nvm, the PruneEvent has it too.

pkg/apply/filter/filter.go
```diff
@@ -14,8 +14,7 @@ import (
 type ValidationFilter interface {
 	// Name returns a filter name (usually for logging).
 	Name() string
-	// Filter returns true if validation fails. If true a
-	// reason string is included in the return. If an error happens
-	// during filtering it is returned.
-	Filter(obj *unstructured.Unstructured) (bool, string, error)
+	// Filter returns an error if validation fails, indicating that actuation
+	// should be skipped for this object.
+	Filter(obj *unstructured.Unstructured) error
```
mortent, line 19: So the type of the error returned from the filter will determine what should happen. We should document here that the `FatalError` is special.

pkg/print/stats/stats.go
mortent: Yeah, the assumption up until now has been that a non-nil error means the apply failed. When we are using the field for other uses, we should add an `ApplyFailed` operation.

pkg/printers/json/formatter_test.go
mortent: We need to be careful about breaking changes like this in the json printer. Unlike the other output formats, we know users are relying on the format here.

pkg/testutil/object.go
mortent: This is already a big PR, so do a follow-up if possible.

mortent (commented):
A few small comments, but this looks good to me.
/lgtm

pkg/apply/applier_test.go
karlkfi: Should I change it in this PR or a followup?

pkg/printers/json/formatter_test.go
karlkfi: This may or may not be breaking, because it would already have been `resourceApplied`, it just wouldn't have had an error. So it's only breaking if the caller is looking for the error first and not the "ApplyUnspecified" operation. Granted the likelihood is high, because there's no ApplyFailed... So maybe I should add ApplyFailed in another PR as a prerequisite.

pkg/apply/event/event.go
karlkfi: Yes, each of these has an error string explaining why, mostly for kpt tho. Config Sync likes to catch them and translate them into specific CS error codes.

pkg/apply/applier_test.go
karlkfi: Added ApplySkipped in https://github.com/kubernetes-sigs/cli-utils/pull/570

pkg/print/stats/stats.go
karlkfi: Added ApplyFailed in https://github.com/kubernetes-sigs/cli-utils/pull/570

PR #563, opened 2022-03-05
feat: Add inventory.StatusPolicy

karlkfi:
- StatusPolicyNone disables inventory status updates.
- StatusPolicyAll fully enables inventory status updates.
- This allows an opt-out feature for working around the problem
  that adding status can make the inventory larger than the max
  etcd object size, causing the applier to exit without applying
  or pruning anything. With StatusPolicyNone, the user can still
  safely prune objects to make their inventory smaller, and then
  re-enable the status with StatusPolicyAll.
- Note: the default ConfigMap does not currently support status,
  so this only affects custom inventory impls.


karlkfi:
/hold

Working on adding some tests. Apparently we never had any tests added for the inventory status, so none of them broke when I disabled it.

pkg/inventory/status-policy.go
```diff
@@ -0,0 +1,18 @@
+// Copyright 2019 The Kubernetes Authors.
```
Liujingfang1, line 1: 2019 -> 2022

Liujingfang1 (commented):
/lgtm

Liujingfang1:
The e2e testcases with `StatusPolicyAll` looks good. Is there any testcase for `StatusPolicyNone`?

karlkfi:
The ConfigMap provider uses StatusPolicyNone and the Custom provider uses StatusPolicyAll.

There's no tests for a custom provider with StatusPolicyAll and no status subresource, but we don't have nay clients doing that atm, so I didn't bother.

I also didn't add any unit tests. I would really like it if we had some for the ClusterClient that tested the various edge cases. But I'm at least relatively confident it works as-is now. Since you added the status update, it would be awesome if you could back fill some unit tests, @Liujingfang1. But they don't need to block this PR.

PR #564, opened 2022-03-06
fix: Avoid logging UID error for skipped apply

karlkfi:
Minor fix to avoid the `Empty object UID from InventoryManager` error level log message. Skipped apply objects won't have a UID, because the server sets the UID on first apply.


seans3 (approved):
/lgtm

PR #566, opened 2022-03-07
Use empty object status list when the StatusPolicyNone is used

Liujingfang1:
This reduces the request size when StatusPolicyNone is used

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -56,7 +56,7 @@ func main() {
 	initCmd := initcmd.NewCmdInit(f, ioStreams)
 	updateHelp(names, initCmd)
 	loader := manifestreader.NewManifestLoader(f)
-	invFactory := inventory.ClusterClientFactory{}
+	invFactory := inventory.ClusterClientFactory{StatusPolicy: inventory.StatusPolicyNone}
 	applyCmd := apply.Command(f, invFactory, loader, ioStreams)
 	updateHelp(names, applyCmd)
 	previewCmd := preview.Command(f, invFactory, loader, ioStreams)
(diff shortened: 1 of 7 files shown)


cmd/main.go
```diff
@@ -56,7 +56,7 @@ func main() {
 	initCmd := initcmd.NewCmdInit(f, ioStreams)
 	updateHelp(names, initCmd)
 	loader := manifestreader.NewManifestLoader(f)
-	invFactory := inventory.ClusterClientFactory{}
+	invFactory := inventory.ClusterClientFactory{StatusPolicy: inventory.StatusPolicyAll}
```
karlkfi, line 59: do we need this if the default ConfigMap inventory doesn't store the status anyway? That's why i had left it to default to None before.

karlkfi (changes requested):
The default inventory provider was deliberately StatusPolicyNone, because the custom inventory provider tests StatusPolicyAll. Why change it?

You might want to add some unit tests that test a custom provider so that we know the status update works when StatusPolicyAll and doesn't write them when StatusPolicyNone.

cmd/main.go
Liujingfang1: We injected the apply status into the ConfigMap inventory. Updated the InventoryCM, now we don't need to change this to All.

Liujingfang1:
> The default inventory provider was deliberately StatusPolicyNone, because the custom inventory provider tests StatusPolicyAll. Why change it?

We injected the apply status into the inventory configmap. If the passed in `[]ObjectStatus` is empty, it throws an error. Updated the code to allow this.

> You might want to add some unit tests that test a custom provider so that we know the status update works when StatusPolicyAll and doesn't write them when StatusPolicyNone.

Added an updated unit tests for this.

karlkfi (approved):
/lgtm

PR #568, opened 2022-03-07
[WIP] feat: immediate status check after apply

karlkfi:
This change adds an immediate status check after apply using the response object. This should short-circuit waiting for the StatusPoller for objects that reconcile immediately (e.g. ConfigMaps).

Unfortunately, this won't help CRDs or Namespaces, which actually need their controller to update their status.

This PoC impl is a hack. To do ti right, we'll need to refactor the StatusPoller to expose the injected StatusReaders, otherwise the ones used after apply won't match the ones injected by the user. Alternatively, we could just have users inject a list of StatusReaders and construct the StatusPoller ourselves.

### DIFF
--- a/pkg/apply/task/apply_task.go
+++ b/pkg/apply/task/apply_task.go
@@ -11,6 +11,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/resource"
@@ -19,6 +20,7 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/cmd/apply"
 	cmddelete "k8s.io/kubectl/pkg/cmd/delete"
+	"sigs.k8s.io/cli-utils/pkg/apply/cache"
 	applyerror "sigs.k8s.io/cli-utils/pkg/apply/error"
 	"sigs.k8s.io/cli-utils/pkg/apply/event"
 	"sigs.k8s.io/cli-utils/pkg/apply/filter"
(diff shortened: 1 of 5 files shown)


pkg/apply/task/apply_task.go
```diff
@@ -162,26 +169,74 @@ func (a *ApplyTask) Start(taskContext *taskrunner.TaskContext) {
...
+	if err != nil {
+		return fmt.Errorf("error creating ClusterReader: %w", err)
+	}
+
+	statusReaders, defaultStatusReader := polling.DefaultStatusReaders(a.Mapper)
+	statusReaders = append(statusReaders, defaultStatusReader)
+	for _, statusReader := range statusReaders {
+		if !statusReader.Supports(id.GroupKind) {
+			continue
+		}
```
Liujingfang1, line 223: I think you can pass in a map instead of  a list of engine.StatusReader. Then it avoids triggeirng `DefaultStatusReaders` and avoids the for loop in line 220.

karlkfi: The impl here is a hack. I'm gonna refactor this code into kstatus somehow. I think we need to make a MetaStatusReader impl that loops over a list of StatusReaders. Also, I'm not sure why the list of StatusReaders doesn't just include the default StatusReader at the end, but maybe it's because the author wanted to allow users to add new ones before the default one but after the others... I think I can get around that by just having the entire list be a single injected MetaStatusReader. Then I can just drop the list here and use the same MetaStatusReader as is injected into the StatusPoller.

Liujingfang1: Sounds good

karlkfi:
I think I'm going to put this optimization on hold for now. It has both positive and negative performance impacts.

Pros:
- For objects that are immediately ready (e.g. ConfigMap, CronJob, Secret), it would speed up detection of reconciliation. 
- Immediate reporting of StatusEvents might update the table printer more quickly, but usually just with "Pending", which they should already default to (I think pending is just left blank for now tho).

Cons:
- For objects that include generated objects (e.g. Deployment, ReplicaSet, StatefulSet), the status reader performs additional API calls (usually a LIST) to get the status of the generated objects.
- Immediate reconcile status lookup will almost always be not-reconciled for most objects. 

I'm not sure the pros outweigh the cons.

Liujingfang1:
I think we can apply this optimization only to the objects that are immediately ready.

karlkfi:
> I think we can apply this optimization only to the objects that are immediately ready. 

We would have to have a list of status readers specific to those "always ready" objects. Right now they're buried in the StatusPoller's defaultStatusReader. I can't think of a good way to expose them without duplicating the logic in two places. Plus, the StatusReader interface requires providing a ClusterReader, which we don't want it to use. So it'd be a big hack to try to share code, even if we wanted to.

Can you think of a better way to implement it that doesn't add a bunch of tech debt?

PR #570, opened 2022-03-24
feat: improve event status consistency

karlkfi:
## Event Changes

1. Rename Event "Operation" to "Status" to better reflect the meaning of the current enum values.
2. Ensure all Apply/Prune/Delete/Wait\* event status support:
    1. Pending
    2. Successful
    3. Skipped
    4. Failed
3. Combine Applied, Changed, Unchanged, and ServersideApplied into ApplySuccessful
    1. Use ApplySkipped instead of Unchanged when an object is skipped/filtered.
    2. Use ApplySuccessful instead of Unchanged when client-side-apply skips the PATCH because the GET detects no changes.
4. Rename ActionGroupEvent "Type" to "Status" for consistency\*\* with the other event status.

\* Timeout is only currently possible for Wait events, but we may add it to other events status in the future. For now, those events and API calls do not have a client-side timeout.

\*\* Action groups will continue to use "Started" & "Finished" as their status enum values, because action groups have no success or failure mode, so renaming them Pending and Successful would be misleading, but if we ever decide to add an error, then renaming them might make sense for consistency.

### All Printers

1. Print new short Status enum string wherever Operation was printed before
    - Pending
    - Successful
    - Skipped
    - Failed
    - Timeout
2. Track failure stats by checking the status field is -Failed, not just err != nil (Skip events may also error as of https://github.com/kubernetes-sigs/cli-utils/pull/562).

### Event Printer

1. Add a summary at the end of the Event printer to report stats.
    - This replaces the current mechanism that prints stats after the last ActionGroup of each type, because that method is hard to understand (apply & prune wait stats are combined).
2. Replace past tense short hand with more consistent two word status:
    - applied -> apply successful
    - pruned -> prune successful
    - deleted -> delete successful
3. Add output lines for action group started & finished events:
    - Format: "&lt;action> phase &lt;status>" (e.g. "apply phase started")
    - This replaces the stats that were printing on the last finished of each group type.
    - This should help explain why apply/prune/delete/wait aren't all done at the same time, and make it more obvious that the actuation is phased/grouped.
    - This includes inventory action group events, which previously were not printed, making it more transparent what is happening and easier to debug inventory update failures.

### JSON Printer

1. Rename the "operation" field to "status" to match the event change.
2. Remove the "eventType" field, because it duplicates the "type" + "status" fields.
3. Add a "summary" line to print the aggregated summary stats after all other events have been consumed.
    - One summary line for each action group type:
        - Apply
        - Prune
        - Delete
        - Wait
    - Format similar to existing stats fields:
        - action: "Apply|Prune|Delete|Wait"
        - count: %d
        - successful: %d
        - skipped: %d
        - failed: %d
        - timeout: %d (wait summary only) 
        - timestamp: %s (ISO 8601)
        - type: "summary"

### Table Printer

1. Print "Pending" instead of the empty string in the status column, once a pending event has been received.
    - Previously pending events were ignored
    - This should make it clearer in table view which objects are part of the current action group (aka phase) without having to add a new column to say so explicitly.

## BREAKING CHANGES
- Event "operations" and "type" are now "status"
- Event status enums renamed/refactored
- JSON printer schema changed to match events


pkg/apply/event/event.go
```diff
@@ -158,72 +158,71 @@ func (ee ErrorEvent) String() string {
...
+type ApplyEventStatus int
 
 const (
-	ApplyUnspecified ApplyEventOperation = iota
-	ServersideApplied
-	Created
-	Unchanged
-	Configured
+	ApplyPending    ApplyEventStatus = iota // Pending
+	ApplySuccessful                         // Successful
```
karlkfi, line 209: Do we need to preserve the distinction between Client-side Apply (Updated and Created) and Server-side Apply in the ApplyEvent?

If so, can it be done with a different field, like `ApplyMode: ServerSide|ClientSide`? 
Or do we also need to distinguish between Created, Updated, and Unchanged?

I think in the context of Config Sync, any time there's a retry the new Created/Updated will just be replaced by Unchanged, because the live state will already be as desired. So it doesn't really matter there. But maybe it matters for kpt, which doesn't retry automatically?

karlkfi:
I expect example tests to fail here, because I haven't updated them with the new event printer changes yet.

pkg/apply/event/event.go
mortent: I don't think we need to distinguish between SSA and client-side apply in the event. We use whatever approach the user provided as the argument (with one special case I think).

Client-side apply currently does tell us whether a resource was created, updated or unchanged, which is kinda useful. I know SSA doesn't distinguish between updated and unchanged, but does it tell us if a resources was created rather than updated/unchanged?

pkg/apply/taskrunner/task.go
```diff
@@ -131,13 +131,13 @@ func (w *WaitTask) Start(taskContext *TaskContext) {
 	}()
 }
 
-func (w *WaitTask) sendEvent(taskContext *TaskContext, id object.ObjMetadata, op event.WaitEventOperation) {
+func (w *WaitTask) sendEvent(taskContext *TaskContext, id object.ObjMetadata, op event.WaitEventStatus) {
```
mortent, line 134: nit: should the parameter be changed to `status` rather than `op`?

mortent (commented):
Some small comments, but looks good to me.

examples/alphaTestExamples/helloapp.md
```diff
@@ -174,11 +174,11 @@ Run preview to check which commands will be executed
 ```
 kapply preview $BASE | tee $OUTPUT/status
 
-expectedOutputLine "3 resource(s) applied. 3 created, 0 unchanged, 0 configured, 0 failed"
+expectedOutputLine "apply result: 3 attempted, 3 successful, 0 skipped, 0 failed"
 
 kapply preview $BASE --server-side | tee $OUTPUT/status
 
-expectedOutputLine "3 resource(s) applied. 0 created, 0 unchanged, 0 configured, 0 failed, 3 serverside applied"
+expectedOutputLine "apply result: 3 attempted, 3 successful, 0 skipped, 0 failed"
```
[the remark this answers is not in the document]
karlkfi: - The first number in the previous output was the total number for some actions and the successful number in others, leading to both inconsistency and inaccuracy. The new output makes explicit the difference between attempted and successful.
- Using the present tense can be done by simply lower-casing the String value of the action, leading to simpler event handling.

pkg/apply/taskrunner/task.go
karlkfi: good catch. i caught all the "operation" ones but not the "op" ones. will fix.

pkg/apply/event/event.go
karlkfi: > Client-side apply currently does tell us whether a resource was created, updated or unchanged, which is kinda useful.

Yeah, my feeling was that if we need this we can add another field to clarify it. That actually IS an "CRUD operation" and not just a "status". So we could have both operation & status just for apply events, if you think it's needed. 

I don't think anyone is actually using it from the JSON output today (of the clients we know about), and it wasn't exposed through Config Sync until 1.11.0 added it to the ResourceGroup object status. So I doubt anyone will complain. And we can always add it back later.

> I know SSA doesn't distinguish between updated and unchanged, but does it tell us if a resources was created rather than updated/unchanged?

Even if the API exposes it, the kubectl library we're using doesn't expose it. That's where these "operations" initially come from.

mortent (approved):
/approve

mortent (approved):
/lgtm

karlkfi:
Rebased. Tests passed. Needs another LGTM.

PR #572, opened 2022-04-09
feat: replace StatusPoller w/ StatusWatcher

karlkfi:
feat: replace StatusPoller w/ StatusWatcher

- Add DefaultStatusWatcher that wraps DynamicClient and manages
  informers for a set of resource objects.
  - Supports two modes: root-scoped & namespace-scoped.
  - Root-scoped mode uses root-scoped informers to efficiency and
    performance.
  - Namespace-scoped mode uses namespace-scoped informers to
    minimize the permissions needed to run and the size of the
    in-memory object cache.
  - Automatic mode selects which mode to use based on whether the
    objects being watched are in one or multiple namespaces.
    This is the default mode, optimizing for performance.
  - If CRDs are being watched, the creation/deletion of CRDs can
    cause informers for those custom resources to be created/deleted.
  - In namespace-scope mode, if namespaces are being watched, the
    creation/deletion of namespaces can also trigger informers to
    be created/deleted.
  - All creates/updates/deletes to CRDs also cause RESTMapper reset.
  - Allow pods to be unschedulable for 15s before reporting the
    status as Failed. Any update resets the timer.
- Add BlindStatusWatcher for testing and disabling for dry-run.
- Add DynamicClusterReader that wraps DynamicClient.
  This is now used to look up generated resources
  (ex: Deployment > ReplicaSets > Pods).
- Add DefaultStatusReader which uses a DelegatingStatusReader to
  wrap a list of conventional and specific StatusReaders.
  This should make it easier to extend the list of StatusReaders.
- Move some pending WaitEvents to be optional in tests, now that
  StatusWatcher can resolve their status before the WaitTask starts.
- Add a new Thousand Deployments stress test (10x kind nodes)
- Add some new logs for easier debugging
- Add internal SyncEvent so that apply/delete tasks don't start
  until the StatusWatcher has finished initial synchronization.
  This helps avoid missing events from actions that happen while
  synchronization is incomplete.
- Filter optional pending WaitEvents when testing.

BREAKING CHANGE: Replace StatusPoller w/ StatusWatcher
BREAKING CHANGE: Remove PollInterval (obsolete with watcher)


karlkfi:
I still have some tests to write and refactoring to do, but I got all the tests passing locally.

karlkfi:
Extracted https://github.com/kubernetes-sigs/cli-utils/pull/574

karlkfi:
Extracted https://github.com/kubernetes-sigs/cli-utils/pull/575

karlkfi:
Extracted https://github.com/kubernetes-sigs/cli-utils/pull/578

karlkfi:
Extracted https://github.com/kubernetes-sigs/cli-utils/pull/576

karlkfi:
Extracted: https://github.com/kubernetes-sigs/cli-utils/pull/579

karlkfi:
This is finally ready for review.

I've added unit tests for the new watcher, and all the existing e2e/stress tests pass. 
Notably, the stress tests are ~30% faster (17m -> 12m)!

pkg/kstatus/watcher/watcher_dynamic.go
```diff
@@ -0,0 +1,825 @@
...
+// containsNotFoundMessage checks if the error string contains the message for
+// StatusReasonNotFound.
+// See k8s.io/apimachinery@v0.23.2/pkg/api/errors/errors.go
+// This is necessary because the Informer doesn't properly wrap errors.
+func containsNotFoundMessage(err error) bool {
+	return strings.Contains(err.Error(), resourceNotFoundMessage)
+}
+
+// uniqueGKs returns a set of unique GroupKinds from a set of object identifiers.
+func uniqueGKs(ids object.ObjMetadataSet) []schema.GroupKind {
```
Liujingfang1, line 588: If this function returns `[]schema.GroupVersionKind`, then in each watcher, you don't need to look up the GVK from the restmapper. You can also get the `pendingWatchers` from it.

pkg/kstatus/watcher/watcher_dynamic.go
```diff
@@ -0,0 +1,825 @@
...
+		// TODO: Allow non-fatal errors to be reported using a specific error type.
+		return
+	}
+	klog.V(3).Infof("CRD updated for %s", groupKind)
+
+	klog.V(3).Info("Resetting RESTMapper")
+	// Reset mapper to invalidate cache.
+	meta.MaybeResetRESTMapper(w.Mapper)
+
+	if w.pendingWatchers.Pop(groupKind) {
```
Liujingfang1, line 461: What happens if there is already a watcher for this groupKind? Do we need to restart it?

ash2k (commented):
I'm sorry, I haven't fully reviewed it - ran out of time.

pkg/kstatus/watcher/watcher_dynamic.go
karlkfi: The uniqueness is by GroupKind tho, we wouldn't want to allow multiple Versions of the same GK to be watched simultaneously.

pkg/kstatus/watcher/watcher_dynamic.go
karlkfi: If there's already an informer then the GK doesn't get added to `pendingWatchers`. It only gets added to `pendingWatchers` if the CRD doesn't exist yet or if the previous informer has been deleted. No informer instance is restarted, because they're not designed to support that. A new informer instance is created, if needed (ex: CRD deleted and then re-created).

pkg/kstatus/polling/clusterreader/dynamic_reader.go
```diff
@@ -0,0 +1,81 @@
...
+
+// DynamicClusterReader is an implementation of the ClusterReader that delegates
+// all calls directly to the underlying DynamicClient. No caching.
+type DynamicClusterReader struct {
+	DynamicClient dynamic.Interface
+	Mapper        meta.RESTMapper
+}
+
+func (n *DynamicClusterReader) Get(ctx context.Context, key client.ObjectKey, obj *unstructured.Unstructured) error {
+	mapping, err := n.Mapper.RESTMapping(obj.GroupVersionKind().GroupKind())
```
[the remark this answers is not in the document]
karlkfi: FWIW, this DynamicClusterReader is only being used for GETing generated-resources, like ReplicaSets & Pods when a Deployment is being applied. So it's not being used for the primary objects the user cares about, and it's only being used for checking the objects status (tho it does get returned on the event, in case the caller wants to do something with it). Because it's not a primary object tho, we don't actually have the version, unless we assume generated resources with the same group also share the version (which doesn't help for Pods, which aren't in the apps group like ReplicaSets and Deployment).

But yes, if used more generically, I could see wanting to return the same version, if specified in the Unstructured obj.

pkg/kstatus/polling/clusterreader/dynamic_reader.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	serverObj, err := n.DynamicClient.Resource(mapping.Resource).
+		Namespace(key.Namespace).
+		Get(ctx, key.Name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	serverObj.DeepCopyInto(obj)
```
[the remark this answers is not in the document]
karlkfi: The DynamicClusterReader impls the ClusterReader interface, which is already used by the StatusPoller. I wanted to try to re-use the ClusterReader and StatusReader interfaces to reduce the number of required interface changes. 

But I agree that the interfaces here could be further optimized for the StatusWatcher use case.

Even having a DynamicClusterReader at all means we're losing caching on these objects and need to GET them every time we want to read the status of the parent object. This is suboptimal, but simpler that trying to write another layer of informers for generated resources... That could be a follow-up tho, if more optimization is important.

Just having the watcher reduces the stress-test time by 30%, but it's not testing any Deployments or other generate-resources. We might need a different stress test to measure it.

pkg/kstatus/watcher/watcher_dynamic.go
Liujingfang1: If an informer exist for a GK and the CRD for this type is updated, does the informer get affected?

karlkfi: Not currently. Presumably the existing informer will continue receiving updates, using whichever GVK it was already receiving. If the CRD was updated without a version bump, the changes should be reflected automatically. If the CRD version was incremented, the old default version would continue to be received.

I haven't tested for this, but this is what I assume would happen based on what i know of how the informers work.

Do you think this behavior will be problematic?

karlkfi: I could add code to replace the informer if a new GVK is added by CRD update. I'm not sure how to prevent dropping events or producing duplicate events in that case tho.

karlkfi:
heh. that's the flaky test i just fixed with https://github.com/kubernetes-sigs/cli-utils/pull/581

karlkfi:
There were some concerns about memory usage with informers, so I ran some stress tests:

# Time Spent (20 QPS)
Polling: 16m50s
Watching: 12m1s
Diff: - 4m49s (- 28.6%)

# Time Spent (-1 QPS)
Polling: 7m0s
Watching: 6m28s
Diff: - 32s (- 7.7%)

These shorter times seem to fluctuate up and down due to other factors (like kind cluster deployment), so the speed diff with throttling disabled on an otherwise empty local cluster might not be significant.

# Memory Allocations (-1 QPS)
Polling: 4745.35MB
Watching: 1761.93MB
Diff: - 2983.42 (- 62.8%)

Liujingfang1:
@karlkfi  thank you for testing the time and memory allocation difference. 

The change looks good to me overall. Would like @mortent also take a look.

/lgtm

karlkfi:
Summarizing line-comments for posterity:
- I added code to handle delete & re-add of CRDs, but that's not a scenario we're likely to see during an apply/destroy, unless it's caused by a 3rd party, because the current task queue always deletes after applying. The existing e2e tests do prove that adding and deleting CRDs with CRs in the same sync works tho.
- CRD version upgrades are probably going to be ignored by the watcher. It should continue returning the GVK it was originally asked for, which is the default version at the beginning of the watch, or whenever the CRD is first created. We're not doing much with the response object other than checking status and standard metadata fields, so this shouldn't be a big deal, but it could mean that new fields used for checking status won't be available until the next sync after a CRD is updated. We can improve this behavior later.
- The DynamicClusterReader may not be generally useful outside of the DynamicStatusWatcher, because it doesn't support specifying the version to GET/LIST. The fact that it looks up mapping every time is also more expensive than might be necessary. The ClusterReader API itself is not really ideal, but already used by the Poller, so it was used for the Watcher for consistency. There are lots of potential enhancements we could do here, but they may require changes to (or removal of) the Poller as well.

pkg/kstatus/watcher/watcher_dynamic.go
karlkfi: > When the CRD version is changed from "v1alpha" to "v1beta1" and "v1alpha1" is not supported. If the informer is for "v1alpha1" version, I'm not sure if it receives events when the version is changed to "v1beta1".

IME, when a new version is added, the old version usually sticks around for a while, rather than being immediately deleted. This allows CRs to be migrated from old to new. Otherwise, you'd have to delete all the old CRs before replacing "v1alpha1" with "v1beta1". When this happens, the old version should continue to work, it just won't necessarily be the default version. AFAIK, all installed versions need to be convertible to and from the default versions. So even if a controller upgrades to use a new "v1beta1" version, clients using the old "v1alpha1" version should still works. And that's what happening here. The Informer is a client using a specific version and that should keep working until that version is deleted from the CRD.

However, I agree that we should probably handle schema updates. And I agree that can be added in a future PR.

I suspect that the ideal behavior for a general use watcher impl would be to have the user specify which version they want to watch, but this won't be possible with the current API using ObjMetadata. The best we can do when only specifying group & kind as input is to always return the object using the cluster default version. I think it should be possible to do this with the existing informers, but it's going to require some client-side event de-duplication in order to avoid dropping events when migrating between informers for old and new versions. I suspect this probably requires a new informer wrapper that takes GK and manages the GVK itself.

pkg/kstatus/watcher/watcher_dynamic.go
```diff
@@ -0,0 +1,825 @@
...
+		klog.Warningf("Invalid CRD added: missing group and/or kind: %v", id)
+		// Don't return an error, because this should not inturrupt the task queue.
+		// TODO: Allow non-fatal errors to be reported using a specific error type.
+		return
+	}
+	klog.V(3).Infof("CRD added for %s", groupKind)
+
+	klog.V(3).Info("Resetting RESTMapper")
+	// Reset mapper to invalidate cache.
+	meta.MaybeResetRESTMapper(w.Mapper)
```
mortent, line 429: Do we need to do this here? We could wait until the CRD becomes established, but I'm not sure if we are guaranteed to get that update callback if the CRD becomes established immediately.

pkg/kstatus/watcher/watcher_dynamic.go
```diff
@@ -0,0 +1,825 @@
...
+	w.muxer.Input(eventCh)
+
+	// Build & Start an informer for each unique GroupKind.
+	for _, groupKind := range w.GroupKinds {
+		err := w.startForGroupKind(
+			ctx,
+			groupKind,
+		)
+		if err != nil {
+			handleError(eventCh, err)
```
mortent, line 159: Do we keep running here even if we fail to start the informer for a GroupKind? Unless it has changed recently, the expectation from clients should be that an ErrorEvent is used for fatal errors and will be followed by the event channel being closed.

pkg/kstatus/watcher/watcher_dynamic.go
```diff
@@ -0,0 +1,825 @@
...
+		Status:     status.NotFoundStatus,
+		Message:    "Resource not found",
+	}
+
+	return &event.ResourceStatus{
+		Identifier: id,
+		Resource:   nil, // deleted object has no
+		Status:     result.Status,
+		Message:    result.Message,
+		// Presumably if the object is deleted, so are all the resources it generated.
```
mortent, line 304: This is only true if resources are deleted using foreground cascading deletion: https://kubernetes.io/docs/concepts/architecture/garbage-collection/#foreground-deletion

pkg/kstatus/polling/clusterreader/dynamic_reader.go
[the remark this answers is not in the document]
karlkfi: The new DynamicStatusWatcher and DynamicClusterReader both use the `k8s.io/client-go/dynamic.Interface` client. But the ClusterReader interface still takes the `sigs.k8s.io/controller-runtime/pkg/client.ObjectKey` from controller-runtime. 

I don't think it's worth rewriting the poller to use client-go. And I can't change the ClusterReader interface without changing the StatusReader interface. So I opted to keep using both for this first pass.

In the future tho, I think we can just delete the poller and convert the StatusReader/ClusterReader fully over to use client-go. That would also unblock using watchers for generated resources, which would be faster.

pkg/kstatus/watcher/watcher_dynamic.go
karlkfi: AFAIK, the informers de-dupe missed events automatically. So if you miss an add and an update, it just gives you an add. But even if we check the status and only reset if it's reconciled, there's still some time between when it became reconciled and when we get the event that it was reconciled that the CRD is registered but our mapper cache says it's not. 

Since reset itself is cheap (it's only expensive when you make the next mapping request), I opted to reset in both add and update. If we get no other events between the add and the update, and no other async code does a lookup, the mapper will still be reset when we get the update event. But if there IS an async mapping call between, it will fetch the latest mappings, rather than remain potentially out of date.

Also, FWIW, the kernel team is working on some enhancements upstream to make resetting cheaper, so that it can hit an endpoint to determine which cached CRDs to invalidate.

pkg/kstatus/watcher/watcher_dynamic.go
karlkfi: Good catch. I'll make it return. It looks like that's what the StatusPoller does for every error.

pkg/kstatus/watcher/watcher_dynamic.go
karlkfi: True, but the Poller also has the same behavior in the `baseStatusReader`. If lookupResource returns NotFound, it builds a ResourceStatus with nil GeneratedResources.

I changed the comment to be more clear, but I don't think the behavior is a problem for our use case, because if deletion policy is NOT configured to be foreground, the desired behavior for confirming deletion implies ignoring the status of generated resources. So I left a TODO to improve it for more general use, if that becomes a priority.

karlkfi:
/hold

looks like the watcher tests are flaky. I was able to finally reproduce a similar error locally after a few dozen attempts. It sometimes exists without any events and no error saying why.

karlkfi:
Ah, important similarity: when there are no events it also timed out after 10s+

ash2k:
I still couldn't find time to review this, but based on description I have these concerns:
- Currently cli-utils only accesses objects that the user works with. With the informer model it will require reading/watching same GKs but across all namespaces. Users of a consumer program may not have permissions for that. It's typical to have permissions for a single namespace so cli-utils will not work in this situation, right? i.e. the new version requires more permissions. What can we do here? Maybe detect if all objects are within a namespace? Or maybe cli-utils can expose a flag to let the app/user configure it to use per-namespace informers? This also may be desired for huge clusters where it's better to watch several individual namespaces rather than the whole lot.
- My app uses multiple (0 or more, depending on user-supplied config) appliers. Can I share the informers among them? Sorry, still haven't looked at the code. Presumably `DynamiciStatusWatcher` is exactly for that purpose (one instance can be supplied to multiple appliers/etc), but I wanted to confirm.

p.s. thank you for taking on such massive tasks like this one!

karlkfi:
@ash2k: Good catch. Permissions are gonna be a problem. 

I had specifically avoided making an Informer for each namespace because it's much more expensive & slow, but I might have to to support clients with less permissions. It might still be more efficient than the Poller, but not by as much.

Unfortunately, all the tests we have in this repo use cluster-admin, so doing cluster-wise ListAndWatch worked fine. I'm gonna need to add some more granular setup to test having limited permissions.

As-is, DynamicStatusWatcher takes a specific ObjMetadataSet to watch. And while the DynamicStatusWatcher can be re-used, it doesn't share the informers between Watch calls.

The best option I can think of is to have the caller construct either a cluster-scoped watcher (more efficient) or a namespace-scoped watcher (fewer required permissions). This would allow cluster-scoped appliers and namespace-scoped appliers to be equally efficient, but it wouldn't allow for one applier to deploy to multiple namespaces without cluster-wide LIST/WATCH permission.

Would that be good enough for you? Or do you need a single applier to be able to deploy to multiple namespaces without cluster-wide LIST/WATCH permission?

FWIW, in Config Sync, the standard installation spins up a new reconciler (wrapping the cli-utils applier) for each RootSync/RepoSync. When it's for a RootSync, it needs cluster-admin, but when it's for a RepoSync it only needs namespace-admin. So we definitely need to support at least those use cases. 

As for sharing Watchers between Appliers, I hadn't really considered that, but it's similar to what @Liujingfang1 wanted to do in Config Sync to share the Watcher with another component. I think we would need two separate layers for that: one that starts/stops informers for all the required resource types in the required scopes, and another that calculates the status. That might be a cleaner design than doing both together, but it's gonna be a fair amount of work to rebuild.

ash2k:
@karlkfi 

> Would that be good enough for you? Or do you need a single applier to be able to deploy to multiple namespaces without cluster-wide LIST/WATCH permission?

Yes, I think allowing cli-utils consumer (me) to select between cluster-wide or namespaced mode is a good enough solution. Similarly, an app can let the user decide which of the approaches they need. It's a simple model and I like that.

> I think we would need two separate layers for that: one that starts/stops informers for all the required resource types in the required scopes, and another that calculates the status.

Yes, I think that would make sense. It could be a wrapper around informer factory that is doing "reference counting" for each informer or something like that. Maybe a new custom interface would be better. Also, this feels like what informer factory should already be, but it's dumber than that. Such "informer factory v2" would be great to have upstream - personally I needed it a few times over the years.

karlkfi:
Another dilemma: The current design doesn't work if you're applying a CR AND it's CRD is created async by another process AND there are no other CRDs in the apply set. Since the watcher wouldn't be watching CRDs, the CR informer would be stopped when it errors with NotFound, but never restarted.

I think I'll have to put some sort of retry-backoff loop in there to re-create the informer.

On the plus side, the Applier already rejects this use case when validating the object set. But we probably actually want to allow it, because it would fix a common problem where the applier doesn't work when the CRD is created indirectly through another object being created (ex: gatekeeper ConstraintTemplates dynamically create CRDs for Constraints).

That said, I'm not sure if there's some other reason we need the mapping up front, or whether it can be punted until apply time.

karlkfi:
Alright, I wrote a while new stack of abstractions bottom up... but I couldn't get it working. So after a few days of that, I gave up, took what I learned, and went back and modified the original design that I had here. This updated design is similar but more refined and solves the issues we brought up in comments (I hope).

The biggest change is that there is now an Options struct with a RESTScopeStrategy field, which takes 3 values: root, namespace, and automatic (default). In automatic mode, root-scope is used if the set of objects contains both root-scoped & namespace-scoped resources or multiple namespaces. This should work fine, as discussed, but if we need to allow the caller to force namespace-mode (because they want to apply to multiple namespaces without cluster-admin), we should be able to easily route the strategy option down to the Applier/Destroyer Options in the future.

Also, to make sure that the DynamicStatusReader wasn't gonna be a huge scale problem without caching or listing, I wrote another stress test that deploys 1,000 Deployments with 1 pod each. This means that each Deployment status lookup needs to make another couple GETs to get the ReplicaSet and Pod status as well. We can optimize this in the future, but it seems to work ok for now. Because of this change, I had to add a kind config that makes 10 worker nodes, instead of just 1. That was the easiest way to support 1,000 pods without a bunch of fiddling with IP ranges and resource capacity. I'm hoping this works in CI...

karlkfi:
Oh, one other big change: Because of some flaky tests, I figured out that the Informer was losing events, because it wasn't being fully synchronized before objects were being deleted. So the delete update never came and the prune wait task hangs. The workaround was to route through the Informer's `HasSynced` function and wrap it to send a `SyncEvent` on the status event channel, then make the TaskRunner block tasks until the StatusWatcher was synchronized (meaning all the started informers are synced). Hopefully this doesn't become a hazard at scale. But I think with the new optimization to use root-scoped LISTs there are MANY FEWER lists calls made in large cluster scenarios. Plus, the Informer automatically paginates them, which can help avoid locking up etcd and ruining performance. So even with the task start delay, the end-to-end time should still be much faster.

karlkfi:
Updated profiling numbers... 

## Time Spent (-1 QPS)

Polling: 3m25s
Watching: 3m12s
Diff: - 13s (- 6.34%)

## Memory Allocations (-1 QPS)

Polling: 7.34 GB
Watching: 1.60 GB
Diff: - 5.74 (- 78.2%)

karlkfi:
/unhold

This is ready for review again!

pkg/kstatus/watcher/watcher_dynamic.go
karlkfi: Done. All error events are now terminal.

karlkfi:
That’s another flaky test from the StatusWatcher being so fast the Pending WaitEvent was skipped. Easy fix is to pull out the Pending events as optional, as I’ve done elsewhere. Will fix tomorrow.

pkg/apply/taskrunner/runner.go
```diff
@@ -59,27 +57,30 @@ func (tsr *TaskStatusRunner) Run(
...
 		}
 		return err
 	}
 
 	// Find and start the first task in the queue.
-	currentTask, done := nextTask(taskQueue, taskContext)
-	if done {
-		return complete(nil)
-	}
+	// currentTask, done := nextTask(taskQueue, taskContext)
```
mortent, line 77: Should we delete this?

pkg/apply/taskrunner/runner.go
```diff
@@ -147,8 +167,10 @@ func (tsr *TaskStatusRunner) Run(
 
 			// send a status update to the running task, but only if the status
 			// has changed and the task is tracking the object.
-			if currentTask.Identifiers().Contains(id) {
-				currentTask.StatusUpdate(taskContext, id)
+			if currentTask != nil {
```
mortent, line 170: currentTask can be nil here when we are still waiting for the StatusWatcher to become synchronized right?

pkg/kstatus/watcher/event_funnel.go
```diff
@@ -0,0 +1,89 @@
...
+	// counterCh is used to track the number of open input channels
+	counterCh chan int
+}
+
+func newEventFunnel(ctx context.Context) *eventFunnel {
+	funnel := &eventFunnel{
+		ctx:       ctx,
+		outCh:     make(chan event.Event),
+		doneCh:    make(chan struct{}),
+		counterCh: make(chan int),
```
mortent, line 32: Maybe add a comment on the counterCh. It is not obvious how it is being used here.

mortent (commented):
Just some small comments/questions, but this looks great. Nice work!

I'll just lgtm for now to make sure others have a chance to take a look as well.
/lgtm

pkg/apply/taskrunner/runner.go
karlkfi: oops, yup.

pkg/apply/taskrunner/runner.go
karlkfi: Yeah, if we get a status update before full synchronization, which is likely when there are multiple resources being watched.

pkg/kstatus/watcher/event_funnel.go
karlkfi: Ah, yeah. I initially used a `sync.WaitGroup`, but it turns out those can't be incremented while being waited on. So I had to make my own method that could wait until zero, but could also be added to while waiting. 

I'll add a comment.

karlkfi:
Fixed the flaky test with a new `FilterOptionalEvents` added throughout the e2e tests.

karlkfi:
Comments addressed. Tests passed. Please take another look.

mortent (commented):
/lgtm

karlkfi:
/hold

Block until we get a release with k8s v1.24 out.

pkg/kstatus/watcher/object_status_reporter.go
```diff
@@ -0,0 +1,872 @@
...
+
+	handler.DeleteFunc = func(iobj interface{}) {
+		// Bail early if the context is cancelled, to avoid unnecessary work.
+		if ctx.Err() != nil {
+			return
+		}
+
+		if tombstone, ok := iobj.(cache.DeletedFinalStateUnknown); ok {
+			// Last state unknown. Possibly stale.
+			// TODO: Should we propegate this uncertainty to the caller?
```
ash2k, line 518: We probably should?

pkg/kstatus/watcher/object_status_reporter.go
```diff
@@ -0,0 +1,872 @@
...
+}
+
+// onCRDAdd handles creating a new informer to watch the new resource type.
+func (w *ObjectStatusReporter) onCRDAdd(obj *unstructured.Unstructured) {
+	gk, found := object.GetCRDGroupKind(obj)
+	if !found {
+		id := object.UnstructuredToObjMetadata(obj)
+		klog.Warningf("Invalid CRD added: missing group and/or kind: %v", id)
+		// Don't return an error, because this should not inturrupt the task queue.
+		// TODO: Allow non-fatal errors to be reported using a specific error type.
```
ash2k, line 561: 👍

pkg/kstatus/watcher/object_status_reporter.go
```diff
@@ -0,0 +1,872 @@
...
+			w.stopInformer(gkn)
+		})
+	default:
+		// Keep retrying
+		klog.Warningf("Watch error (will retry): %v: %v", gkn, err)
+	}
+}
+
+// resourceNotFoundMessage is the condition message for metav1.StatusReasonNotFound.
+const resourceNotFoundMessage = "the server could not find the requested resource"
```
ash2k, line 736: Would be nice to have a test that fails if/when the message is changed upstream.

pkg/kstatus/watcher/watcher.go
```diff
@@ -0,0 +1,38 @@
...
+// Options can be provided when creating a new StatusWatcher to customize the
+// behavior.
+type Options struct {
+	// RESTScopeStrategy specifies which strategy to use when listing and
+	// watching resources. By default, the strategy is selected automatically.
+	RESTScopeStrategy RESTScopeStrategy
+}
+
+//go:generate stringer -type=RESTScopeStrategy -linecomment
+type RESTScopeStrategy int
```
ash2k, line 32: nit: This can probably be `byte` =)

ash2k (commented):
Solid work! I've left a few comments. This PR adds quite a few TODOs. Are you planning to address those later?

pkg/kstatus/watcher/object_status_reporter.go
karlkfi: How tho? The callback interface informer uses is pretty terrible, because it uses `interface{}` objects as arguments, making it impossible to know what type you're going to get as a caller without looking at the implimentation. This particular pattern of wrapping the object with a struct with a getter is especially bad because it's not a standard object type. I didn't even know it was a possible edge case until an e2e test failed randomly.

The only idea I've had is to maybe add an annotation, but that would be equally hard to discover and would require removal by the receiver.

pkg/kstatus/watcher/watcher.go
karlkfi: I've never seen a go enum use byte. Int is more extensible. But perhaps you're suggesting that this particular enum will never be extended?

pkg/kstatus/watcher/event_funnel.go
```diff
@@ -0,0 +1,90 @@
...
+	go func() {
+		defer close(funnel.counterCh)
+		openInputChannels := 0
+	loop:
+		for {
+			select {
+			case delta := <-funnel.counterCh:
+				openInputChannels += delta
+			case <-ctx.Done():
+				if openInputChannels <= 0 {
```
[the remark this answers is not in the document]
karlkfi: good catch! 
If the context closes and the counter isn't zero, it should block on the next counter change.

pkg/kstatus/watcher/default_status_watcher.go
```diff
@@ -0,0 +1,164 @@
...
+	case RESTScopeRoot:
+		scope = meta.RESTScopeRoot
+		targets = rootScopeGKNs(ids)
+		klog.V(3).Infof("DynamicStatusWatcher starting in root-scoped mode (targets: %d)", len(targets))
+	case RESTScopeNamespace:
+		scope = meta.RESTScopeNamespace
+		targets = namespaceScopeGKNs(ids)
+		klog.V(3).Infof("DynamicStatusWatcher starting in namespace-scoped mode (targets: %d)", len(targets))
+	default:
+		panic(fmt.Sprintf("Invalid RESTScopeStrategy: %v", strategy))
```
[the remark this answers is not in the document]
karlkfi: It's tedious, with an event channel, but ok. Usually it's ok to panic on programmer errors, right?

pkg/kstatus/watcher/object_status_reporter.go
```diff
@@ -0,0 +1,872 @@
...
+//
+// ObjectStatusReporter is NOT repeatable. It will panic if started more than
+// once. If you need a repeatable factory, use DefaultStatusWatcher.
+//
+// TODO: support detection of added/removed api extensions at runtime
+// TODO: Watch CRDs & Namespaces, even if not in the set of IDs.
+// TODO: Retry with backoff if in namespace-scoped mode, to allow CRDs & namespaces to be created asynchronously
+type ObjectStatusReporter struct {
+	// InformerFactory is used to build informers
+	InformerFactory *DynamicInformerFactory
```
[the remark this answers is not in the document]
karlkfi: Seems like a reasonable enhancement, but we can probably wait to change it to use an interface until we add a 2nd impl. I try not to prematurely add interfaces, if I can avoid it. Sometime having the struct itself is more flexible, and in this case the coupling is pretty tight to the `SharedIndexInformer` impl details. For example, we know that passing an unstructured example means the callbacks will almost always be unstructured and can be safely cast.

karlkfi:
> This PR adds quite a few TODOs. Are you planning to address those later?

Honestly, if the watcher works well, I don't think I'll get back around to the TODOs for a while. The main goal here is performance/responsiveness and reduced memory use, and I don't want perfect to be the enemy of good. This codebase already has a pile of tech debt and the TODO issues I've made keep getting auto-closed, so I figured TODOs in the code were at least a little more persistent and might be discovered again, the next time someone is in this code.

I will address some of your comments tho. Thanks for reviewing!

pkg/kstatus/watcher/default_status_watcher.go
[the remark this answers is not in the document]
ash2k: Yes. I just thought this parameter is just a string and may come from a config file. Up to you :)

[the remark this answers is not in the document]
karlkfi: Given there's a method for returning errors, I'll use that. It's not that bad, just a few extra lines I can hide in a function.

In this case, the strategy input is actually an enum (int) with a generated String() func, which is being converted into a (dumb IMO) global struct wrapping a string. I don't know why the apimachinery scopes use an interface and global struct,  but I suspect it predates the `-linecomment` open in the stringer codegen, which helped standardize the way to do enums in Go. So I wrote this as a translation layer to use an enum, and also to add the "automatic" option as input (RESTScopeStrategy -> RESTScope).

pkg/kstatus/watcher/event_funnel.go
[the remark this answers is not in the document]
karlkfi: fixed

pkg/kstatus/watcher/default_status_watcher.go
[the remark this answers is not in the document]
karlkfi: fixed

pkg/kstatus/watcher/event_funnel.go
```diff
@@ -0,0 +1,111 @@
...
+		for {
+			select {
+			case delta, open := <-counterCh:
+				if !open {
+					counterCh = nil
+				} else {
+					inputs += delta
+					// Counter reached zero after context was closed.
+					if ctxDoneCh == nil && inputs <= 0 {
+						close(counterCh)
```
ash2k, line 55: It's risky to close a channel someone might send to as they will panic. With this change calling `AddInputChannel()` is not safe as it will panic even if context is done.

WDYT of this approach?
```go
func newEventFunnel(ctx context.Context) *eventFunnel {
	funnel := &eventFunnel{
		ctx:       ctx,
		outCh:     make(chan Event),
		doneCh:    make(chan struct{}),
		counterCh: make(chan int),
	}
	// Wait until the context is done. Then wait until the counter reaches zero,
	// indicating all input channels are closed. Then close the out and done
	// channels to signal completion.
	go func() {
		defer close(funnel.outCh)
		defer close(funnel.doneCh)
		ctxDoneCh := ctx.Done()

		// Count input channels that have been added and not closed.
		inputs := 0
		for {
			select {
			case delta := <-funnel.counterCh:
				inputs += delta
			case <-ctxDoneCh:
				// Stop waiting for context closure.
				ctxDoneCh = nil
			}
			if ctxDoneCh == nil && inputs <= 0 {
				// Context is closed and all input channels are closed.
				break
			}
		}
	}()
	return funnel
}
```

pkg/kstatus/watcher/object_status_reporter.go
karlkfi: I added a test for both List and Watch failure. Interestingly, the Watch code seems to correctly wrap errors, so `apierrors.IsNotFound(err)` works for that, but the List code wraps with a `%v`, not a `%w`, so we have to use both methods in the error handler.

FWIW, if a CRD is deleted while a Watch is executing, it looks like the Watch connection just closes, which will cause the ListAndWatch to return and retry. Then the List will error on the retry, triggering the Informer to be stopped by the ObjectStatusReporter error handler.

pkg/kstatus/watcher/event_funnel.go
karlkfi: > With this change calling AddInputChannel() is not safe as it will panic even if context is done.

With the current impl, if AddInputChannel is called after the context is closed, it won't try to send +1 on the counterCh:
```
select {
	case <-m.ctx.Done(): // skip, if context is closed
		return
	case m.counterCh <- 1: // increment counter
	}
```

It may look like they execute in parallel, but selects execute top down, only attempting subsequent cases if the previous cases block, the same as if statements operating left to right.

Your code is a little more elegant, except you forgot to close the counterCh.

karlkfi: Updated to use your approach with a few tweaks.

ash2k: > It may look like they execute in parallel, but selects execute top down, only attempting subsequent cases if the previous cases block, the same as if statements operating left to right.

https://go.dev/ref/spec#Select_statements says:

> If one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection.

Sometimes this crashes, sometimes succeeds https://go.dev/play/p/X62ydvz9HKj

> Your code is a little more elegant, except you forgot to close the counterCh.

That was the whole point :) Why do we need to close it? If we close it then we risk crashing blocked and future callers of `AddInputChannel()`.

karlkfi: > If one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection.

I stand corrected. 

> That was the whole point :) Why do we need to close it?

I always figured it was good practice to close channels for cleanup, but according to this, they don't need to be closed unless you're waiting for them to close. https://stackoverflow.com/a/36613932

karlkfi: Alright, I removed the `close(counterCh)` and added a `EventFunnelClosedError` to return from `AddInputChannel` if the context is closed. The handling of this error should help prevent informers from being started after the context is already cancelled.

ash2k (approved):
/lgtm

Thanks!

mortent (approved):
/approve

karlkfi:
/unhold

v0.30.0 passed all tests in Config Sync, so this this unblocked for merge.

PR #576, opened 2022-04-15
chore: Disable client-side throttling for tests

karlkfi:
Recent versions of kind have server-side throttling, and there's nothing else on the test cluster, so we can speed up the tests.


ash2k (approved):
/lgtm

mortent (approved):
/approve

PR #579, opened 2022-04-20
fix: Send Failed status when erroring

karlkfi:
- Caller should switch on Type == Failed, not Error != nil
- Error should always be non-nil for Failed and Skipped events


mortent (approved):
/approve

PR #581, opened 2022-04-20
fix: Sort reconcile events to fix flakey test

karlkfi:
Change the sorting algorithm to fix some flaky tests that became more frequent when client-side throttling was disabled.

With the current impl, only the WaitEvents have unpredictable ordering, reasoning is in comments in the sorting algorithm.

### DIFF
--- a/pkg/apply/applier_test.go
+++ b/pkg/apply/applier_test.go
@@ -343,15 +343,7 @@ func TestApplier(t *testing.T) {
 						Type:      event.Started,
 					},
 				},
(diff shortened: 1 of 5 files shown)


karlkfi:
This turns out to be a bigger problem than I thought. 

It turns out the Prune Success events are being sorted when they don't need to be, because they're (for now) consistently ordered by reverse apply order.

I'm going to have to fix the sorting algorithm again.

mortent (approved):
/approve
/lgtm

PR #584, opened 2022-04-27
feat: Use server-side throttling, if enabled

karlkfi:
- Disable client-side throttling, but only if server-side throttling is enabled

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -4,13 +4,18 @@
 package main
 
 import (
+	"context"
 	"flag"
+	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/client-go/rest"
 	"k8s.io/component-base/cli"
+	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/cmd/util"
 	"sigs.k8s.io/cli-utils/cmd/apply"
 	"sigs.k8s.io/cli-utils/cmd/destroy"
@@ -18,6 +23,7 @@ import (
 	"sigs.k8s.io/cli-utils/cmd/initcmd"
 	"sigs.k8s.io/cli-utils/cmd/preview"
 	"sigs.k8s.io/cli-utils/cmd/status"
+	"sigs.k8s.io/cli-utils/pkg/flowcontrol"
 	"sigs.k8s.io/cli-utils/pkg/inventory"
 	"sigs.k8s.io/cli-utils/pkg/manifestreader"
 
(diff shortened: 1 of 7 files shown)


cmd/main.go
```diff
@@ -7,23 +7,29 @@ import (
 	"flag"
 	"os"
 	"strings"
+	"time"
+
+	"context"
```
mortent, line 12: Nit: Why the linebreak between this and the previous ones?

cmd/main.go
```diff
@@ -7,23 +7,29 @@ import (
...
 	"sigs.k8s.io/cli-utils/cmd/preview"
 	"sigs.k8s.io/cli-utils/cmd/status"
+	"sigs.k8s.io/cli-utils/pkg/flowcontrol"
 	"sigs.k8s.io/cli-utils/pkg/inventory"
 	"sigs.k8s.io/cli-utils/pkg/manifestreader"
 
 	// This is here rather than in the libraries because of
 	// https://github.com/kubernetes-sigs/kustomize/issues/2060
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/rest"
```
mortent, line 32: Nit: Shoudl probably be with the other regular k8s imports above.

cmd/main.go
```diff
@@ -54,18 +62,24 @@ func main() {
...
+	previewCmd.PreRunE = preRunE
 	updateHelp(names, previewCmd)
 	diffCmd := diff.NewCommand(f, ioStreams)
+	diffCmd.PreRunE = preRunE
 	updateHelp(names, diffCmd)
 	destroyCmd := destroy.Command(f, invFactory, loader, ioStreams)
+	destroyCmd.PreRunE = preRunE
 	updateHelp(names, destroyCmd)
 	statusCmd := status.Command(f, invFactory, loader)
+	statusCmd.PreRunE = preRunE
```
mortent, line 82: Doesn't cobra provide a way to register a preRunE function on all commands, rather than having to add them to each individually? Alternatively, could we just loop over all the commands and register it?

mortent (approved):
/approve
Just a few nits.

cmd/main.go
karlkfi: fixed

cmd/main.go
karlkfi: fixed

cmd/main.go
karlkfi: > Doesn't cobra provide a way to register a preRunE function on all commands, rather than having to add them to each individually?

Yes, but we don't need it running for `help` or `completion`, since they don't talk tot he server.

> Alternatively, could we just loop over all the commands and register it?

Done.

mortent (approved):
/lgtm

PR #585, opened 2022-05-17
fix: Stop StatusWatcher on Forbidden API error

karlkfi:
- This matches previous StatusPoller behavior which would error
  and exit if there was a 403 Forbidden error from the apiserver.
- Handle status error before synchronization with immediate exit

Without this fix, the StatusWatcher gets stuck in an endless retry loop until cancelled or permission is granted, printing the following warning:

> W0516 21:03:57.305558   13045 reflector.go:324] pkg/mod/k8s.io/client-go@v0.24.0/tools/cache/reflector.go:167: failed to list /v1, Kind=ConfigMap: configmaps is forbidden: User "system:serviceaccount:default:user" cannot list resource "configmaps" in API group "" at the cluster scope

### DIFF
--- a/pkg/kstatus/watcher/dynamic_informer_factory.go
+++ b/pkg/kstatus/watcher/dynamic_informer_factory.go
@@ -5,6 +5,8 @@ package watcher
 
 import (
 	"context"
+	"regexp"
+	"strings"
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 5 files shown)


pkg/kstatus/watcher/dynamic_informer_factory.go
```diff
@@ -55,3 +57,34 @@ func (f *DynamicInformerFactory) NewInformer(ctx context.Context, mapping *meta.
...
+// https://github.com/kubernetes/apimachinery/blob/v0.24.0/pkg/api/errors/errors.go#L458
+// https://github.com/kubernetes/apimachinery/blob/v0.24.0/pkg/api/errors/errors.go#L208
+// https://github.com/kubernetes/apiserver/blob/master/pkg/endpoints/handlers/responsewriters/errors.go#L51
+const resourceForbiddenMessagePattern = `(.+) is forbidden: User "(.*)" cannot (.+) resource "(.*)" in API group "(.*)"`
+
+// resourceForbiddenMessageRegexp is the pre-compiled Regexp of
+// resourceForbiddenMessagePattern.
+var resourceForbiddenMessageRegexp = regexp.MustCompile(resourceForbiddenMessagePattern)
+
+// containsNotFoundMessage checks if the error string contains the message for
```
Liujingfang1, line 86: containsForbiddenMesssage

pkg/kstatus/watcher/dynamic_informer_factory.go
```diff
@@ -55,3 +57,34 @@ func (f *DynamicInformerFactory) NewInformer(ctx context.Context, mapping *meta.
 		f.Indexers,
 	)
 }
+
```
janetkuo, line 60: Suggest adding TODO to remove regex once client-go is updated to a version including the fix https://github.com/kubernetes/kubernetes/pull/110076 (1.25+).

karlkfi: done

pkg/kstatus/watcher/dynamic_informer_factory.go
karlkfi: fixed

PR #587, opened 2022-06-06
chore: add stress test that validates retries

karlkfi:
- Duplicate the 1,000 Deployment test, but use a 1m reconcile timeout in
  a retry loop.
- This verifies that the applier and destroyer are re-entrant at scale.

### DIFF
--- a/test/e2e/e2eutil/common.go
+++ b/test/e2e/e2eutil/common.go
@@ -14,6 +14,7 @@ import (
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
+	"github.com/onsi/gomega/gstruct"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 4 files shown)


test/stress/thousand_deployments_test.go
```diff
@@ -114,7 +114,7 @@ func thousandDeploymentsTest(ctx context.Context, c client.Client, invConfig inv
 		Expect(e.ErrorEvent.Err).To(BeNil())
 	}
 	for _, e := range destroyerEvents {
-		Expect(e.PruneEvent.Error).To(BeNil(), "PruneEvent: %v", e.PruneEvent)
+		Expect(e.DeleteEvent.Error).To(BeNil(), "DeleteEvent: %v", e.DeleteEvent)
```
janetkuo, line 117: Why change `PruneEvent` to `DeleteEvent`?

Is there a reason that we run the same for loop 3 times instead of combining them?

karlkfi: > Why change PruneEvent to DeleteEvent?

For the Destroyer there are no PruneEvents. It replaces them with DeleteEvents. The PruneEvents are only used by the Applier. This change fixes the test to catch the right event errors. 

Personally, I think they should all be DeleteEvents, since there's no real difference between them, but that's tech debt from years ago that we haven't gotten around to cleaning up.

> Is there a reason that we run the same for loop 3 times instead of combining them?

No particular reason. That was just the existing behavior. I could clean it up like I did in the other test. That would make it easier to debug by printing the whole event list.

karlkfi: added an function to clean it up

mortent: Yeah, distinguishing between PruneEvent and DeleteEvent seemed like a good idea at the time. But agree with @karlkfi that they should all just be DeleteEvents.

mortent (approved):
/approve

PR #588, opened 2022-06-07
fix: make deletion idempotent

karlkfi:
- Don't error when a delete returns NotFound.
  This means the object was deleted asynchronously by another client
  between the initial GET for task scheduling and the delete task itself.
  Log a warning, because it's possible the deletion may have violated
  dependency ordering.
- Update apply and delete task logs to only log errors if the verbosity
  is greater than -v=4. This should help reduce duplicate log messages.

### DIFF
--- a/pkg/apply/prune/prune.go
+++ b/pkg/apply/prune/prune.go
@@ -103,6 +103,10 @@ func (p *Pruner) Prune(
 		uid := obj.GetUID()
 		if uid == "" {
 			err := object.NotFound([]interface{}{"metadata", "uid"}, "")
+			if klog.V(4).Enabled() {
+				// only log event emitted errors if the verbosity > 4
+				klog.Errorf("prune uid lookup errored (object: %s): %v", id, err)
+			}
 			taskContext.SendEvent(eventFactory.CreateFailedEvent(id, err))
 			taskContext.InventoryManager().AddFailedDelete(id)
 			continue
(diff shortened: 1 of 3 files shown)


pkg/apply/prune/prune.go
```diff
@@ -103,6 +103,10 @@ func (p *Pruner) Prune(
 		uid := obj.GetUID()
 		if uid == "" {
 			err := object.NotFound([]interface{}{"metadata", "uid"}, "")
+			if klog.V(4).Enabled() {
+				// only log event emitted errors if the verbosity > 4
+				klog.Errorf("prune uid lookup errored (object: %s): %v", id, err)
```
Liujingfang1, line 108: What's the difference between this and klog.V(4).Error

karlkfi: Error doesn't exist on the versioned logger, only Info. Normally errors aren't specific to log level.
I could change it to Infof, but it IS an error...

karlkfi: Actually... it looks like it does exist, but is deprecated in favor of ErrorS...

```
// Deprecated: Use ErrorS instead.
func (v Verbose) Error(err error, msg string, args ...interface{}) {
	if v.enabled {
		logging.errorS(err, v.logr, logging.filter, 0, msg, args...)
	}
}

// ErrorS is equivalent to the global Error function, guarded by the value of v.
// See the documentation of V for usage.
func (v Verbose) ErrorS(err error, msg string, keysAndValues ...interface{}) {
	if v.enabled {
		logging.errorS(err, v.logr, logging.filter, 0, msg, keysAndValues...)
	}
}
```

Liujingfang1: Ack. We can change it to ErrorS later. No need to handle it in the PR.

karlkfi: I don't actually understand why `Error` is deprecated or why there's not an `Errorf`, but `ErrorS` seems like it has too many arguments for normal use...

karlkfi: actually, i take that back... ErrorS could work, but it accepts the error as the first arg, instead of requiring concatenation with %v... I guess in json format it puts the error in its own json field.

PR #599, opened 2022-08-03
Use builder for creating the destroyer

mortent:
Use the same builder pattern for creating the destroyer as we use for the applier.


ash2k (approved):
/lgtm

PR #600, opened 2022-08-03
Allow easier creation of DelegatedStatusReader with custom status readers

mortent:
Just to avoid requiring users to set up all the built-in status readers when they want to provide custom ones.

### DIFF
--- a/pkg/kstatus/polling/statusreaders/default.go
+++ b/pkg/kstatus/polling/statusreaders/default.go
@@ -20,19 +20,27 @@ import (
 // statusreaders to cover all built-in Kubernetes resources and other CRDs that
 // follow known status conventions.
 func NewDefaultStatusReader(mapper meta.RESTMapper) engine.StatusReader {
+	return NewStatusReader(mapper)
+}
+
+// NewStatusReader returns a DelegatingStatusReader that includes the statusreaders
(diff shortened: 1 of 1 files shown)


pkg/kstatus/polling/statusreaders/default.go
```diff
@@ -20,19 +20,27 @@ import (
...
 	replicaSetStatusReader := NewReplicaSetStatusReader(mapper, defaultStatusReader)
 	deploymentStatusReader := NewDeploymentResourceReader(mapper, replicaSetStatusReader)
 	statefulSetStatusReader := NewStatefulSetResourceReader(mapper, defaultStatusReader)
 
+	statusReaders = append(statusReaders, []engine.StatusReader{
+		deploymentStatusReader,
+		statefulSetStatusReader,
+		replicaSetStatusReader,
+		defaultStatusReader,
+	}...)
```
ash2k, line 40:
```suggestion
	statusReaders = append(statusReaders,
		deploymentStatusReader,
		statefulSetStatusReader,
		replicaSetStatusReader,
		defaultStatusReader,
	)
```
I think this will work the same way

mortent: Agree, that is better. I have updated the PR.

karlkfi (approved):
/lgtm

PR #601, opened 2022-08-05
Change printing functions to support the status output for other inventory types

chunglu-chou:
The change includes adding a status filter option and inventory name for event format status printer, and since the output structure will be changed for this update, the test cases are also modified to fit the new expected structure.

### DIFF
--- a/cmd/status/cmdstatus.go
+++ b/cmd/status/cmdstatus.go
@@ -13,6 +13,7 @@ import (
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"sigs.k8s.io/cli-utils/cmd/flagutils"
 	"sigs.k8s.io/cli-utils/cmd/status/printers"
+	"sigs.k8s.io/cli-utils/cmd/status/printers/printer"
 	"sigs.k8s.io/cli-utils/pkg/apply/poller"
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/cli-utils/pkg/inventory"
@@ -22,6 +23,7 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/cli-utils/pkg/manifestreader"
+	"sigs.k8s.io/cli-utils/pkg/object"
 )
 
 func GetRunner(factory cmdutil.Factory, invFactory inventory.ClientFactory, loader manifestreader.ManifestLoader) *Runner {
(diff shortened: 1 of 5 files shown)


cmd/status/cmdstatus.go
```diff
@@ -109,6 +111,17 @@ func (r *Runner) runE(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
+	printData := printer.PrintData{}
+	// initialize maps in printData
+	printData.InvNameMap = make(map[object.ObjMetadata]string)
```
mortent, line 116: I think it is more idomatic go to do the initialization while you are creating printData. So:
```golang
printData := printer.PrintData{
  InvNameMap: make(map[object.ObjMetadata]string),
  StatusSet: make(map[string]bool),
}
```

cmd/status/printers/event/printer.go
```diff
@@ -52,25 +63,27 @@ func (ep *Printer) Print(ch <-chan pollevent.Event, identifiers object.ObjMetada
...
-func (ep *Printer) printStatusEvent(se pollevent.Event) {
+func (ep *Printer) printStatusEvent(se pollevent.Event) error {
 	switch se.Type {
 	case pollevent.ResourceUpdateEvent:
 		id := se.Resource.Identifier
-		printResourceStatus(id, se, ep.IOStreams)
+		var invName string
+		var ok bool
+		if invName, ok = ep.Data.InvNameMap[id]; !ok {
+			return fmt.Errorf("resource not found")
```
mortent, line 73: This is more helpful if we also include the id of the problematic resource.

cmd/status/printers/event/printer.go
```diff
@@ -52,25 +63,27 @@ func (ep *Printer) Print(ch <-chan pollevent.Event, identifiers object.ObjMetada
...
 }
 
-func (ep *Printer) printStatusEvent(se pollevent.Event) {
+func (ep *Printer) printStatusEvent(se pollevent.Event) error {
 	switch se.Type {
 	case pollevent.ResourceUpdateEvent:
 		id := se.Resource.Identifier
-		printResourceStatus(id, se, ep.IOStreams)
+		var invName string
+		var ok bool
```
mortent, line 71: Do we need to define these here? It seems like we can just use the `:=` assignment in the if statements.

chunglu-chou: @mortent  For now since the invName variable is used outside of the if statement so it cannot be accessed if using the `:=` assignment in the if statement. I could also add an else statement and then move the following code into the else statement to get rid of the definitions outside, but I am not sure if having an else statement followed by an if statement that leads to return is a great idea or not. Do you have any suggestions on that?

mortent: Yeah, you are correct. I didn't notice the `invName` was used outside of the if statement

PR #602, opened 2022-08-09
Support the status output in table format for other inventory types

chunglu-chou:
Add a new column in the table output format representing the inventory name.

### DIFF
--- a/cmd/status/printers/table/adapter.go
+++ b/cmd/status/printers/table/adapter.go
@@ -4,6 +4,8 @@
 package table
 
 import (
+	"strings"
+
 	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/collector"
 	pe "sigs.k8s.io/cli-utils/pkg/kstatus/polling/event"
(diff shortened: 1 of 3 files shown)


cmd/status/printers/table/adapter.go
```diff
@@ -52,13 +56,16 @@ func (rss *ResourceState) Error() error {
 	return rss.err
 }
 
-func (ca *CollectorAdapter) LatestStatus() *ResourceState {
+func (ca *CollectorAdapter) LatestStatus(invNameMap map[object.ObjMetadata]string, statusSet map[string]bool) *ResourceState {
```
mortent, line 59: Do we have to provide this as parameters to this function? I don't think these are changed as part of the printing process, so it seems like it could be passed in when we create the `CollectorAdapter`?

chunglu-chou: Yeah that makes sense, they do not need to be passed as parameters. Thanks for pointing it out, I will update this.

mortent (approved):
/lgtm
/approve

PR #603, opened 2022-08-10
Add json printer

chunglu-chou:
Add json printer to support json output format including a new field named inventory-name.
Add testing data for json output format.


mortent:
Code looks good. Could we add a test for this functionality?

chunglu-chou:
@mortent Yes, I think I could reuse the test cases in the current one and just make it output as json format to verify the result.

mortent (approved):
/lgtm

PR #604, opened 2022-08-17
add Loader interface for easier customized implementation

chunglu-chou:
This change exposes an interface for loading the manifest, allowing easier development for customized implementations on the function for reading manifest and parsing it to inventory info.

### DIFF
--- a/cmd/status/cmdstatus.go
+++ b/cmd/status/cmdstatus.go
@@ -30,7 +30,7 @@ func GetRunner(factory cmdutil.Factory, invFactory inventory.ClientFactory, load
 	r := &Runner{
 		factory:           factory,
 		invFactory:        invFactory,
-		loader:            loader,
+		loader:            NewInventoryLoader(loader),
 		pollerFactoryFunc: pollerFactoryFunc,
 	}
 	c := &cobra.Command{
@@ -59,7 +59,7 @@ type Runner struct {
 	Command    *cobra.Command
 	factory    cmdutil.Factory
 	invFactory inventory.ClientFactory
-	loader     manifestreader.ManifestLoader
+	loader     Loader
 
 	period    time.Duration
 	pollUntil string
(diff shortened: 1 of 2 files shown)


pkg/manifestreader/manifestloader.go
```diff
@@ -6,13 +6,18 @@ package manifestreader
...
+	"sigs.k8s.io/cli-utils/cmd/flagutils"
+	"sigs.k8s.io/cli-utils/pkg/common"
+	"sigs.k8s.io/cli-utils/pkg/inventory"
 )
 
 // ManifestLoader is an interface for reading
 // and parsing the resources
 type ManifestLoader interface {
 	ManifestReader(reader io.Reader, path string) (ManifestReader, error)
+	GetInvInfo(cmd *cobra.Command, args []string) (inventory.Info, error)
```
mortent, line 20: We shouldn't mix the new Loader interface with the ManifestLoader interface. As mentioned in another comment, lets keep the new interface and the implementation with the command.

pkg/manifestreader/manifestloader.go
```diff
@@ -70,3 +98,7 @@ func mReader(path string, reader io.Reader, readerOptions ReaderOptions) Manifes
 	}
 	return mReader
 }
+
+type Loader interface {
```
mortent, line 102: I think we can just keep this interface and the implementation in the same package as the status cobra command. It is only used there, so it makes sense to keep it there.

cmd/status/cmdstatus.go
```diff
@@ -59,7 +59,7 @@ type Runner struct {
 	Command    *cobra.Command
 	factory    cmdutil.Factory
 	invFactory inventory.ClientFactory
-	loader     manifestreader.ManifestLoader
+	loader     *InventoryLoader
```
mortent, line 62: Shouldn't the type here be the Loader interface?

pkg/manifestreader/manifestloader.go
chunglu-chou: @mortent Yeah that makes sense, I already created a Loader interface with a GetInvInfo function to the status cobra command.

cmd/status/cmdstatus.go
chunglu-chou: @mortent Oh yes, it should be the Loader interface. I will change this and push a new one.

mortent (approved):
/lgtm

PR #605, opened 2022-08-19
add filtering and listing functionalities

chunglu-chou:
Add filter options on inventory names, namespaces, and statuses.
Add functions for retrieving a list of live inventory objects from the cluster.

### DIFF
--- a/cmd/status/cmdstatus.go
+++ b/cmd/status/cmdstatus.go
@@ -6,11 +6,13 @@ package status
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/util/slice"
 	"sigs.k8s.io/cli-utils/cmd/flagutils"
 	"sigs.k8s.io/cli-utils/cmd/status/printers"
 	"sigs.k8s.io/cli-utils/cmd/status/printers/printer"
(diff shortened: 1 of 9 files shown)


cmd/status/cmdstatus.go
```diff
@@ -66,49 +95,167 @@ type Runner struct {
...
+	PollerFactoryFunc func(cmdutil.Factory) (poller.Poller, error)
 }
 
-// runE implements the logic of the command and will delegate to the
-// poller to compute status for each of the resources. One of the printer
-// implementations takes care of printing the output.
-func (r *Runner) runE(cmd *cobra.Command, args []string) error {
+func (r *Runner) preRunE(*cobra.Command, []string) error {
+	if !slice.ContainsString(PollUntilOptions, r.pollUntil, nil) {
+		return fmt.Errorf("pollUntil must be one of %s, %s, %s, %s",
```
mortent, line 111: Nit: I think it would be easier to just use `strings.Join(PollUntilOptions, ",")` here

cmd/status/cmdstatus.go
```diff
@@ -66,49 +95,167 @@ type Runner struct {
...
+// runE implements the logic of the command and will delegate to the
+// poller to compute status for each of the resources. One of the printer
+// implementations takes care of printing the output.
+func (r *Runner) runE(cmd *cobra.Command, args []string) error {
+	var printData *printer.PrintData
+	var err error
+	switch r.invType {
+	case Local:
+		if len(args) != 0 {
+			fmt.Printf("%c[%dm", printcommon.ESC, printcommon.YELLOW)
```
mortent, line 238: I think you can simplify this with:
https://github.com/kubernetes-sigs/cli-utils/blob/eb8ebd73633e0ea5a2cd1406005f828118cef289/pkg/print/common/color.go#L25

mortent (commented):
Just a few small comments, but overall looks good.

mortent (approved):
/lgtm
/approve

PR #606, opened 2022-09-02
Update dependencies to Kubernetes v1.25.3

ash2k:


ash2k:
This is ready for review.

mortent (approved):
/approve

ash2k:
@seans3 @karlkfi PTAL

karlkfi:
Upgrading k8s libs is always a bit risky... 

For kpt & config-sync we're officially required to maintain compatibility with the last 3 Anthos/GKE minor versions, but the release channels still support the last 6 minor versions, and kubectl only officially supports the last 3 minor versions...

Practically speaking, the interfaces exposed by the k8s libraries aren't actually always reverse compatible tho, and they often require changes when you upgrade. So even if we didn't run into that here, a downstream consumer might not have an easy job of it. For example, we've had cases where interfaces were replaced/refactored and the new version no longer did what we used the old one for.

We've also been sort of unofficially putting off upgrading the version of Go, just to make sure no one adds usage of any reverse incompatible features (namely generics) that would require all downstream projects to also upgrade to the latest version of Go. But we haven't really discussed when it's safe to upgrade or when it's safe to start using generics.

That said, I hate to say no to upgrading...

Let me see if I can get someone to attempt to upgrade kpt and config-sync and see if those are going to run into any serious problems before we merge this.

ash2k:
@karlkfi friendly ping :)

karlkfi:
Yeah, sorry. We’re all kind drowning in broken and flakey e2e tests in config-sync and kpt right now. I haven’t found anyone with spare bandwidth to validate the upgrade  yet.

karlkfi:
Alright. I asked @rquitales to validate K8s v1.25 in kpt and ConfigSync and it looks like we also need to update Go to at least 1.18 and probably also golinter, but we don't want to do it all at the same time.

So @rquitales is working on a Go 1.18 bump for cli-utils, and then a seperate golangci-lint update, either to the latest 1.49 or 1.50, depending on what changed upstream.

Once that's done, either you can rebase this PR or @rquitales can make one that has just the k8s 1.25 changes in it.

We decided to stay at Go 1.18 for a while, instead of going to 1.19, just so downstream users of cli-utils and kpt aren't forced to upgrade, and so we don't accidentally use new 1.19 features that would force using 1.19 to compile. We'll probably eventually upgrade to 1.19, just not right now. Maybe when 1.20 come out.

ash2k:
@karlkfi Thanks for the update.

> Go 1.18 bump for cli-utils

I personally don't mind using 1.18 but [Kubernetes 1.25 uses 1.19](https://github.com/kubernetes/kubernetes/blob/v1.25.2/go.mod#L9) and that's why I [bumped the build here to 1.19](https://github.com/kubernetes/test-infra/pull/27354) too. So, I guess, if you want to go back, make sure to also downgrade the build. I don't know if 1.25 builds fine with 1.18, it may.

Makefile
```diff
@@ -19,16 +19,16 @@ all: generate license fix vet fmt test lint tidy
...
 "$(MYGOBIN)/ginkgo":
-	go install github.com/onsi/ginkgo/v2/ginkgo@v2.1.3
+	go install github.com/onsi/ginkgo/v2/ginkgo@v2.2.0
 
 "$(MYGOBIN)/mdrip":
 	go install github.com/monopole/mdrip@v1.0.2
 
 "$(MYGOBIN)/kind":
-	go install sigs.k8s.io/kind@v0.12.0
+	go install sigs.k8s.io/kind@v0.16.0
```
karlkfi, line 31: Is this the only place we have the kind version pinned?

karlkfi: nope, it's not.

PR #607, opened 2022-11-01
chore: Bump deepcopy-gen

rquitales:
This commit bumps geepcopy-gen to v0.25.3 to match upstream Kubernetes K8s resources when generating deepcopy methods.


ash2k:
I suppose this needs to go together with https://github.com/kubernetes-sigs/cli-utils/pull/606

rquitales:
> 

Yes, this PR was drafted for another pending PR, but it'll just be merged in with #606 instead.

PR #608, opened 2022-11-01
chore: Bump golangci-lint version

rquitales:
This commit bumps golangci-lint to v1.50.0 as the older version has memory issues that causes presubmits to time out.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,7 +10,6 @@ linters:
   disable-all: true
   enable:
     - bodyclose
-    - deadcode
     - depguard
     - dogsled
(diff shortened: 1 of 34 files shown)


karlkfi (approved):
/lgtm

.golangci.yml
```diff
@@ -10,7 +10,6 @@ linters:
   disable-all: true
   enable:
     - bodyclose
-    - deadcode
```
karlkfi, line 13: Why were these 3 checkers removed?

karlkfi (approved):
/lgtm

PR #612, opened 2022-12-22
fix: pass error to applyFailed events

manfredlift:
There was a bug where an error from an apply filter wasn't propagated to the applyFailed event, which leaded to printed events without error message set on them:
> {"group":"rbac.authorization.k8s.io","kind":"RoleBinding","name":"redacted","namespace":"redacted","status":"Failed","timestamp":"2022-12-21T12:43:46Z","type":"apply"}


karlkfi (approved):
/lgtm

Thanks!

Issue #613, opened 2023-01-06
Bug: objects flagged by CurrentUIDFilter should be removed from the inventory


manfredlift:
During pruning the code is filtering objects with `CurrentUIDFilter`:
> // CurrentUIDFilter implements ValidationFilter interface to determine
// if an object should not be pruned (deleted) because it has recently
// been applied.

This must mean that the object has ended up being tracked in the inventory by more than one reference and the one that was flagged by the filter should be removed from the inventory. Otherwise, it gets stuck in the inventory indefinitely and during every invocation of prune it'll be logged as skipped.

A good example of this is `Ingress` kind existing in both groups `extensions` and `networking.k8s.io` in k8s 1.19 & 1.20. This means when you rename the apiGroup in your local yaml to the new recommended group, then the reference with the old group name gets permastuck in the inventory.

PR #614, opened 2023-01-10
fix: objects flagged by CurrentUIDFilter should be removed from the inventory

manfredlift:
Fixes https://github.com/kubernetes-sigs/cli-utils/issues/613

### Description

During pruning the code is filtering objects with `CurrentUIDFilter`:
> // CurrentUIDFilter implements ValidationFilter interface to determine
// if an object should not be pruned (deleted) because it has recently // been applied.

This must mean that the object has ended up being tracked in the inventory by more than one reference and the one that was flagged by the filter should be removed from the inventory. Otherwise, it gets stuck in the inventory indefinitely and during every invocation of prune it'll be logged as skipped.

A good example of this is `Ingress` kind existing in both groups `extensions` and `networking.k8s.io` in k8s 1.19 & 1.20. This means when you rename the apiGroup in your local yaml to the new recommended group, then the reference with the old group name gets permastuck in the inventory.

### DIFF
--- a/pkg/apply/prune/prune.go
+++ b/pkg/apply/prune/prune.go
@@ -156,6 +156,7 @@ func (p *Pruner) Prune(
 				// Remove the object from inventory if it was determined that the object should not be pruned,
 				// because it had recently been applied. This probably means that the object is in the inventory
 				// more than one time with a different group (e.g. kind Ingress and apiGroups networking.k8s.io & extensions)
+				// due to being cohabitated: https://github.com/kubernetes/kubernetes/blob/v1.25.0/pkg/kubeapiserver/default_storage_factory_builder.go#L124-L131
 				var deleteAfterApplyErr *filter.ApplyPreventedDeletionError
 				if errors.As(filterErr, &deleteAfterApplyErr) {
 					if !opts.DryRunStrategy.ClientOrServerDryRun() {
--- a/pkg/apply/prune/prune.go
+++ b/pkg/apply/prune/prune.go
@@ -153,6 +153,17 @@ func (p *Pruner) Prune(
 					}
 				}
 
+				// Remove the object from inventory if it was determined that the object should not be pruned,
+				// because it had recently been applied. This probably means that the object is in the inventory
+				// more than one time with a different group (e.g. kind Ingress and apiGroups networking.k8s.io & extensions)
+				var deleteAfterApplyErr *filter.ApplyPreventedDeletionError
+				if errors.As(filterErr, &deleteAfterApplyErr) {
+					if !opts.DryRunStrategy.ClientOrServerDryRun() {
+						// Register for removal from the inventory.
+						taskContext.AddAbandonedObject(id)
+					}
+				}
+
 				taskContext.SendEvent(eventFactory.CreateSkippedEvent(obj, filterErr))
 				taskContext.InventoryManager().AddSkippedDelete(id)
 				break
--- a/pkg/apply/prune/prune_test.go
+++ b/pkg/apply/prune/prune_test.go
@@ -325,7 +325,7 @@ func TestPrune(t *testing.T) {
 				},
 			},
 		},
-		"UID match means prune skipped": {
+		"UID match means prune skipped and object abandoned": {
 			clusterObjs: []*unstructured.Unstructured{pod},
 			pruneObjs:   []*unstructured.Unstructured{pod},
 			pruneFilters: []filter.ValidationFilter{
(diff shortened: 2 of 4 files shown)


pkg/apply/prune/prune.go
```diff
@@ -153,6 +153,17 @@ func (p *Pruner) Prune(
 					}
 				}
 
+				// Remove the object from inventory if it was determined that the object should not be pruned,
+				// because it had recently been applied. This probably means that the object is in the inventory
+				// more than one time with a different group (e.g. kind Ingress and apiGroups networking.k8s.io & extensions)
```
karlkfi, line 158: Are you sure that `ApplyPreventedDeletionError` **always** means that there's a "duplicate" object in the source of truth? I'm concerned there may be side effects for other use cases that we're not testing for yet.

While I did re-write this check into the new Filter pattern, I'm not entirely sure what use case it was meant to handle in the first place.

Would it help if we tracked not just the UID but also the GVK that was applied, so it doesn't overlap when the API group is different?

manfredlift: > Are you sure that ApplyPreventedDeletionError always means that there's a "duplicate" object in the source of truth? I'm concerned there may be side effects for other use cases that we're not testing for yet.

I am not 100% sure, but can't come up with any other possible reasons for it!

> While I did re-write this check into the new Filter pattern, I'm not entirely sure what use case it was meant to handle in the first place.

This check originates from https://github.com/kubernetes-sigs/cli-utils/pull/113. Perhaps it'd be valuable to hear @seans3 thoughts on this change?

> Would it help if we tracked not just the UID but also the GVK that was applied, so it doesn't overlap when the API group is different?

I am not sure tracking UID+GVK is the best option as then we'll have two references to exactly the same object. E.g. in the apply set we'd apply `Ingress.networking.k8s.io` and then we'd prune `Ingress.extensions` right away as it's not filtered by `CurrentUIDFilter`. In reality it's the same object and the filter was created to prevent pruning objects just applied.

seans3: Where there are duplicate object references, one of them should be removed from the inventory. A duplicate reference is when multiple `ObjMetadata` (Group/Kind/Namespace/Name) point to the same object (by `UID`). The `UID` check exists specifically for this case (it also exists in `kubectl apply/prune`). The `UID` is the only metadata field which can definitively determine the uniqueness of an object.

As previously described, the most common instance this duplicate reference happens is when a `Deployment` or an `Ingress` is updated to the latest version from the `extensions` group and then applied. The `ObjMetatdata` check is unable to determine the objects are the same, but the `UID` check figures it out (and the old `ObjMetadata` reference should be removed).

I will LGTM and approve when we are all on the same page.

karlkfi: Gotcha. Sounds like something we need an end to end test for to make sure we don’t break it in future refactors. Maybe just a v1beta1 to v1 upgrade, for example. 

Right now, the uid filter skips the delete, but doesnt remove the old object from the inventory. 

I suspect there’s also a better place to put this code. The inv_set_task has code to decide what stays in the inventory using information from the inventory cache in the context. You don’t need to check error types there. 

In the future, I want to make the inventory update dynamically, but for now all that logic is in the inv_set_task.

karlkfi: On second thought, inv_set_task doesn’t know the difference between the types of skips. So maybe this way is fine.

karlkfi: @manfredlift can you add a uid filter test under test/e2e ?

we have namespace filter, skip invalid, and dependency filter e2e tests, but none for upgrading to a new resource version.

manfredlift: @karlkfi upgrading to a new version would not actually result in this issue as in the inventory we track objects using groupKind (without the version) **not** the GVK.

To demonstrate the issue, I added an e2e test that upgrades an `Event` (co-existing/co-habitating in both `v1` and `events.k8s.io/v1`) from `v1` to `events.k8s.io/v1`. This example is a bit contrived, but in k8s v1.25, which the e2e tests use, I could not any better examples (in the past the same scenario has existed for `Ingress` and `Deployment` though!)

manfredlift: @seans3 added an e2e test to demonstrate & verify the issue.

karlkfi: Would it make more sense to use 2 CRDs then, so it doesn’t break when k8s is upgraded?

manfredlift: I had the same thoughts initially, but...

AFAIK, CRDs don't support being cohabitated (https://github.com/kubernetes/kubectl/issues/186#issuecomment-369612606) where the object with different API groups is stored in storage only once (with same UID) and converted into either group on the fly. For CRDs, this support is only there between versions of the same API group.

These are some examples of cohabitating resources:
https://github.com/kubernetes/kubernetes/blob/8fdaac238e4f7f560066ea3a324cdbcaae564ac9/pkg/kubeapiserver/default_storage_factory_builder.go#L122-L129

karlkfi: Cool. That makes it more clear what we have this exception for and when we can retire it, assuming k8s doesn't add new ones in the future.

Can you add this link to that list in the filter description or e2e test comment?
https://github.com/kubernetes/kubernetes/blob/v1.26.0/pkg/kubeapiserver/default_storage_factory_builder.go#L124

seans3 (approved):
LGTM. Karl can give official LGTM label when he is ready.

pkg/apply/prune/prune.go
manfredlift: Done!

karlkfi (approved):
/lgtm

Thanks a lot for tracking down and fixing this regression!

PR #615, opened 2023-01-16
Bump dependencies

ash2k:
Most importantly Kubernetes 1.26 libraries.


ash2k:
PTAL @karlkfi @mortent

pkg/apply/applier.go
```diff
@@ -311,7 +311,7 @@ func handleError(eventChannel chan event.Event, err error) {
 // for the passed non cluster-scoped localObjs, plus the namespace
 // of the passed inventory object. This is used to skip deleting
 // namespaces which have currently applied objects in them.
-func localNamespaces(localInv inventory.Info, localObjs []object.ObjMetadata) sets.String {
+func localNamespaces(localInv inventory.Info, localObjs []object.ObjMetadata) sets.String { // nolint:staticcheck
```
mortent, line 314: What is the linter complaining about here? Is it something related to `sets.String`?

mortent (commented):
/lgtm

pkg/apply/applier.go
ash2k: Yes, it's deprecated.

https://github.com/kubernetes/apimachinery/blob/f357b1fa74b71ca0cffc090e447983950c2bafc5/pkg/util/sets/string.go#L17-L25

karlkfi (approved):
/lgtm

Thanks for bumping to k8s 1.26 without bumping the Go version.

PR #617, opened 2023-03-16
Allow to pass watcher strategy

ash2k:
Currently strategy is not specified and is always `RESTScopeAutomatic`. I have a case where I need to set it to `RESTScopeNamespace`, but there is no way to set it. This PR exposes an option to make it possible.

See https://github.com/kubernetes-sigs/cli-utils/pull/572#issuecomment-1111604485 and the following conversation for the background.


ash2k:
@karlkfi @seans3 @mortent PTAL

seans3 (approved):
Thanks for the addition.

/approve
/lgtm

PR #622, opened 2023-07-12
fix: retain inventory for pruning fail/skip

sdowell:
Prior to this change, the inventory always was deleted at the end of a Destroy event. This would occur even in the case of a pruning skip/failure, resulting in the objects being removed from the inventory without being deleted. This change makes it so that the inventory is only deleted if all objects have been deleted and reconciled.


pkg/apply/task/inv_set_task.go
```diff
@@ -54,79 +57,114 @@ func (i *InvSetTask) Identifiers() object.ObjMetadataSet {
...
+
+// Start will reconcile the state of the inventory, depending on the value of
+// Destroy.
+//
+// If Destroy is set, the intent is to delete the inventory. The inventory will
+// only be deleted if all prunes were successful (none failed/skipped). If any
+// prunes were failed or skipped, the inventory will be updated.
+//
+// If Destroy is false, the inventory will be updated.
+func (i *DeleteOrUpdateInvTask) Start(taskContext *taskrunner.TaskContext) {
```
karlkfi, line 149: i usually like to put public methods above private methods, since it's more likely someone came to the file to find them. interface before impl.

pkg/apply/task/inv_set_task.go
```diff
@@ -54,79 +57,114 @@ func (i *InvSetTask) Identifiers() object.ObjMetadataSet {
...
-		objStatus := taskContext.InventoryManager().Inventory().Status.Objects
-
-		klog.V(4).Infof("set inventory %d total objects", len(invObjs))
-		err := i.InvClient.Replace(i.InvInfo, invObjs, objStatus, i.DryRun)
-
-		klog.V(2).Infof("inventory set task completing (name: %q)", i.Name())
+		if i.Destroy {
+			skippedDeletes := taskContext.InventoryManager().SkippedDeletes()
+			failedDeletes := taskContext.InventoryManager().FailedDeletes()
+			// if any deletes were failed/skipped, retain the inventory
```
karlkfi, line 154: I suspect we should also keep the inventory if any reconciles failed.

This would slightly change the behavior of the ConfigSync finalizer, but the fact that CS ignores reconciler failures is a CS impl detail, not something we want to do in cli-utils or kpt.

In fact, because the CS finalizer is a preview annotation, we can still change how it works to wait for reconciles, even if the applier doesn't.

karlkfi: i think you can also clean up the code here a little by extracting a `success() bool` so you can just do `if destroy && !success() { err = deleteInventory() } else { err = updateInventory() }; ch <- result(err); return`

pkg/apply/task/inv_set_task.go
sdowell: done

pkg/apply/task/inv_set_task.go
sdowell: Done. Refactored the code and added reconcile failed/skipped check to the success condition

pkg/apply/solver/solver.go
```diff
@@ -211,26 +211,22 @@ func (t *TaskQueueBuilder) Build(taskContext *taskrunner.TaskContext, o Options)
...
-		tasks = append(tasks, &task.InvSetTask{
-			TaskName:      "inventory-set-0",
-			InvClient:     t.InvClient,
-			InvInfo:       t.invInfo,
-			PrevInventory: prevInvIds,
-			DryRun:        o.DryRunStrategy,
-		})
+	prevInvIds, _ := t.InvClient.GetClusterObjs(t.invInfo)
+	klog.V(2).Infoln("adding delete/update inventory task")
+	var setInventoryTaskName string
```
karlkfi, line 216: nit: just `taskName`

karlkfi (changes requested):
good start!

needs some e2e tests validating that destroy can fail, be retried, and succeed.

pkg/apply/solver/solver.go
sdowell: done

sdowell:
@karlkfi Added an e2e test and updated the logic to not consider delete/reconcile skips. I also added timeoutReconciles as a condition to consider a failure

pkg/apply/task/inv_set_task.go
```diff
@@ -54,79 +83,99 @@ func (i *InvSetTask) Identifiers() object.ObjMetadataSet {
...
+	if apierrors.IsNotFound(err) {
+		err = nil
+	}
+	klog.V(2).Infof("delete inventory task completing (name: %q)", i.Name())
+	return err
+}
 
-// StatusUpdate is not supported by the InvSetTask.
-func (i *InvSetTask) StatusUpdate(_ *taskrunner.TaskContext, _ object.ObjMetadata) {}
+func (i *DeleteOrUpdateInvTask) success(taskContext *taskrunner.TaskContext) bool {
```
karlkfi, line 167: func comment: success returns true when destroy actuation and reconciliation was fully successful. When true, it's safe to delete the inventory.

the name could maybe also be destroySuccessful, since it doesn't work for checking apply success.

test/e2e/e2e_test.go
```diff
@@ -191,6 +191,10 @@ var _ = Describe("E2E", func() {
 					pruneRetrieveErrorTest(ctx, c, invConfig, inventoryName, namespace.GetName())
 				})
 
+				It("PruneReconcileError", func() {
```
karlkfi, line 194: rename DestroyReconciliationFailure

test/e2e/e2eutil/common.go
```diff
@@ -198,6 +210,21 @@ func AssertUnstructuredExists(ctx context.Context, c client.Client, obj *unstruc
...
+	ref := mutation.ResourceReferenceFromUnstructured(obj)
+	resultObj := ref.ToUnstructured()
+
+	err := c.Get(ctx, types.NamespacedName{
+		Namespace: obj.GetNamespace(),
+		Name:      obj.GetName(),
+	}, resultObj)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred(),
+		"expected GET not to error (%s): %s", ref, err)
+	gomega.Expect(resultObj.GetDeletionTimestamp()).NotTo(gomega.BeNil(),
```
karlkfi, line 223: i dont know if deletion timestamp is always nil after parsing. there's an IsZero func I think. The k8s timestamp is a weird wrapper.

pkg/apply/task/inv_set_task.go
sdowell: done

test/e2e/e2eutil/common.go
sdowell: done

test/e2e/e2e_test.go
sdowell: done

pkg/apply/solver/solver.go
```diff
@@ -211,26 +211,22 @@ func (t *TaskQueueBuilder) Build(taskContext *taskrunner.TaskContext, o Options)
...
-			InvClient:     t.InvClient,
-			InvInfo:       t.invInfo,
-			PrevInventory: prevInvIds,
-			DryRun:        o.DryRunStrategy,
-		})
+	prevInvIds, _ := t.InvClient.GetClusterObjs(t.invInfo)
+	klog.V(2).Infoln("adding delete/update inventory task")
+	var taskName string
+	if o.Destroy {
+		taskName = "delete-inventory-0"
```
karlkfi, line 218: lets change the test name of this task too. It will require some find/replace, but I want to make it clear that it's not always a delete, because it shows up in logs. how about "inventory-delete-or-update-0"?

test/e2e/e2eutil/common.go
```diff
@@ -198,6 +210,21 @@ func AssertUnstructuredExists(ctx context.Context, c client.Client, obj *unstruc
...
+	resultObj := ref.ToUnstructured()
+
+	err := c.Get(ctx, types.NamespacedName{
+		Namespace: obj.GetNamespace(),
+		Name:      obj.GetName(),
+	}, resultObj)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred(),
+		"expected GET not to error (%s): %s", ref, err)
+	gomega.Expect(resultObj.GetDeletionTimestamp().IsZero()).To(gomega.BeFalse(),
+		"expected deletion timestamp to be non-nil (%s)", ref)
```
karlkfi, line 224: to be non-zero

sdowell: done

pkg/apply/solver/solver.go
sdowell: done

pkg/apply/solver/solver.go
```diff
@@ -211,34 +211,30 @@ func (t *TaskQueueBuilder) Build(taskContext *taskrunner.TaskContext, o Options)
...
+	})
 
 	return &TaskQueue{tasks: tasks}
 }
 
 // AppendApplyTask appends a task to the task queue to apply the passed objects
 // to the cluster. Returns a pointer to the Builder to chain function calls.
 func (t *TaskQueueBuilder) newApplyTask(applyObjs object.UnstructuredSet,
-	applyFilters []filter.ValidationFilter, applyMutators []mutator.Interface, o Options) taskrunner.Task {
+		applyFilters []filter.ValidationFilter, applyMutators []mutator.Interface, o Options) taskrunner.Task {
```
karlkfi, line 237: is this a format on save thing? I don't know if a go formatting standard was chosen for this project, but this looks inconsistent.

karlkfi: maybe a go version change in fmt?

sdowell: yeah probably my IDE being opinionated

sdowell: reverted

karlkfi (approved):
/lgtm

sdowell:
`Job execution failed: Pod got deleted unexpectedly`
/retest

PR #623, opened 2023-07-17
fix: retain objects which failed to reconcile

sdowell:
This change updates the set-inventory logic to retain objects which
failed to reconcile. This ensures that if you run the applier/destroyer
multiple times, an object that is failing to reconcile will be retained
in the inventory. Before this change, an object failing to reconcile
could be lost after multiple attempts (e.g. multiple destroys).

### DIFF
--- a/pkg/apply/task/inv_set_task.go
+++ b/pkg/apply/task/inv_set_task.go
@@ -131,6 +131,18 @@ func (i *DeleteOrUpdateInvTask) updateInventory(taskContext *taskrunner.TaskCont
 	klog.V(4).Infof("keep in inventory %d skipped prunes", len(pruneSkips))
 	invObjs = invObjs.Union(pruneSkips)
 
+	// If an object failed to reconcile and was previously stored in the inventory,
+	// then keep it in the inventory so it can be waited on next time.
+	reconcileFailures := i.PrevInventory.Intersection(im.FailedReconciles())
+	klog.V(4).Infof("set inventory %d failed reconciles", len(reconcileFailures))
+	invObjs = invObjs.Union(reconcileFailures)
+
+	// If an object timed out reconciling and was previously stored in the inventory,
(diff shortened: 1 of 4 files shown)


sdowell:
This change is based on https://github.com/kubernetes-sigs/cli-utils/pull/622

pkg/apply/task/inv_set_task.go
```diff
@@ -54,79 +83,113 @@ func (i *InvSetTask) Identifiers() object.ObjMetadataSet {
...
+	// then keep it in the inventory so it can be applied/pruned next time.
+	// It's likely that all the skipped deletes are already in the inventory,
+	// because the set of resources to prune comes from the inventory,
+	// but we're doing the intersection anyway just to be sure.
+	pruneSkips := i.PrevInventory.Intersection(im.SkippedDeletes())
+	klog.V(4).Infof("keep in inventory %d skipped prunes", len(pruneSkips))
+	invObjs = invObjs.Union(pruneSkips)
+
+	// If an object failed to reconcile and was previously stored in the inventory,
+	// then keep it in the inventory so it can be applied/pruned next time.
```
karlkfi, line 135: so it can be waited on next time

pkg/apply/task/inv_set_task_test.go
```diff
@@ -160,6 +162,16 @@ func TestInvSetTask(t *testing.T) {
 			invalidObjs:   object.ObjMetadataSet{idInvalid},
 			expectedObjs:  object.ObjMetadataSet{id3},
 		},
+		"object failed to reconcile": {
```
karlkfi, line 165: can you add permutations for both apply and delete reconcile failures?

sdowell: done

karlkfi (approved):
/lgtm

PR #624, opened 2023-07-18
test: reduce size of stress test deployments

sdowell:
The stress tests create 1000 copies of this deployment on the kind cluster, which means that these deployments are in contention for resources with the kind control plane and test suite. Reducing the size of these test deployments should help the stress test run with fewer resources available, and will hopefully get the presubmits passing.


karlkfi (approved):
/lgtm

PR #630, opened 2023-09-04
feat: add support for skipping reconciliation

manfredlift:
Example patch for https://github.com/kubernetes-sigs/cli-utils/issues/631

### DIFF
--- a/pkg/apply/prune/prune.go
+++ b/pkg/apply/prune/prune.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/cmd/util"
+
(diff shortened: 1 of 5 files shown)


pkg/apply/prune/prune.go
```diff
@@ -202,6 +206,13 @@ func (p *Pruner) Prune(
 		}
 		taskContext.InventoryManager().AddSuccessfulDelete(id, obj.GetUID())
 		taskContext.SendEvent(eventFactory.CreateSuccessEvent(obj))
+		if opts.MarkAsReconciled {
+			err := taskContext.InventoryManager().SetSuccessfulReconcile(id)
```
munnerz, line 210: Hm, I wonder if we should denote somehow that this has been skipped rather than marking it as successfully reconciled. That might allow higher-level libraries/functions/CLIs/whatever to better understand what happened during a run/application?

Issue #329, opened 2021-02-25
Status for Job in-progress


jlordiales:
Hi!
I'm curious about the reasoning behind https://github.com/kubernetes-sigs/cli-utils/blob/535f781a3c1b1d66b06a93f59e7a03c03af81477/pkg/kstatus/status/core.go#L488-L490, specifically `it will have the Current status while the job is running and after it has been completed successfully`.

Why isn't `InProgress` the returned status for a job that is currently in-progress? Wouldn't that be more intuitive if you want to poll until the job has either succeeded (status: `Current`) or failed (status: `Failed`)?

mortent:
@jlordiales 
So the primary goal for the kstatus library is to determine what we call "reconcile status", i.e. whether the desired state as expressed by a KRM resource has been fully realized in the cluster. In practice this means we want to report whether the controllers have completed their work on the resource. This is pretty intuitive for resources like Deployments, but trickier for a resource like Job. But to be consistent, a Job is considered reconciled if it has started (there is additional complexity here around how we should treat a job that starts but fails almost immediately..).

But you are not the first to ask about this, so we do want to make this configurable. We don't have a design for this yet, but I'm thinking that we will support an annotation that users can apply to a job and this will cause kstatus to wait until the job has completed.

jlordiales:
Got it! Looking forward to the configuration option for this. 

I'll go ahead and close this issue. Thanks!

PR #633, opened 2023-11-14
pod.status.phase = Failed (incorrectly?) marked as `CurrentStatus` 

oistein:
I was trying to compute status of a pod and to my surprise it came back as `Current` when it really had failed. Given that this code has been here for almost 3 years, this might be on purpose? However, if not, please review this PR.

### DIFF
--- a/pkg/kstatus/status/core.go
+++ b/pkg/kstatus/status/core.go
@@ -422,7 +422,7 @@ func podConditions(u *unstructured.Unstructured) (*Result, error) {
 		}, nil
 	case "Failed":
 		return &Result{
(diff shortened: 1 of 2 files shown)


mortent:
@oistein So this is on purpose, although there are several alternative ways to handle this. The idea is that what kstatus computes is reconcile status, i.e. were the controller able to realize the desired state as expressed by a resource. And for a Pod, we consider that the desired state has been realized when the Pod is running, meaning that whether the Pod completed successfully or not is a consideration. A similar discussion is availabe in https://github.com/kubernetes-sigs/cli-utils/issues/329

PR #634, opened 2023-11-27
Properly discover HPA status

makkes:
This commit adds a conditions function for the HorizontalPodAutoscaler kind. It follows the definition documented
[here](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#appendix-horizontal-pod-autoscaler-status-conditions).

### DIFF
--- a/pkg/kstatus/status/core.go
+++ b/pkg/kstatus/status/core.go
@@ -20,21 +20,22 @@ type GetConditionsFn func(*unstructured.Unstructured) (*Result, error)
 // legacyTypes defines the mapping from GroupKind to a function that can
 // compute the status for the given resource.
 var legacyTypes = map[string]GetConditionsFn{
-	"Service":                    serviceConditions,
-	"Pod":                        podConditions,
-	"Secret":                     alwaysReady,
-	"PersistentVolumeClaim":      pvcConditions,
-	"apps/StatefulSet":           stsConditions,
-	"apps/DaemonSet":             daemonsetConditions,
-	"extensions/DaemonSet":       daemonsetConditions,
-	"apps/Deployment":            deploymentConditions,
-	"extensions/Deployment":      deploymentConditions,
-	"apps/ReplicaSet":            replicasetConditions,
-	"extensions/ReplicaSet":      replicasetConditions,
-	"policy/PodDisruptionBudget": pdbConditions,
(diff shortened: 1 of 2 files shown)


makkes:
@karlkfi would you be able to take a look at this, please? 🙏🏻

pkg/kstatus/status/core.go
```diff
@@ -74,6 +75,58 @@ func alwaysReady(u *unstructured.Unstructured) (*Result, error) {
 	}, nil
 }
 
+func hpaConditions(u *unstructured.Unstructured) (*Result, error) {
```
karlkfi, line 78: Can you add a comment that describes what this is expecting for HPA to be detected as Current?

pkg/kstatus/status/status_compute_test.go
```diff
@@ -1269,6 +1269,85 @@ func TestJobStatus(t *testing.T) {
 	}
 }
 
+var hpaNoStatus = `
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
+metadata:
+  name: test
+  namespace: qual
+  generation: 1
```
karlkfi, line 1278: Does HPA have an observedGeneration?

Usually we expect that observedGeneration must equal generation, otherwise the object is still reconciling.

Ideally the CRD would require observedGeneration and default it to zero, that way the status is always defined after being applied, even if the controller hasn't updated the status yet.

pkg/kstatus/status/status_compute_test.go
```diff
@@ -1269,6 +1269,85 @@ func TestJobStatus(t *testing.T) {
...
+    status: "True"
+  - type: ScalingLimited
+    status: "True"
+`
+
+func TestHPAStatus(t *testing.T) {
+	testCases := map[string]testSpec{
+		"hpaNoStatus": {
+			spec:               hpaNoStatus,
+			expectedStatus:     UnknownStatus,
```
karlkfi, line 1317: Usually when an object doesn't have status, it's shown as Current, with the assumption that there might not be any status. The way to work around that is to default `status.observedGeneration=0` in the CRD. Is that not true for the HPA CRD?

I wonder if we should get that changed upstream.

karlkfi (commented):
Sorry to let this rot. Added some review comments.

PR #635, opened 2023-12-24
fix: treat rate limiter deadline as context deadline

levenleven:
We often bump into race condition where resource status [won't be read](https://github.com/kubernetes/client-go/blob/5a0a4247921dd9e72d158aaa6c1ee124aba1da80/rest/request.go#L614) because context is [reaching the deadline](https://cs.opensource.google/go/x/time/+/refs/tags/v0.5.0:rate/rate.go;l=275). This results in `Unknown` status being returned.

This PRs treats this edge case the same way as `context.Canceled` and `context.DeadlineExceeded` are treated.

### DIFF
--- a/pkg/kstatus/polling/statusreaders/common.go
+++ b/pkg/kstatus/polling/statusreaders/common.go
@@ -167,7 +167,7 @@ func errResourceToResourceStatus(err error, resource *unstructured.Unstructured,
 	// If the error is from the context, we don't attach that to the ResourceStatus,
 	// but just return it directly so the caller can decide how to handle this
 	// situation.
-	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) || isRateLimiterContextDeadlineExceeded(err) {
 		return nil, err
 	}
 	identifier := object.UnstructuredToObjMetadata(resource)
@@ -193,7 +193,7 @@ func errIdentifierToResourceStatus(err error, identifier object.ObjMetadata) (*e
 	// If the error is from the context, we don't attach that to the ResourceStatus,
 	// but just return it directly so the caller can decide how to handle this
 	// situation.
-	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) || isRateLimiterContextDeadlineExceeded(err) {
 		return nil, err
 	}
 	if apierrors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


pkg/kstatus/polling/statusreaders/common.go
```diff
@@ -209,3 +209,10 @@ func errIdentifierToResourceStatus(err error, identifier object.ObjMetadata) (*e
 		Error:      err,
 	}, nil
 }
+
+func errWouldExceedContextDeadline(err error) bool {
+	if e := errors.Unwrap(err); e != nil {
+		return e.Error() == "rate: Wait(n=1) would exceed context deadline"
```
ash2k, line 215: This is very fragile. There should be at least a test for this. Ideally the wait package should have a type for this.

Also, this doesn't feel like the right spot for this logic.

levenleven: Agree, ideally `rate` package would return a typed error here

levenleven: Added a test.

> Also, this doesn't feel like the right spot for this logic.

Do you have a suggestion what would be a better place?

ash2k: Maybe somewhere up the call stack. I'm not a maintainer of this project, just a user, so this may be fine with maintainers.

karlkfi:
FWIW, we mostly avoid this problem by disabling client-side throttling and depending on server-side throttling to handle slowing down requests. If you do the same, you shouldn't ever see any `rate: Wait(n=%d) would exceed context deadline` errors.

Unfortunately, that error doesn't have a type, which makes it fragile to catch. And unwrapping only once is also fragile. You would need to either fully unwrap recursively or check for a partial match in the full error message to catch all the edge cases.

That said, what is the behavior you're looking for? Won't it still be status unknown if the context is cancelled?

levenleven:
@karlkfi Thanks for looking into this!

> That said, what is the behavior you're looking for? Won't it still be status unknown if the context is cancelled?

No, the poller would keep the previous known status. With current behavior the error is swallowed and the status is overridden with `Unknown`.

> You would need to either fully unwrap recursively or check for a partial match in the full error message to catch all the edge cases.

If this has a chance to be accepted I can do that 🙂

pkg/kstatus/polling/statusreaders/common.go
```diff
@@ -209,3 +209,15 @@ func errIdentifierToResourceStatus(err error, identifier object.ObjMetadata) (*e
...
+func errWouldExceedContextDeadline(err error) bool {
+	for {
+		next := errors.Unwrap(err)
+		if next == nil {
+			break
+		}
+		err = next
+	}
+
+	return err != nil && err.Error() == "rate: Wait(n=1) would exceed context deadline"
```
levenleven, line 222: @karlkfi unwrapped recursively, could you have a look?

pkg/kstatus/polling/statusreaders/common.go
```diff
@@ -209,3 +209,15 @@ func errIdentifierToResourceStatus(err error, identifier object.ObjMetadata) (*e
 		Error:      err,
 	}, nil
 }
+
+func errWouldExceedContextDeadline(err error) bool {
```
karlkfi, line 213: - rename: `isRateLimiterContextDeadlineExceeded`
- add a method comment explaining why we need this and what it does
- link to the problem in the commit: https://cs.opensource.google/go/x/time/+/refs/tags/v0.10.0:rate/rate.go;l=276

pkg/kstatus/polling/statusreaders/common.go
karlkfi: i dont love it, but it'll work.

pkg/kstatus/polling/statusreaders/common.go
levenleven: @karlkfi https://github.com/kubernetes-sigs/cli-utils/pull/635/commits/31ee857a41bc4f740ea5a87076357455d9e0f79a looks ok?

karlkfi (approved):
Looks good. Thanks!

levenleven:
@karlkfi sorry for another ping, looks like `lgtm` label is required?

PR #637, opened 2024-03-15
Add conversion func for status-policy

fsommar:
Hi 👋 

This change is intended for Kpt. We use `kpt live apply` in our CI/CD pipelines, but we don't need it saving statuses for all resources into the inventory. We want to be able to pass `--status-policy=none`.

I considered adding support for changing the status-policy flag in kapply as well, but I couldn't think of a way to introduce the flag without introducing it for all subcommands, or copying the description across for the applicable subcommands. Let me know if you want me to implement the flag here too.

### DIFF
--- a/cmd/flagutils/utils.go
+++ b/cmd/flagutils/utils.go
@@ -15,6 +15,9 @@ const (
 	InventoryPolicyStrict     = "strict"
 	InventoryPolicyAdopt      = "adopt"
 	InventoryPolicyForceAdopt = "force-adopt"
+	StatusPolicyFlag          = "status-policy"
+	StatusPolicyAll           = "all"
+	StatusPolicyNone          = "none"
 )
 
 // ConvertPropagationPolicy converts a propagationPolicy described as a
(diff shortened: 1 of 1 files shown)


justinsb:
Thanks @fsommar 

/approve
/lgtm

(Let's see if I have permissions here!)

fsommar:
Thank you @mortent! Can I get help with creating a tag to be used in https://github.com/kptdev/kpt/pull/4129?

cc @justinsb

karlkfi:
I'll be doing some other changes for watch filtering soon, so I'll cut a cli-utils release afterwards.

karlkfi:
As for kapply, I think adding a global flag is probably fine, if you add it and tests to exercise it.

tomasaschan:
Pending that cli-utils release, will kptdev/kpt#4129 be accepted if it pins cli-utils to the merge commit of this change, even if there is no release that has this change yet?

(We've been blocked on these PRs for over a month, and it doesn't feel like fun to block even more on some other, unrelated, work...)

fsommar:
There's a v0.36.0 of cli-utils released. I've rebased my work in kptdev/kpt on top of that: https://github.com/kptdev/kpt/pull/4129/commits/5c792649490fb69df8acb7d1c2d7104a88369d0a.

PR #641, opened 2024-06-12
Bump deps

ash2k:
Please see individual commits.

Kubernetes 1.30 uses Go 1.22, so I updated the required Go version too.

I also ran an import formatter to fix the imports after the find/replace of the deprecated type. I hope it's ok. Formatter command was `go run github.com/daixiang0/gci@v0.13.4 write cmd pkg -s standard -s default`.

Type import change rename - [it's an alias](https://github.com/kubernetes/cli-runtime/blob/v0.30.1/pkg/genericclioptions/io_options.go#L29), so the same type, and hence it's not a breaking change.


ash2k:
Opened https://github.com/kubernetes/test-infra/pull/32742 to bump the used Go version.

mortent (commented):
/lgtm

PR #642, opened 2024-06-20
feat: allow disabling/overriding ObjectFilter in status watcher

floreks:
With the addition of `Filters` to the `DefaultStatusWatcher` it is now possible to apply server-side filtering via labels/fields. This PR allows dropping or overriding client-side filtering with a custom `ObjectFilter`. In our case we are making sure that every resource in the inventory will contain a static label that can be used for server-side filtering and further client-side filtering is obsolete.

### DIFF
--- a/pkg/kstatus/watcher/default_status_watcher.go
+++ b/pkg/kstatus/watcher/default_status_watcher.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
+
 	"sigs.k8s.io/cli-utils/pkg/kstatus/polling/clusterreader"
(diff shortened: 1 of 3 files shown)


floreks:
There might not be many use cases for the custom implementation of `ObjectFilter`. Being able to disable it completely when relying on server-side filtering is what I was looking for.  If that would be something you would be interested in, let me know and I will adjust the code.

PS. From what I can see current `AllowListObjectFilter` is simply going through the array and doing a struct comparison.
https://github.com/kubernetes-sigs/cli-utils/blob/4d57e6e23edac29f2ea3ae9aa8b0ff0a85fc01cc/pkg/kstatus/watcher/object_filter.go#L27-L30
https://github.com/kubernetes-sigs/cli-utils/blob/4d57e6e23edac29f2ea3ae9aa8b0ff0a85fc01cc/pkg/object/objmetadata_set.go#L56-L62

Issue #487, opened 2021-11-25
[Bug]: pkg/object/ParseObjMetadata incorrectly assumes rbac resource cannot have underscore in its name


SomtochiAma:
If a clusterrole / binding has an underscore in its name (e.g roles_ops which is valid in kubernetes), Using the [ParseObjMetadata](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/object/objmetadata.go#L89) function failes with 
```
too many fields within: 
```

This can be resolved by replacing the underscore with a double underscore (like it is [here](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/object/objmetadata.go#L139) done for colon) or removing the check for underscores in the function.

karlkfi:
/remove-lifecycle rotten

Still a bug: 
https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/object/objmetadata.go#L96

PR #645, opened 2024-08-22
Fix: Ignore extra underscored when Object is an RBAC object in objmetadata

liamchenoweth:
Ref. #487
Skips the test for extra underscores when we are dealing with RBAC objects as they are only restricted to path objects for naming. Included new tests for this new use case

### DIFF
--- a/pkg/object/objmetadata_test.go
+++ b/pkg/object/objmetadata_test.go
@@ -102,6 +102,30 @@ func TestParseObjMetadata(t *testing.T) {
 			inventory: &ObjMetadata{},
 			isError:   true,
 		},
+		"RBAC with underscores": {
+			invStr: "test-namespace_leader_locking_kube_scheduler___rbac.authorization.k8s.io_RoleBinding",
+			inventory: &ObjMetadata{
+				Namespace: "test-namespace",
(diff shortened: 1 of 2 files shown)


karlkfi:
This seems like it might break the ObjMetadata string format which uses underscores as segment delimiters and double underscores as colons. With this change, an RBAC object with double underscores wouldn't make it through a format+parse cycle. Maybe that's an acceptable limitation in order to support single underscores in RBAC names?

I'm not really sure why underscores were originally chosen. It seems like backslashes would have been a better choice, like we used for the depends-on and apply-time-mutation annotations.

Changing the ID format might be possible, but would probably require migration. And it might prohibit downgrades unless we shipped bi-directional migration beforehand, or made it an opt-in feature for a while.

pkg/object/objmetadata_test.go
```diff
@@ -102,6 +102,30 @@ func TestParseObjMetadata(t *testing.T) {
 			inventory: &ObjMetadata{},
 			isError:   true,
 		},
+		"RBAC with underscores": {
```
karlkfi, line 105: TestParseObjMetadata only tests parsing, unfortunately, not round-trip format+parse. I think this change would break the round-trip.

PR #648, opened 2024-12-04
chore: clean up inventory Client interface

sdowell:
The Client interface had several repetetive and/or extraneous methods which would make it more cumbersome to create different implementations of the interface. This trims down the interface to a core set of methods which are necessary for the applier.

Removed Client methods:
- ApplyInventoryNamespace -> InvAddTask.createNamespace
- GetClusterInventoryInfo -> ClusterClient.getClusterInventoryInfo
- GetClusterInventoryObjs -> ClusterClient.getClusterInventoryObjs
- ListClusterInventoryObjs -> Runner.listClusterInventoryObjs

### DIFF
--- a/pkg/apply/applier.go
+++ b/pkg/apply/applier.go
@@ -64,26 +64,6 @@ func (a *Applier) prepareObjects(localInv inventory.Info, localObjs object.Unstr
 	for _, localObj := range localObjs {
 		inventory.AddInventoryIDAnnotation(localObj, localInv)
 	}
-	// If the inventory uses the Name strategy and an inventory ID is provided,
-	// verify that the existing inventory object (if there is one) has an ID
-	// label that matches.
-	// TODO(seans): This inventory id validation should happen in destroy and status.
-	if localInv.Strategy() == inventory.NameStrategy && localInv.ID() != "" {
-		prevInvObjs, err := a.invClient.GetClusterInventoryObjs(localInv)
-		if err != nil {
-			return nil, nil, err
-		}
-		if len(prevInvObjs) > 1 {
-			panic(fmt.Errorf("found %d inv objects with Name strategy", len(prevInvObjs)))
-		}
-		if len(prevInvObjs) == 1 {
-			invObj := prevInvObjs[0]
-			val := invObj.GetLabels()[common.InventoryLabel]
-			if val != localInv.ID() {
-				return nil, nil, fmt.Errorf("inventory-id of inventory object in cluster doesn't match provided id %q", localInv.ID())
-			}
-		}
-	}
 	pruneObjs, err := a.pruner.GetPruneObjs(localInv, localObjs, prune.Options{
 		DryRunStrategy: o.DryRunStrategy,
 	})
--- a/pkg/inventory/inventory-client.go
+++ b/pkg/inventory/inventory-client.go
@@ -16,7 +16,6 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/klog/v2"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
-	"k8s.io/kubectl/pkg/util"
 	"sigs.k8s.io/cli-utils/pkg/apis/actuation"
 	"sigs.k8s.io/cli-utils/pkg/common"
 	"sigs.k8s.io/cli-utils/pkg/object"
@@ -39,12 +38,6 @@ type Client interface {
 	Replace(inv Info, objs object.ObjMetadataSet, status []actuation.ObjectStatus, dryRun common.DryRunStrategy) error
 	// DeleteInventoryObj deletes the passed inventory object from the APIServer.
 	DeleteInventoryObj(inv Info, dryRun common.DryRunStrategy) error
-	// ApplyInventoryNamespace applies the Namespace that the inventory object should be in.
-	ApplyInventoryNamespace(invNamespace *unstructured.Unstructured, dryRun common.DryRunStrategy) error
-	// GetClusterInventoryInfo returns the cluster inventory object.
-	GetClusterInventoryInfo(inv Info) (*unstructured.Unstructured, error)
-	// GetClusterInventoryObjs looks up the inventory objects from the cluster.
-	GetClusterInventoryObjs(inv Info) (object.UnstructuredSet, error)
 	// ListClusterInventoryObjs returns a map mapping from inventory name to a list of cluster inventory objects
 	ListClusterInventoryObjs(ctx context.Context) (map[string]object.ObjMetadataSet, error)
 }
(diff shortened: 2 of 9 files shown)


cmd/status/cmdstatus.go
```diff
@@ -374,3 +372,42 @@ func (ir *InventoryLoader) GetInvInfo(cmd *cobra.Command, args []string) (invent
...
+	if err != nil {
+		return nil, err
+	}
+
+	// retrieve the list from the cluster
+	clusterInvs, err := dc.Resource(mapping.Resource).List(ctx, metav1.ListOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, err
+	}
+	if apierrors.IsNotFound(err) {
```
karlkfi, line 396: nit: This pattern works, and technically has less cyclomatic complexity, but i usually find it more readable to do the IsNotFound check inside the err != nil clause.

karlkfi: ah, i see you just moved and renamed this. NBD.

cmd/status/cmdstatus.go
```diff
@@ -374,3 +372,42 @@ func (ir *InventoryLoader) GetInvInfo(cmd *cobra.Command, args []string) (invent
 	inv := inventory.WrapInventoryInfoObj(invObj)
 	return inv, nil
 }
+
+func (r *Runner) listClusterInventoryObjs(ctx context.Context) (map[string]object.ObjMetadataSet, error) {
```
karlkfi, line 376: I was hoping we could hide the listing of inventory objects inside the inventory client, making it an implementation detail the caller doesn't know about. But maybe that needs to be a subsequent change?

The `map[string]object.ObjMetadataSet` here looks like it's just being looped through to get a full list of "Identifiers" for the `PrintData` built by `listInvFromCluster`. That flattening could just be done inside the inventory client.

That would require changing the `PrintData` to remove the `InvNameMap` and then updating `printStatusEvent`  in the json & event printers. Then instead of printing the actual inventory object name, we print the ID that might be the name or label.

I don't see anywhere in kpt that uses the "inventory-name" field from the json printer. So it probably wouldn't break kpt itself. That said, I have no idea if anyone uses the json output from kpt live status for integration purposes. Even so, the status event output doesn't have a field for "inventory-namespace", so it would be hard for a caller to depend on the name to do their own GET or anything.

pkg/apply/applier.go
```diff
@@ -64,26 +64,6 @@ func (a *Applier) prepareObjects(localInv inventory.Info, localObjs object.Unstr
 	for _, localObj := range localObjs {
 		inventory.AddInventoryIDAnnotation(localObj, localInv)
 	}
-	// If the inventory uses the Name strategy and an inventory ID is provided,
-	// verify that the existing inventory object (if there is one) has an ID
-	// label that matches.
-	// TODO(seans): This inventory id validation should happen in destroy and status.
-	if localInv.Strategy() == inventory.NameStrategy && localInv.ID() != "" {
```
karlkfi, line 71: was this validation moved somewhere?

karlkfi: Does the new code support both NameStrategy and LabelStrategy?

pkg/inventory/inventory-client.go
```diff
@@ -344,10 +335,17 @@ func (cic *ClusterClient) getClusterInventoryObjsByName(inv Info) (object.Unstru
 	if apierrors.IsNotFound(err) {
 		return object.UnstructuredSet{}, nil
 	}
+	if inv.ID() != "" {
+		if inventoryID, err := retrieveInventoryLabel(clusterInv); err != nil {
+			return nil, err
+		} else if inv.ID() != inventoryID {
+			return nil, fmt.Errorf("inventory-id of inventory object in cluster doesn't match provided id %q", inv.ID())
```
karlkfi, line 342: error should probably include the name & namespace of the problematic inventory object, not just the id

cmd/status/cmdstatus.go
sdowell: This method is only being used by the `status` CLI command, so it doesn't need to be supported by all implementations of the interface. The CLI is statically built from this repo and only supports the ConfigMap type, so the implementation only needs to support ConfigMap.

I'm thinking of this method as an edge case that doesn't need to be implemented by all instances of the Client/Inventory interface. This will simplify the further refactoring of the inventory client interface and introductions of new types of clients

pkg/apply/applier.go
sdowell: This was moved to `getClusterInventoryObjsByName`. It's only needed for NameStrategy because LabelStrategy already filters by matching ID/label

pkg/inventory/inventory-client.go
sdowell: Done

cmd/status/cmdstatus.go
karlkfi: `kpt live status` actually uses the status.Runner built for `kapply status`. So it indirectly uses this code to build the `PrintData`, which is used both for printing and passed to the status poller/watcher.

My thinking is that this all needs refactoring to avoid exposing the inventory names and only expose a single inventory id.

If we do that, then we can push the listing down into the `Client.GetClusterObjs` or similar getter that returns the list of managed objects.

Additionally, we can replace `Client.GetClusterObjs` with something simpler like `Client.GetObjectIDs(id) ObjMetadataSet` that just returns the list of IDs from the inventory spec. Then a utility func can be used to get those objects from either the cluster or some other in-memory object cache. We don't need the inventory client to hide the whole cluster, just the inventory management.

karlkfi: Or maybe `inventory.Client.GetSpec(id) actuation.InventorySpec`.

I don't know if we need a `inventory.Client.GetStatus(id) actuation.InventoryStatus`, but that's the other logical getter.

We just don't want to `GetInventory(id) actuation.Inventory`, because that includes TypeMeta & ObjectMeta which requires the inventory to be a single object.

karlkfi: If you want to do that in a subsequent PR, LMK.

I think overall removing these methods from the Client interface is a good start. But we'll need to move this list code again later.

sdowell: > My thinking is that this all needs refactoring to avoid exposing the inventory names and only expose a single inventory id. If we do that, then we can push the listing down into the Client.GetClusterObjs or similar getter that returns the list of managed objects.

Note that this method returns _all inventories_ in the cluster, as opposed to _managed objects for a given inventory_ in the cluster. It seems a bit strange to me to have as part of the interface, as it appears to be specific to the use case of the status command (unused by the applier).

I didn't realize that kpt imports the status command to construct the `status.Runner` for ResourceGroup. That makes this a bit more sticky, so we might need to keep the listing method attached to the inventory client interface in that case. This means we'll need to decide how the listing interface should look for sharded inventories. I assume we will want each value in the key/value pair to represent a joined inventory, as opposed to having a separate key/value pair for each inventory shard.

karlkfi (approved):
/lgtm

PR #653, opened 2025-02-04
feat: rewrite inventory client with basic CRUD

sdowell:
This change rewrites the inventory client interface using simple CRUD methods. Prior to this change the inventory client consisted of several tightly coupled interfaces which made the client more complicated and less flexible.

The new client interface is simpler to implement and is intended to support implementations for sharded inventories, which may consist of multiple underlying KRM objects. The prior interfaces made assumptions of a single KRM object in several places, which made it incompatible with a sharded inventory.


ash2k:
/ok-to-test
/unassign

I'm sorry, I don't have time to review this atm.

sdowell:
> /ok-to-test /unassign @ash2k
> 
> I'm sorry, I don't have time to review this atm.

That's okay, I'm working with @karlkfi on this change.

pkg/apply/applier.go
```diff
@@ -96,8 +98,20 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.Info, objects objec
...
+		clusterInventory, err := a.invClient.Get(ctx, invInfo, inventory.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			clusterInventory = invInfo.InitialInventory()
+		} else if err != nil {
+			handleError(eventChannel, err)
+			return
+		}
+		if clusterInventory.ID() != invInfo.ID() {
+			handleError(eventChannel, fmt.Errorf("inventory-id of inventory object %s/%s in cluster doesn't match provided id %q",
+				invInfo.Namespace(), invInfo.Name(), invInfo.ID()))
```
karlkfi, line 110: should this return after handleError?

pkg/apply/common_test.go
```diff
@@ -64,10 +64,14 @@ func (i inventoryInfo) toUnstructured() *unstructured.Unstructured {
 	}
 }
 
-func (i inventoryInfo) toWrapped() inventory.Info {
+func (i inventoryInfo) toInfo() inventory.Info {
```
karlkfi, line 67: Why do we have both `inventoryInfo` and `inventory.Info`?

karlkfi: Ah, struct & interface.

Do we still need this to be an interface or could we just use a public struct?

pkg/apply/common_test.go
```diff
@@ -125,7 +129,7 @@ func newTestInventory(
 ) inventory.Client {
 	// Use an Client with a fakeInfoHelper to allow generating Info
 	// objects that use the FakeRESTClient as the UnstructuredClient.
-	invClient, err := inventory.ClusterClientFactory{StatusPolicy: inventory.StatusPolicyAll}.NewClient(tf)
```
karlkfi, line 128: StatusPolicyAll removed because it's the default?

pkg/apply/destroyer.go
```diff
@@ -79,10 +81,22 @@ func (d *Destroyer) Run(ctx context.Context, invInfo inventory.Info, options Des
...
+		clusterInventory, err := d.invClient.Get(ctx, invInfo, inventory.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			clusterInventory = invInfo.InitialInventory()
+		} else if err != nil {
+			handleError(eventChannel, err)
+			return
+		}
+		if clusterInventory.ID() != invInfo.ID() {
+			handleError(eventChannel, fmt.Errorf("inventory-id of inventory object %s/%s in cluster doesn't match provided id %q",
+				invInfo.Namespace(), invInfo.Name(), invInfo.ID()))
```
karlkfi, line 93: return after handleError?

pkg/apply/solver/solver_test.go
```diff
@@ -150,10 +150,9 @@ func TestTaskQueueBuilder_ApplyBuild(t *testing.T) {
 					Objects:   object.UnstructuredSet{},
 				},
 				&task.DeleteOrUpdateInvTask{
-					TaskName:      "inventory-set-0",
-					InvClient:     &inventory.FakeClient{},
-					InvInfo:       invInfo,
-					PrevInventory: object.ObjMetadataSet{},
```
karlkfi, line 156: Dont need the PrevInventory any more?

pkg/apply/task/inv_add_task.go
```diff
@@ -29,13 +29,14 @@ var (
...
 type InvAddTask struct {
-	TaskName      string
-	InvClient     inventory.Client
-	DynamicClient dynamic.Interface
-	Mapper        meta.RESTMapper
-	InvInfo       inventory.Info
-	Objects       object.UnstructuredSet
-	DryRun        common.DryRunStrategy
+	TaskName         string
+	ClusterInventory inventory.Inventory
```
karlkfi, line 33: ClusterInventory name could perhaps be more clear. Is this the inventory, untouched, as it was when we started? Or is it the cached one we update along the way?

"Inventory" might be sufficient.

pkg/apply/task/inv_add_task.go
```diff
@@ -71,7 +72,16 @@ func (i *InvAddTask) Start(taskContext *taskrunner.TaskContext) {
...
 		currentObjs := object.UnstructuredSetToObjMetadataSet(i.Objects)
-		_, err := i.InvClient.Merge(taskContext.Context(), i.InvInfo, currentObjs, i.DryRun)
+		inventoryObjs := i.ClusterInventory.Objects()
+		unionObjs := inventoryObjs.Union(currentObjs)
+
+		i.ClusterInventory.SetObjects(unionObjs)
+
+		var err error
+		if !i.DryRun.ClientOrServerDryRun() {
+			err = i.InvClient.CreateOrUpdate(taskContext.Context(), i.ClusterInventory, inventory.UpdateOptions{})
```
karlkfi, line 82: I think we talked about this, but what happens if the inventory has changed between the Get at the beginning of the Apply/Destroy and this Update call? Conflict, error, and exit?

How many other API calls are happening between these two calls?

Example cases we could test for: 
- a controller (ex: ResourceGroup controller) updates the status
- a controller updates the metadata
- a mutating webhook adds metadata to the inventory object

pkg/apply/task/inv_add_task.go
```diff
@@ -94,7 +104,7 @@ func inventoryNamespaceInSet(inv inventory.Info, objs object.UnstructuredSet) *u
 	for _, obj := range objs {
 		gvk := obj.GetObjectKind().GroupVersionKind()
 		if gvk == namespaceGVKv1 && obj.GetName() == invNamespace {
-			inventory.AddInventoryIDAnnotation(obj, inv)
+			inventory.AddInventoryIDAnnotation(obj, inv.ID())
```
karlkfi, line 107: huh. what is this call doing here? 

Seems weird of modify the object in a function called `inventoryNamespaceInSet`. That name makes it sound like it should return a bool.

Maybe it would be more clear if named `inventoryNamespaceFromSet`? But then, why are we modifying it in this function? Seems like something the caller should do and say why they're doing it.

pkg/apply/task/inv_add_task_test.go
```diff
@@ -167,8 +168,8 @@ func TestInvAddTask(t *testing.T) {
 			if result.Err != nil {
 				t.Errorf("unexpected error running InvAddTask: %s", result.Err)
 			}
-			actual, _ := client.GetClusterObjs(t.Context(), nil)
-			if !tc.expectedObjs.Equal(actual) {
+			actual, _ := client.Get(t.Context(), nil, inventory.GetOptions{})
```
karlkfi, line 171: Not super clear what this is doing.
Why ignore the error?
Why take a nil?

pkg/apply/task/inv_add_task.go
karlkfi: Or if we're not changing it after the Get, maybe `InitialInventory`?

pkg/apply/task/inv_set_task.go
```diff
@@ -152,24 +154,44 @@ func (i *DeleteOrUpdateInvTask) updateInventory(taskContext *taskrunner.TaskCont
...
 	klog.V(4).Infof("keep in inventory %d invalid objects", len(invalidObjects))
 	invObjs = invObjs.Union(invalidObjects)
 
 	klog.V(4).Infof("get the apply status for %d objects", len(invObjs))
-	objStatus := taskContext.InventoryManager().Inventory().Status.Objects
+	objStatus := taskContext.InventoryManager().Inventory().ObjStatuses
 
 	klog.V(4).Infof("set inventory %d total objects", len(invObjs))
-	err := i.InvClient.Replace(taskContext.Context(), i.InvInfo, invObjs, objStatus, i.DryRun)
+	// Skip entire function for dry-run.
```
karlkfi, line 165: Could this be done at the top of the function?

Or should we maybe compute the new inventory and log the diff when doing a dry-run?

pkg/apply/task/inv_set_task.go
```diff
@@ -152,24 +154,44 @@ func (i *DeleteOrUpdateInvTask) updateInventory(taskContext *taskrunner.TaskCont
...
+	if i.DryRun.ClientOrServerDryRun() {
+		klog.V(4).Infoln("dry-run replace inventory object: not applied")
+		return nil
+	}
+
+	i.ClusterInventory.SetObjects(invObjs)
+	if err := i.InvClient.CreateOrUpdate(taskContext.Context(), i.ClusterInventory, inventory.UpdateOptions{}); err != nil {
```
karlkfi, line 172: I think we might want to change this interface to take a method to mutate the object. That way it can handle RetryOnConflict.

pkg/apply/prune/prune_test.go
```diff
@@ -122,18 +122,11 @@ metadata:
 
 // Returns a inventory object with the inventory set from
 // the passed "children".
-func createInventoryInfo(children ...*unstructured.Unstructured) inventory.Info {
+func createInventoryInfo(children ...*unstructured.Unstructured) inventory.Inventory {
```
karlkfi, line 125: `createInventoryInfo` could be `newInventory`

pkg/apply/task/delete_inv_task_test.go
```diff
@@ -121,15 +121,17 @@ func TestDeleteInvTask(t *testing.T) {
 				if tc.err != result.Err {
 					t.Errorf("running DeleteOrUpdateInvTask expected error (%s), got (%s)", tc.err, result.Err)
 				}
-			} else {
-				if result.Err != nil {
-					t.Errorf("unexpected error running DeleteOrUpdateInvTask: %s", result.Err)
-				}
+				return
```
karlkfi, line 124: Do we not want to compare the expected objects in this case?

pkg/inventory/inventory-info.go
```diff
@@ -29,5 +29,7 @@ type Info interface {
 	// if the Id is necessary and how to use it for pruning objects.
 	ID() string
 
-	Strategy() Strategy
+	// InitialInventory returns an empty initialized inventory object.
```
karlkfi, line 32: I was never a big fan of this Info type. It seems confusing and overloaded.

In Config Sync, we tend to use a -Ref suffix for GKNN references, sometimes GVKNN. But I guess `inventory.Ref` is probably a worse name than `inventory.Info`, and `inventory.Reference` isn't much better. Maybe `inventory.Meta` or `inventory.Identifier` or `inventory.ID`? I guess ID would overlap with the string ID. So maybe Meta is better. Or we just leave Info, tho that's super generic.

Having the Info object also function as a factory is a bit surprising tho.

Would it make sense to have an `inventory.Factory` and separate `inventory.Meta`?

karlkfi: Do we even need the name & namespace? 
Can the inventory client handle converting from ID to NN?
Do the callers actually need to know the NN or can the inventory client just handle everything using the ID?

pkg/inventory/inventory_test.go
```diff
@@ -415,8 +415,3 @@ func TestLegacyInventoryName(t *testing.T) {
 func copyInventoryInfo() *unstructured.Unstructured {
 	return inventoryObj.DeepCopy()
 }
-
-func copyInventory() Info {
```
karlkfi, line 419: unused?

pkg/inventory/inventorycm.go
```diff
@@ -34,182 +27,76 @@ var ConfigMapGVK = schema.GroupVersionKind{
 // WrapInventoryObj takes a passed ConfigMap (as a resource.Info),
 // wraps it with the ConfigMap and upcasts the wrapper as
 // an the Inventory interface.
-func WrapInventoryObj(inv *unstructured.Unstructured) Storage {
-	return &ConfigMap{inv: inv}
+func WrapInventoryObj(inv *unstructured.Unstructured) Inventory {
+	internalInventory, err := configMapToInventory(inv)
+	if err != nil {
+		panic(err) //TODO: fix
```
karlkfi, line 33: Seems weird that a function called `WrapInventoryObj` could error. What's the case for this?

pkg/inventory/manager.go
```diff
@@ -14,28 +14,28 @@ import (
 
 // Manager wraps an Inventory with convenience methods that use ObjMetadata.
 type Manager struct {
-	inventory *actuation.Inventory
+	inventory BaseInventory
```
karlkfi, line 17: What's the story with this inventory field? Is it a cache? Why does the manager need to cache the inventory?

pkg/inventory/policy.go
```diff
@@ -141,11 +141,11 @@ func CanPrune(inv Info, obj *unstructured.Unstructured, policy Policy) (bool, er
 	}
 }
 
-func AddInventoryIDAnnotation(obj *unstructured.Unstructured, inv Info) {
+func AddInventoryIDAnnotation(obj *unstructured.Unstructured, inventoryID string) {
```
karlkfi, line 144: Do we want an `inventory.ID` type that's a string alias?

pkg/inventory/policy_test.go
```diff
@@ -32,6 +32,10 @@ func (i *fakeInventoryInfo) Strategy() Strategy {
 	return NameStrategy
 }
 
+func (i *fakeInventoryInfo) InitialInventory() Inventory {
```
karlkfi, line 35: maybe rename `NewInventory` or `NewEmptyInventory`

pkg/apply/task/inv_set_task.go
karlkfi: Hmm, will taking a mutate method work for a sharded client?
I guess it should, right?

pkg/inventory/inventorycm.go
sdowell: Sent https://github.com/kubernetes-sigs/cli-utils/pull/659

pkg/apply/common_test.go
karlkfi: Should we rename it ConfigMapInventoryClient?

pkg/apply/task/inv_set_task.go
karlkfi: Maybe just clarify the comment then "Exit before updating the inventory, but after logging the above changes"

pkg/apply/task/inv_add_task_test.go
karlkfi: Gotcha. 

nit: Can you just pass in something believable with a comment that it doesn't matter? It would just make it a bit easier to read.

pkg/inventory/inventory-info.go
sdowell: Did some refactoring to remove the need for Name, but we still need Namespace

pkg/apply/common_test.go
sdowell: https://github.com/kubernetes-sigs/cli-utils/pull/661

cmd/status/cmdstatus.go
```diff
@@ -179,7 +181,7 @@ func (r *Runner) loadInvFromDisk(cmd *cobra.Command, args []string) (*printer.Pr
 		// check if the object is under one of the targeted namespaces
 		if _, ok := r.namespaceSet[obj.Namespace]; ok || len(r.namespaceSet) == 0 {
 			// add to the map for future reference
-			printData.InvNameMap[obj] = inv.Name()
+			printData.InvNameMap[obj] = string(inv.ID())
```
karlkfi, line 184: Perhaps rename InvNameMap to InvIDMap

karlkfi: Maybe also change the value type to ID?

pkg/apply/applier_test.go
```diff
@@ -1976,26 +1976,23 @@ func TestReadAndPrepareObjects(t *testing.T) {
 	}{
 		"objects include inventory": {
 			invInfo: inventoryInfo{
-				name:      invInfo.Name(),
 				namespace: invInfo.Namespace(),
-				id:        invInfo.ID(),
+				id:        string(invInfo.ID()),
```
karlkfi, line 1980: Perhaps use ID as the field type?

pkg/apply/prune/prune.go
```diff
@@ -235,17 +235,13 @@ func (p *Pruner) removeInventoryAnnotation(ctx context.Context, obj *unstructure
...
 	opts Options,
 ) (object.UnstructuredSet, error) {
 	ids := object.UnstructuredSetToObjMetadataSet(objs)
-	invIDs, err := p.InvClient.GetClusterObjs(ctx, inv)
-	if err != nil {
-		return nil, err
-	}
 	// only return objects that were in the inventory but not in the object set
-	ids = invIDs.Diff(ids)
+	ids = inv.Objects().Diff(ids)
```
karlkfi, line 244: Occurs to me while reading this that the Objects method actually returns ObjectIDs/ObjectRefs/ObjectMetas, not the objects themselves. It might be slightly easier to read if we rename the getter/setter.

pkg/apply/task/inv_add_task.go
karlkfi: Alright, I see we use "Inventory" in the TaskQueueBuilder and "ClusterInventory" in the tasks. I don't think the name ClusterInventory helps clarify what gets stored in it or how to use it. It seems like it starts as what was on the cluster, but it could become stale and then we modify it. So we might as well just call it "Inventory" everywhere and be clear in the struct field comments.

pkg/apply/task/inv_add_task.go
karlkfi: Has this been solved? Not sure where to expect the retry. Here? Or in CreateOrUpdate?

pkg/inventory/inventory-client.go
```diff
@@ -24,455 +22,287 @@ import (
...
-		return map[string]object.ObjMetadataSet{}, nil
-	}
-
-	identifiers := make(map[string]object.ObjMetadataSet)
-
-	for i, inv := range clusterInvs.Items {
-		invName := inv.GetName()
-		identifiers[invName] = object.ObjMetadataSet{}
-		invObj, err := cic.InventoryFactoryFunc(&clusterInvs.Items[i])
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
karlkfi, line 264: Since we've used this wrong in the past, perhaps add a comment here to explain that this only makes up to 4 attempts with minimal delay, to avoid needing to retry the whole Apply/Destroy when there's a temporary conflict.

pkg/inventory/inventory-client.go
```diff
@@ -24,455 +22,287 @@ import (
...
-	}
-	if dryRun.ClientOrServerDryRun() {
-		klog.V(4).Infof("dry-run delete inventory object: not deleted")
+		// Update observedGeneration, if it exists
+		_, ok, err = unstructured.NestedInt64(uObj.Object, "status", "observedGeneration")
+		if err != nil {
+			return err
+		}
+		if ok {
+			err = unstructured.SetNestedField(uObj.Object, uObj.GetGeneration(), "status", "observedGeneration")
```
karlkfi, line 280: nit: since you're already calling NestedInt64 up there, you might as well just use the returned value, rather than making uObj.GetGeneration() do it again.

pkg/inventory/inventory-client.go
```diff
@@ -24,455 +22,287 @@ import (
...
+			if err != nil {
+				return err
+			}
+		}
+		newObj, err := cic.client.Namespace(uObj.GetNamespace()).UpdateStatus(ctx, uObj, metav1.UpdateOptions{})
```
karlkfi, line 285: In my changes to the kpt RG client, I added logging before each Create/Update/UpdateStatus. Do we want to do the same for the ConfigMap client?

pkg/inventory/inventorycm.go
```diff
@@ -34,182 +27,76 @@ var ConfigMapGVK = schema.GroupVersionKind{
...
-}
-
-func (icm *ConfigMap) ID() string {
-	// Empty string if not set.
-	return icm.inv.GetLabels()[common.InventoryLabel]
-}
-
-func (icm *ConfigMap) Strategy() Strategy {
-	return LabelStrategy
+func buildObjMap(objMetas object.ObjMetadataSet, objStatus []actuation.ObjectStatus) map[string]string {
```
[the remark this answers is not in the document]
karlkfi: ooh. It's building the ConfigMap `data` map, which contains the object IDs. 

Ok, so buildDataMap might be a clearer name.

pkg/inventory/inventorycm.go
```diff
@@ -34,182 +27,69 @@ var ConfigMapGVK = schema.GroupVersionKind{
...
-// GetObject returns the wrapped object (ConfigMap) as a resource.Info
-// or an error if one occurs.
-func (icm *ConfigMap) GetObject() (*unstructured.Unstructured, error) {
-	// Create the objMap of all the resources, and compute the hash.
-	objMap := buildObjMap(icm.objMetas, icm.objStatus)
-	// Create the inventory object by copying the template.
-	invCopy := icm.inv.DeepCopy()
+func inventoryToConfigMap(inv *UnstructuredInventory) (*unstructured.Unstructured, error) {
+	newConfigMap := inv.ClusterObj.DeepCopy()
+	objMap := buildObjMap(inv.Objects(), inv.ObjectStatuses())
```
karlkfi, line 85: objMap -> dataMap

pkg/apply/task/inv_add_task.go
sdowell: Yes, there is a retry in CreateOrUpdate.

pkg/inventory/manager.go
karlkfi, line 17: Why does this use the struct and not the interface?
If the struct is correct, should we use a pointer to avoid copying every time it's set?

pkg/inventory/policy.go
```diff
@@ -141,11 +141,11 @@ func CanPrune(inv Info, obj *unstructured.Unstructured, policy Policy) (bool, er
...
 }
 
-func AddInventoryIDAnnotation(obj *unstructured.Unstructured, inv Info) {
+func AddInventoryIDAnnotation(obj *unstructured.Unstructured, inventoryID ID) {
 	annotations := obj.GetAnnotations()
 	if annotations == nil {
 		annotations = make(map[string]string)
 	}
-	annotations[OwningInventoryKey] = inv.ID()
+	annotations[OwningInventoryKey] = string(inventoryID)
```
karlkfi, line 149: might be good to add a String() method to the ID, to avoid casting every time we use it, and to satisfy the Stringer interface for printing.

pkg/inventory/policy_test.go
```diff
@@ -24,12 +24,12 @@ func (i *fakeInventoryInfo) Namespace() string {
...
-func (i *fakeInventoryInfo) ID() string {
-	return i.id
+func (i *fakeInventoryInfo) ID() ID {
+	return ID(i.id)
 }
 
-func (i *fakeInventoryInfo) Strategy() Strategy {
-	return NameStrategy
+func (i *fakeInventoryInfo) NewEmptyInventory() Inventory {
+	return nil
```
karlkfi, line 32: make `panic("unimplemented")`

test/e2e/invconfig/configmap.go
```diff
@@ -37,18 +36,14 @@ func NewConfigMapTypeInvConfig(cfg *rest.Config) InventoryConfig {
 func newDefaultInvApplierFactory(cfg *rest.Config) applierFactoryFunc {
 	cfgPtrCopy := cfg
 	return func() *apply.Applier {
-		return newApplier(inventory.ClusterClientFactory{
-			StatusPolicy: inventory.StatusPolicyAll,
-		}, cfgPtrCopy)
+		return newApplier(inventory.ClusterClientFactory{}, inventory.StatusPolicyNone, cfgPtrCopy)
```
karlkfi, line 39: hmm. Does the ConfigMap inv client work with StatusPolicyAll? Obviously it doesn't need it, since there's no status... But would doing so make the client error when trying to update the status?

test/e2e/invconfig/invconfig.go
```diff
@@ -39,14 +36,7 @@ type InventoryConfig struct {
 }
 
 func CreateInventoryInfo(invConfig InventoryConfig, inventoryName, namespaceName, inventoryID string) inventory.Info {
-	switch invConfig.Strategy {
-	case inventory.NameStrategy:
```
karlkfi, line 43: Are these NameStrategy & LabelStrategy gone now?

How would we describe that in a change log?

pkg/apply/applier.go
```diff
@@ -96,8 +98,21 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.Info, objects objec
...
 
+		clusterInventory, err := a.invClient.Get(ctx, invInfo, inventory.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			clusterInventory = invInfo.NewEmptyInventory()
+		} else if err != nil {
+			handleError(eventChannel, err)
+			return
+		}
+		if clusterInventory.ID() != invInfo.ID() {
+			handleError(eventChannel, fmt.Errorf("expected inventory object to have inventory-id %q but got %q",
```
karlkfi, line 109: Should we include the inventory object name/namespace in the error message?

What about the whole inventory-id annotation name? Or is that specific to the inventory type?

Will this error make sense when using a sharded client?

pkg/inventory/inventory-client.go
```diff
@@ -24,455 +22,287 @@ import (
...
 
-	if err := wrappedInv.ApplyWithPrune(ctx, cic.dc, cic.mapper, cic.statusPolicy, objs); err != nil {
-		return fmt.Errorf("failed to write updated inventory to cluster: %w", err)
-	}
+var _ Inventory = &UnstructuredInventory{}
 
-	return nil
+// BaseInventory is a boilerplate struct that contains the basic methods
+// to implement Inventory. Can be extended for different inventory implementations.
+type BaseInventory struct {
```
karlkfi, line 121: I was confused several times when reading this name elsewhere in the code. So I've been looking for more descriptive names.

Would InMemoryInventory be accurate? Maybe that's kind of redundant.

Do we ever use it directly or is it only every wrapped by some other implementation like UnstructuredInventory?

pkg/inventory/inventory-client.go
sdowell: Done

pkg/inventory/inventory-client.go
sdowell: Done. Also note this isn't just a ConfigMap client, it can be used for any singleton unstructured object. For an example see the CR used by the e2e tests

pkg/inventory/inventorycm.go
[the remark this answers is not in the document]
sdowell: Done.

pkg/inventory/inventorycm.go
sdowell: Done

pkg/inventory/manager.go
sdowell: The field type could be either the `Inventory` interface or `BaseInventory`, however the instantiation by the `NewManager` constructor would still be `BaseInventory`. This is just used as a simple cache to track object statuses while the applier runs.

It's only set once by the constructor so I don't think it should make a material difference, but let me know what you prefer.

pkg/inventory/policy.go
sdowell: Done

pkg/inventory/policy_test.go
sdowell: Done

test/e2e/customprovider/provider.go
```diff
@@ -211,96 +129,60 @@ func (i InventoryCustomType) Store(objs object.ObjMetadataSet, status []actuatio
...
-	if invInfo == nil {
-		return nil, nil, fmt.Errorf("attempting to create a nil inventory object")
-	}
-
-	mapping, err := mapper.RESTMapping(invInfo.GroupVersionKind().GroupKind(), invInfo.GroupVersionKind().Version)
+func WrapInventoryInfoObj(obj *unstructured.Unstructured) inventory.Info {
+	inv, err := fromUnstructured(obj)
 	if err != nil {
-		return nil, nil, err
+		panic(err)
```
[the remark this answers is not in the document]
sdowell: This is implementing a [different function signature](https://github.com/kubernetes-sigs/cli-utils/blob/master/test/e2e/invconfig/invconfig.go#L22) which is only used by the e2e tests. It could be changed to return an error but it will result in a larger diff for this PR.

test/e2e/invconfig/configmap.go
sdowell: That is correct, the client would error if you tried to update the status while using ConfigMap inventory

test/e2e/invconfig/invconfig.go
sdowell: Instead of needing to implement both strategies for a single client, now the user would just implement/use a different inventory client implementation which uses labels instead of names. This is much more straightforward now as the client interface is greatly simplified. If desired we could surface another client implementation for selecting with labels, or reintroduce an option for the `UnstructuredClient` to support labels. My plan was to start with using `UnstructuredClient` using name/namespace and follow up adding a sharded inventory client which uses labels. I haven't decided yet whether to implement the sharded client as a generic client in cli-utils or if it will be a ResourceGroup-specific client implementation in Config Sync. I'll need to do some prototyping before arriving at that decision.

Also note the previous `LabelStrategy` implementation was incomplete and there were still assumptions in numerous places that the inventory was a singleton object (e.g. that it had a single `Name`). The new client makes no such assumptions and only surfaces a generic `ID`, and its up to the client implementation to decide how the object(s) get fetched/updated/deleted/etc.

cmd/status/cmdstatus.go
sdowell: Done

pkg/apply/prune/prune.go
sdowell: Done.

test/e2e/invconfig/invconfig.go
karlkfi: Can you copy this or rephase it in the PR description?

I've taken to making a list of changes in the style of release notes recently, and I think it's been valuable for both justifying changes and documenting them afterwards. I would suggest something similar for this PR.

test/e2e/invconfig/configmap.go
karlkfi: > if you tried to update the status

Does the ConfigMap Inventory error if you pass the StatusPolicyAll to the Applier?

pkg/inventory/manager.go
karlkfi: I think this using `BaseInventory` instead of `*BaseInventory` here is a subtle bug that may cause unnecessary extra copying of the inventory object. This may have been caught earlier if the field type was the `Inventory` interface, because the BaseInventory methods are all on the pointer.

Also, as a less important point, the Manager using the interface contributes twords proving that the interface is usable as-is and doesn't require using private BaseInventory fields or methods. It could also allow faking, but that's not really necessary in this case because the BaseInventory is so simple.

So I'd prefer using the interface as the struct field type.

cmd/status/cmdstatus.go
```diff
@@ -197,26 +199,31 @@ func (r *Runner) listInvFromCluster() (*printer.PrintData, error) {
...
-	for invName, identifiers := range identifiersMap {
+	identifiersMap := make(map[inventory.ID]object.ObjMetadataSet)
+	for _, inv := range inventories {
+		identifiersMap[inv.ID()] = inv.ObjectRefs()
+	}
+
+	for invID, identifiers := range identifiersMap {
 		// Check if there are targeted inventory names and include the current inventory name
-		if _, ok := r.inventoryNameSet[invName]; !ok && len(r.inventoryNameSet) != 0 {
+		if _, ok := r.inventoryNameSet[invID.String()]; !ok && len(r.inventoryNameSet) != 0 {
```
karlkfi, line 218: inventoryIDSet?

cmd/status/cmdstatus.go
```diff
@@ -197,26 +199,31 @@ func (r *Runner) listInvFromCluster() (*printer.PrintData, error) {
...
 			continue
 		}
 		// Filter objects
 		for _, obj := range identifiers {
 			// check if the object is under one of the targeted namespaces
 			if _, ok := r.namespaceSet[obj.Namespace]; ok || len(r.namespaceSet) == 0 {
 				// add to the map for future reference
-				printData.InvNameMap[obj] = invName
+				printData.InvNameMap[obj] = invID
```
karlkfi, line 226: InvIDMap?

pkg/apply/applier.go
```diff
@@ -96,8 +98,21 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.Info, objects objec
 		}
 		validator.Validate(objects)
 
+		clusterInventory, err := a.invClient.Get(ctx, invInfo, inventory.GetOptions{})
```
karlkfi, line 101: clusterInventory -> inv

pkg/apply/solver/solver_test.go
```diff
@@ -1494,11 +1385,11 @@ func TestTaskQueueBuilder_PruneBuild(t *testing.T) {
 				Pruner:    pruner,
 				Mapper:    mapper,
 				InvClient: fakeInvClient,
+				Inventory: inventoryObj,
 				Collector: vCollector,
 			}
 			taskContext := taskrunner.NewTaskContext(t.Context(), nil, nil)
-			tq := tqb.WithInventory(invInfo).
-				WithPruneObjects(tc.pruneObjs).
+			tq := tqb.WithPruneObjects(tc.pruneObjs).
```
karlkfi, line 1392: Doesn't look new, and you don't need to fix it here, but it's kinda weird that the TaskQueueBuilder takes some public fields and others have to be set with the With- methods...

pkg/apply/taskrunner/task_test.go
```diff
@@ -723,7 +713,7 @@ func TestWaitTask_Failed(t *testing.T) {
 		eventsFunc               func(*cache.ResourceCacheMap, *WaitTask, *TaskContext)
 		waitTimeout              time.Duration
 		expectedEvents           []event.Event
-		expectedInventory        *actuation.Inventory
+		expectedInventory        inventory.BaseInventory
```
karlkfi, line 716: should this also have a pointer?

pkg/inventory/inventory-client.go
```diff
@@ -24,455 +23,304 @@ import (
...
-	}
-	clusterInv, err := cic.getClusterInventoryInfo(ctx, localInv)
-	if err != nil {
-		return fmt.Errorf("failed to read inventory from cluster: %w", err)
-	}
+// UnstructuredInventory implements Inventory while also tracking the actual
+// KRM object from the cluster. This enables the client to update the
+// same object and utilize resourceVersion checks.
+type UnstructuredInventory struct {
+	BaseInventory
```
karlkfi, line 94: Should this have a pointer or does it need to be pointer-less in order for *UnstructuredInventory to satisfy the interface?

pkg/inventory/inventory-client.go
```diff
@@ -24,455 +23,304 @@ import (
...
+		return ""
 	}
+	// Empty string if not set.
+	return ID(ui.ClusterObj.GetLabels()[common.InventoryLabel])
+}
 
-	klog.V(4).Infof("replace cluster inventory: %s/%s", clusterInv.GetNamespace(), clusterInv.GetName())
-	klog.V(4).Infof("replace cluster inventory %d objects", len(objs))
+func (ui *UnstructuredInventory) NewEmptyInventory() Inventory {
+	return ui
```
karlkfi, line 123: This confuses me. But maybe I don't understand how NewEmptyInventory is being used. Why does it just return self here? That doesn't seem "empty" or "new".

pkg/inventory/inventory-client.go
```diff
@@ -24,455 +23,304 @@ import (
...
+// Get the in-cluster inventory
+func (cic *UnstructuredClient) Get(ctx context.Context, inv Info, _ GetOptions) (Inventory, error) {
+	ui, ok := inv.(*UnstructuredInventory)
+	if !ok {
+		return nil, fmt.Errorf("expected UnstructuredInventory")
 	}
-	if localInv == nil {
-		return nil, fmt.Errorf("retrieving cluster inventory object with nil local inventory")
+	if ui == nil {
+		return nil, fmt.Errorf("id is nil")
```
karlkfi, line 196: id -> inv ?

pkg/inventory/inventory-client.go
karlkfi: Ah. I misread the code.

Different question then: 
Why do we care if the status.observedGeneration already exists?
Is this to prevent updating the field on objects that doesn't have it in the scheme, like the ConfigMap?
If so, this seems like a sort of round-about way of doing that. I would sorta expect this to be avoided by checking the status policy, and requiring the status policy be None for ConfigMap inventories.

pkg/inventory/manager.go
karlkfi: Looks like if you try to use the interface, you wont be able to access ObjStatuses. So you'll need to call ObjectStatuses/SetObjectStatuses instead. That might require a little refactoring, but that's probably good, to exercise the interface to ensure it's usable.

test/e2e/customprovider/provider.go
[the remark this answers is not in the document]
karlkfi: doesn't need to block this PR. I'll rebase after it merges.

test/e2e/invconfig/configmap.go
sdowell: Yes, it would error when trying to update the status

pkg/inventory/inventory-client.go
sdowell: Done

pkg/inventory/manager.go
sdowell: After trying to do this refactor I remembered why I used `BaseInventory` when I originally wrote this. The `Inventory` interface requires implementation of the `Info` interface (Namespace/ID/etc), but this struct is just working with a raw objectRef/objectStatus cache and doesn't know about the metadata of the real inventory object.

pkg/inventory/inventory-client.go
karlkfi: Fair enough. This has the gotcha that the resource needs to have a default value in the CRD, otherwise it will never get updated. But that's probably fine for now.

Is it possible that this is a breaking behavioral change for the ConfigMap inventory? I don't know if we had any tests for ConfigMap + StatusPolicyAlll.

pkg/inventory/inventory-client.go
karlkfi: So it's not really a NewInventory (constructor) or ToInventory (conversion), but more of an AsInventory (casting/wrapping), as implemented.

Conceptually, I wouldn't expect Info to itself be an Inventory. Info is just a reference to an Inventory. So having the Info interface include an Inventory constructor method sorta made sense, but having it do conversion or casting to an Interface doesn't really make sense.

I think we need to disentangle the interfaces. What do you think about giving the Inventory interface an Info() getter method, rather than extending the Info interface?

cmd/status/cmdstatus.go
karlkfi: I guess the question is: Does the printer print them as IDs or Names?

If it prints names, how and where do we map ID to name?

If it prints IDs, then it should say ID throughout.

Are we breaking the printer by using the ID instead of the Name?

pkg/inventory/inventory-client.go
karlkfi: Looks like NewEmptyInventory is only used in Applier.Run and Destroyer.Run, to construct a new Inventory when the one specified by the Info is NotFound.

Should the Info just be an ID & Namespace?  Or should the Info also contain the specific type of the Inventory?

If we think of it like a GVKNN (or ObjMetadata or Config Sync core.ID), then it makes sense that it would also know the type of the inventory (aka Group & Kind), but being able to construct full inventory objects makes it more like a factory. In K8s you wouldn't use a GVKNN to construct an Object, you'd ask the Scheme to construct you an Object from the GVKNN or even just the GVK and then set the NN after construction.

So I think what Applier.Run and Destroyer.Run need is:
- an inventory.Info -  to reference the Inventory
- an inventory.Client - to CRUD the Inventory by Info
- an inventory.Factory - to construct a new Inventory, to store in memory until Client.CreateOrUpdate is called.

It seems reasonable for the Client impls to also implement the Factory methods. It might also be reasonable for the Client interface to just extend the Factory interface.

cmd/status/cmdstatus.go
```diff
@@ -164,11 +164,13 @@ func (r *Runner) loadInvFromDisk(cmd *cobra.Command, args []string) (*printer.Pr
 
 	// Based on the inventory template manifest we look up the inventory
 	// from the live state using the inventory client.
-	identifiers, err := invClient.GetClusterObjs(cmd.Context(), inv)
+	clusterInventory, err := invClient.Get(cmd.Context(), inv, inventory.GetOptions{})
```
karlkfi, line 167: inv -> invInfo
clusterInventory -> inv

sdowell: done

pkg/inventory/fake-inventory-client.go
```diff
@@ -96,14 +107,20 @@ func (fic *FakeClient) ClearError() {
 	fic.Err = nil
 }
 
-func (fic *FakeClient) GetClusterInventoryInfo(Info) (*unstructured.Unstructured, error) {
-	return nil, nil
+type FakeInventory struct {
+	BaseInventory
```
karlkfi, line 111: Can BaseInventory be used instead of FakeInventory?

pkg/inventory/inventory-client.go
```diff
@@ -6,473 +6,402 @@ package inventory
...
+	namespace   string
+	labels      map[string]string
+	annotations map[string]string
+}
 
-	clusterObjs, err := cic.GetClusterObjs(ctx, localInv)
-	if err != nil {
-		return fmt.Errorf("failed to read inventory objects from cluster: %w", err)
-	}
+func (i SimpleInfo) ID() ID {
```
karlkfi, line 72: Lets change these to GetID and GetNamespace, just to avoid needing to prefix the field names in the fakes. If we expose the fields on SimpleInfo & SingleObjectInfo we might not need the fakes at all.

pkg/inventory/inventory-client.go
```diff
@@ -6,473 +6,402 @@ package inventory
...
-	// Skip entire function for dry-run.
-	if dryRun.ClientOrServerDryRun() {
-		klog.V(4).Infoln("dry-run replace inventory object: not applied")
-		return nil
-	}
-	clusterInv, err := cic.getClusterInventoryInfo(ctx, localInv)
-	if err != nil {
-		return fmt.Errorf("failed to read inventory from cluster: %w", err)
-	}
+type SimpleInfo struct {
```
karlkfi, line 65: All the public interfaces, structs, methods, fields, funcs, and methods in this file need description comments.

pkg/inventory/inventory-client.go
```diff
@@ -6,473 +6,402 @@ package inventory
...
-	if err != nil {
-		return nil, err
-	}
-	groupResource := mapping.Resource.GroupResource().String()
-	namespace := localObj.Namespace
-	label, err := retrieveInventoryLabel(localInv)
-	if err != nil {
-		return nil, err
+	return NewSingleObjectInfo(ui.ID(), client.ObjectKeyFromObject(ui.ClusterObj),
+		maps.Clone(ui.ClusterObj.GetLabels()), maps.Clone(ui.ClusterObj.GetAnnotations()))
```
karlkfi, line 168: I'm torn between a few variants of this:
- use maps.Clone here defensively, but expensively
- pre-compute the Info at construction time & return a DeepCopy here
- pre-compute the Info at construction time & return the pre-computed Info here, requiring the caller to DeepCopy if they want to pass it somewhere else
- lazily compute the Info and cache it, then return a DeepCopy here
- lazily compute the Info and cache it, then return the lazily-computed Info here

pkg/inventory/inventory-client.go
```diff
@@ -6,473 +6,402 @@ package inventory
...
+
+func (ui *UnstructuredInventory) NewEmptyInventory() Inventory {
+	return ui
+}
+
+var _ Inventory = &UnstructuredInventory{}
+
+// BaseInventory is a boilerplate struct that contains the basic methods
+// to implement Inventory. Can be extended for different inventory implementations.
+type BaseInventory struct {
```
karlkfi, line 201: Looks like now that Inventory has an Info method, BaseInventory no longer implements the whole Inventory interface. That makes me wonder if AbstractInventory or InventoryContents would be a better name.

pkg/inventory/inventory-client.go
```diff
@@ -6,473 +6,402 @@ package inventory
...
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		create := false
+		obj, err := cic.client.Namespace(ui.Namespace()).Get(ctx, ui.Name(), metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			create = true
+		} else if err != nil {
+			return err
+		}
+		if obj != nil { // Obj is nil when IsNotFound, in this case keep initial/empty obj
+			ui.ClusterObj = obj
```
karlkfi, line 332: This will replace or erase the labels and annotations set on the Inventory by the Info...

I think we need to distinguish between the desired and current metadata.

pkg/inventory/inventory-client.go
```diff
@@ -6,473 +6,402 @@ package inventory
...
-}
-
-// getMapping returns the RESTMapping for the provided resource.
-func (cic *ClusterClient) getMapping(obj *unstructured.Unstructured) (*meta.RESTMapping, error) {
-	return cic.mapper.RESTMapping(obj.GroupVersionKind().GroupKind(), obj.GroupVersionKind().Version)
+	if cic.statusPolicy != StatusPolicyAll {
+		return nil
+	}
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		obj, err := cic.client.Namespace(ui.Namespace()).Get(ctx, ui.Name(), metav1.GetOptions{})
```
karlkfi, line 362: I don't know what the clean way to do this would be, but there's an opportunity to avoid a GET here, since the Create & Update calls above will have updated the uObj with the latest state from the cluster, including the status. So we don't need to GET again on the first attempt.

pkg/inventory/inventory-client.go
karlkfi: Deduping this with the other thread about inventory status.

pkg/inventory/inventory_test.go
```diff
@@ -414,10 +413,3 @@ func TestLegacyInventoryName(t *testing.T) {
 func copyInventoryInfo() *unstructured.Unstructured {
```
karlkfi, line 413: Named Info and returning Unstructured is weird.

Perhaps just remove this method and use `inventoryObj.DeepCopy()` as needed.

pkg/inventory/inventory-client.go
karlkfi: Nevermind. We solve this by removing labels and annotations from the Info.

test/e2e/invconfig/configmap.go
karlkfi: I think this is fine, as long as we document somewhere obvious that when using the ConfigMapInventory, you need to also use StatusPolicyNone.

cmd/status/cmdstatus.go
karlkfi: As you mentioned in chat, the kapply CLI only uses the single-object inventory, so it always prints the name. So it's ok if the fields in the CLI command all say Name. 

Perhaps this should cast the Info to the SingleObjectInfo and use the Name method then, instead of needing to explain how it gets the name from the ID?

pkg/inventory/fake-inventory-client.go
karlkfi: Nevermind. BaseInventory doesn't impl the Info method. So it can't be used as the fake.

pkg/inventory/inventory-client.go
karlkfi: I like InventoryContents. It makes sense as a partial impl without the metadata.

pkg/apply/common_test.go
```diff
@@ -38,7 +38,7 @@ import (
 type inventoryInfo struct {
 	name      string
 	namespace string
-	id        string
+	id        inventory.ID
 	set       object.ObjMetadataSet
```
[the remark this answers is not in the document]
sdowell: It was a glue struct to stitch together some of the test scaffolding. I've refactored the tests to not use this struct.

pkg/apply/applier_test.go
```diff
@@ -1431,7 +1431,7 @@ func TestApplier(t *testing.T) {
 			testCtx, testCancel := context.WithTimeout(context.Background(), testTimeout)
 			defer testCancel() // cleanup
 
-			invInfo, err := tc.invInfo.toWrapped()
+			invInfo, err := tc.invInfo.toInfo()
```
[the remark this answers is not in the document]
sdowell: Updated

cmd/status/cmdstatus.go
```diff
@@ -179,7 +181,7 @@ func (r *Runner) loadInvFromDisk(cmd *cobra.Command, args []string) (*printer.Pr
 		// check if the object is under one of the targeted namespaces
 		if _, ok := r.namespaceSet[obj.Namespace]; ok || len(r.namespaceSet) == 0 {
 			// add to the map for future reference
-			printData.InvNameMap[obj] = inv.Name()
+			printData.InvNameMap[obj] = invInfo.(*inventory.SingleObjectInfo).GetName()
```
karlkfi, line 184: nit: add a comment to explain why it's safe to cast without checking.

cmd/status/cmdstatus.go
```diff
@@ -201,11 +203,16 @@ func (r *Runner) listInvFromCluster() (*printer.PrintData, error) {
...
 
-	identifiersMap, err := invClient.ListClusterInventoryObjs(r.ctx)
+	inventories, err := invClient.List(r.ctx, inventory.ListOptions{})
 	if err != nil {
 		return nil, err
 	}
 
+	identifiersMap := make(map[string]object.ObjMetadataSet)
+	for _, inv := range inventories {
+		identifiersMap[inv.Info().(*inventory.SingleObjectInfo).GetName()] = inv.ObjectRefs()
```
karlkfi, line 213: nit: add a comment to explain why it's safe to cast

pkg/apply/solver/solver_test.go
karlkfi: looks like you fixed this. thanks!

pkg/inventory/inventory-client.go
```diff
@@ -24,455 +24,420 @@ import (
...
-	if err != nil {
-		return fmt.Errorf("failed to read inventory objects from cluster: %w", err)
-	}
+// Inventory is an internal representation of the inventory object that is used
+// to track which objects have been applied by the applier. The interface exists
+// so different underlying resource types can be used to implement the inventory.
+type Inventory interface {
+	Info() Info
+	// ObjectRefs returns the list of object references tracked in the inventory
+	ObjectRefs() object.ObjMetadataSet
```
karlkfi, line 122: Do we want to use "Get- " prefix on these methods too, so the field names don't need to abbreviate "Obj"?

pkg/inventory/inventory-client.go
```diff
@@ -24,455 +24,420 @@ import (
...
-
-	klog.V(4).Infof("deleting inventory object: %s/%s", obj.GetNamespace(), obj.GetName())
-	return cic.dc.Resource(mapping.Resource).Namespace(obj.GetNamespace()).
-		Delete(ctx, obj.GetName(), metav1.DeleteOptions{})
-}
-
-// getMapping returns the RESTMapping for the provided resource.
-func (cic *ClusterClient) getMapping(obj *unstructured.Unstructured) (*meta.RESTMapping, error) {
-	return cic.mapper.RESTMapping(obj.GroupVersionKind().GroupKind(), obj.GroupVersionKind().Version)
+	if cic.statusPolicy != StatusPolicyAll || opts.SkipStatusUpdate {
```
karlkfi, line 389: hmm. this is curious. do we need both the struct field and the method options?

pkg/inventory/inventory-client.go
```diff
@@ -24,455 +24,420 @@ import (
...
+			obj, err = cic.client.Namespace(ui.GetNamespace()).Get(ctx, ui.GetName(), metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+		}
+		uObj, err := cic.toUnstructured(obj.DeepCopy(), ui)
+		if err != nil {
+			return err
+		}
+		// Update observedGeneration, if it exists
```
karlkfi, line 407: Add a comment that says something like "If using a custom inventory resource, make sure the observedGeneration defaults to 0, so that it will be updated here, and so kstatus computes the status as InProgress after the object is created, until the status is updated here."

pkg/apply/applier.go
```diff
@@ -96,8 +97,22 @@ func (a *Applier) Run(ctx context.Context, invInfo inventory.Info, objects objec
...
 
+		inv, err := a.invClient.Get(ctx, invInfo, inventory.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			inv, err = a.invClient.NewInventory(invInfo)
+		}
+		if err != nil {
+			handleError(eventChannel, err)
+			return
+		}
+		if inv.Info().ID() != invInfo.ID() {
```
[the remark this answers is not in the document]
karlkfi: leave it in then for now.

karlkfi (commented):
This is looking so much better!

Just a few minor questions.

Also, please update the PR description with release-notes-like line items that say what changed, to help 3rd parties migrate to the latest version with all the library API changes.

pkg/inventory/inventory-client.go
sdowell: Done

pkg/inventory/inventory-client.go
sdowell: Done

pkg/inventory/inventory-client.go
sdowell: Yes, they serves a different purposes.
- statusPolicy is configured when the inventory client is instantiated and completely enables/disables status updates
- `opts.SkipStatusUpdate` is set by the applier during inv_add_task to omit the status update during that task, because the object statuses are stale during that step. The statuses are updated with `inv_set_task`

karlkfi: If InvAddTask doesn't update the status, it won't unblock the ResourceGroup controller. It needs to at least update the observedGeneration. That's how I fixed the race condition between controllers.

There's a bug we need to fix so that the inventory object status contains the correct status and conditions from kstatus, but we can fix that later. 

Even without that, the client still needs to update the inventory object status to set the strategy, actuation, and reconcile status. Setting the strategy declares the intent to apply or delete the object. Actuation & reconcile both get set to Pending because there's a new source that may not match the old source.

sdowell: > Even without that, the client still needs to update the inventory object status to set the strategy, actuation, and reconcile status. Setting the strategy declares the intent to apply or delete the object. Actuation & reconcile both get set to Pending because there's a new source that may not match the old source.

I was under the impression that the status would be stale during InvAddTask because InvAddTask doesn't update any of the object statuses. However I realized its buried in the old inventory client here, which sets the status for apply/prune objects: https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/inventory/inventory-client.go#L143. Given this realization, I think I'll push this logic up into the InvAddTask to update the object statuses before calling Update

karlkfi (approved):
/lgtm

giphy-so-fresh-so-clean

PR #654, opened 2025-02-04
test: fix test logging for inv annotation

sdowell:
The log line implies that it is expected for the objects to not have the inventory annotation, but the actual expectation is for the inventory annotation to exist.


karlkfi (approved):
/lgtm

PR #656, opened 2025-02-19
fix: pass context throughout applier

sdowell:
There were various spots in the applier that were not propagating the context and instead just using an empty context. This change fixes all such locations so that the parent context is always honored.

### DIFF
--- a/pkg/apply/applier_test.go
+++ b/pkg/apply/applier_test.go
@@ -1943,7 +1943,7 @@ func TestApplierCancel(t *testing.T) {
 
 func TestReadAndPrepareObjectsNilInv(t *testing.T) {
 	applier := Applier{}
-	_, _, err := applier.prepareObjects(nil, object.UnstructuredSet{}, ApplierOptions{})
+	_, _, err := applier.prepareObjects(t.Context(), nil, object.UnstructuredSet{}, ApplierOptions{})
 	assert.Error(t, err)
 }
 
@@ -2050,7 +2050,7 @@ func TestReadAndPrepareObjects(t *testing.T) {
 				watcher.BlindStatusWatcher{},
 			)
 
-			applyObjs, pruneObjs, err := applier.prepareObjects(tc.invInfo.toWrapped(), tc.resources, ApplierOptions{})
+			applyObjs, pruneObjs, err := applier.prepareObjects(t.Context(), tc.invInfo.toWrapped(), tc.resources, ApplierOptions{})
 			if tc.isError {
 				assert.Error(t, err)
 				return
--- a/pkg/apply/filter/inventory-policy-apply-filter.go
+++ b/pkg/apply/filter/inventory-policy-apply-filter.go
@@ -33,13 +33,13 @@ func (ipaf InventoryPolicyApplyFilter) Name() string {
 
 // Filter returns an inventory.PolicyPreventedActuationError if the object
 // apply should be skipped.
-func (ipaf InventoryPolicyApplyFilter) Filter(obj *unstructured.Unstructured) error {
+func (ipaf InventoryPolicyApplyFilter) Filter(ctx context.Context, obj *unstructured.Unstructured) error {
 	// optimization to avoid unnecessary API calls
 	if ipaf.InvPolicy == inventory.PolicyAdoptAll {
 		return nil
 	}
 	// Object must be retrieved from the cluster to get the inventory id.
-	clusterObj, err := ipaf.getObject(object.UnstructuredToObjMetadata(obj))
+	clusterObj, err := ipaf.getObject(ctx, object.UnstructuredToObjMetadata(obj))
 	if err != nil {
 		if apierrors.IsNotFound(err) {
 			// This simply means the object hasn't been created yet.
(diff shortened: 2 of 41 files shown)


pkg/apply/applier_test.go
```diff
@@ -1943,7 +1943,7 @@ func TestApplierCancel(t *testing.T) {
 
 func TestReadAndPrepareObjectsNilInv(t *testing.T) {
 	applier := Applier{}
-	_, _, err := applier.prepareObjects(nil, object.UnstructuredSet{}, ApplierOptions{})
+	_, _, err := applier.prepareObjects(context.TODO(), nil, object.UnstructuredSet{}, ApplierOptions{})
```
karlkfi, line 1946: t.Context()?

pkg/apply/filter/inventory-policy-apply-filter.go
```diff
@@ -20,6 +20,7 @@ import (
 // if an object should be applied based on the cluster object's inventory id,
 // the id for the inventory object, and the inventory policy.
 type InventoryPolicyApplyFilter struct {
+	Context   context.Context
```
karlkfi, line 23: Suggest adding context to the Filter method signature in the interface.

pkg/apply/prune/prune.go
```diff
@@ -137,7 +137,7 @@ func (p *Pruner) Prune(
 				if errors.As(filterErr, &abandonErr) {
 					if !opts.DryRunStrategy.ClientOrServerDryRun() {
 						var err error
-						obj, err = p.removeInventoryAnnotation(obj)
+						obj, err = p.removeInventoryAnnotation(taskContext.Context(), obj)
 						if err != nil {
```
karlkfi, line 141: Maybe move the taskContext to the first method arg

pkg/apply/applier_test.go
sdowell: done

pkg/apply/filter/inventory-policy-apply-filter.go
sdowell: done

pkg/apply/prune/prune.go
sdowell: done

pkg/apply/task/inv_set_task.go
```diff
@@ -49,7 +51,7 @@ func (i *DeleteOrUpdateInvTask) Start(taskContext *taskrunner.TaskContext) {
 	go func() {
 		var err error
 		if i.Destroy && i.destroySuccessful(taskContext) {
-			err = i.deleteInventory()
+			err = i.deleteInventory(taskContext.Context())
```
karlkfi, line 54: hmm. should we make deleteInventory take a taskContext to be consistent with updateInventory?

pkg/inventory/inventory-client.go
```diff
@@ -397,35 +397,9 @@ func (cic *ClusterClient) ListClusterInventoryObjs(ctx context.Context) (map[str
 	return identifiers, nil
 }
 
-// createInventoryObj creates the passed inventory object on the APIServer.
-func (cic *ClusterClient) createInventoryObj(obj *unstructured.Unstructured, dryRun common.DryRunStrategy) (*unstructured.Unstructured, error) {
```
karlkfi, line 401: is this unused or moved?

karlkfi: Since this is such a big PR, maybe move this change out by itself.

pkg/apply/task/inv_set_task.go
karlkfi: This was the only one I saw, but please double check that all the other interfaces have consistent use of either the TaskContext or the Context.

pkg/apply/applier_test.go
karlkfi: Looks like the test environment doesn't have a new enough Go version to compile Testing.Context. Can you update that?

pkg/apply/task/inv_set_task.go
sdowell: updateInventory actually uses other fields from the taskContext, which is why it needs it. deleteInventory just needs the `context.Context`. It's such a tiny function it doesn't seem like it matters to me.

pkg/inventory/inventory-client.go
sdowell: It was unused.

sdowell: https://github.com/kubernetes-sigs/cli-utils/pull/657

pkg/apply/applier_test.go
sdowell: https://github.com/kubernetes/test-infra/pull/34378

pkg/apply/task/inv_set_task.go
karlkfi: You know I love consistency, but I agree it's a small enough that it's not a blocker.

pkg/inventory/storage.go
```diff
@@ -40,12 +41,12 @@ type Storage interface {
 	GetObject() (*unstructured.Unstructured, error)
 	// Apply applies the inventory object. This utility function is used
 	// in InventoryClient.Merge and merges the metadata, spec and status.
-	Apply(dynamic.Interface, meta.RESTMapper, StatusPolicy) error
+	Apply(context.Context, dynamic.Interface, meta.RESTMapper, StatusPolicy) error
```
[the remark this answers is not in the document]
sdowell: I'm not planning to upgrade cli-utils in Config Sync until the next change goes in (https://github.com/kubernetes-sigs/cli-utils/pull/653), which completely changes these interfaces.

My intent here was to fix the context issue as an incremental change before the complete interface rewrite. Alternatively we could fix the context issue as part of the interface rewrite or as a subsequent change. I'm fine either way.

If we submit this change and then for some reason need to immediately bump the cli-utils version, we could always update the kpt client implementation to pass a context through.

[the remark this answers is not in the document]
karlkfi: Alright.

I'm gonna extract my third_party fork to its own PR anyway. So that will unblock using these as soon as monday.

karlkfi (approved):
/lgtm

PR #659, opened 2025-03-11
refactor: return error from ConfigMap wrap functions

sdowell:
This change is primarily motivated by a subsequent change which replaces the lazy conversion with an immediate conversion, which requires for an error to be returned. Since this touches so many pieces of code this is broken out as a separate change.

This also fixes several places where errors were not being propagated.

### DIFF
--- a/cmd/apply/cmdapply.go
+++ b/cmd/apply/cmdapply.go
@@ -129,7 +129,10 @@ func (r *Runner) RunE(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	inv := inventory.WrapInventoryInfoObj(invObj)
+	inv, err := inventory.ConfigMapToInventoryInfo(invObj)
+	if err != nil {
+		return err
+	}
 
 	invClient, err := r.invFactory.NewClient(r.factory)
 	if err != nil {
--- a/pkg/apply/applier_test.go
+++ b/pkg/apply/applier_test.go
@@ -1431,7 +1431,9 @@ func TestApplier(t *testing.T) {
 			testCtx, testCancel := context.WithTimeout(context.Background(), testTimeout)
 			defer testCancel() // cleanup
 
-			eventChannel := applier.Run(runCtx, tc.invInfo.toWrapped(), tc.resources, tc.options)
+			invInfo, err := tc.invInfo.toWrapped()
+			require.NoError(t, err)
+			eventChannel := applier.Run(runCtx, invInfo, tc.resources, tc.options)
 
 			// only start sending events once
 			var once sync.Once
@@ -1872,7 +1874,9 @@ func TestApplierCancel(t *testing.T) {
 			testCtx, testCancel := context.WithTimeout(context.Background(), tc.testTimeout)
 			defer testCancel() // cleanup
 
-			eventChannel := applier.Run(runCtx, tc.invInfo.toWrapped(), tc.resources, tc.options)
+			invInfo, err := tc.invInfo.toWrapped()
+			require.NoError(t, err)
+			eventChannel := applier.Run(runCtx, invInfo, tc.resources, tc.options)
 
 			// only start sending events once
 			var once sync.Once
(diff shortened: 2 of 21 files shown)


pkg/apply/common_test.go
```diff
@@ -64,8 +64,8 @@ func (i inventoryInfo) toUnstructured() *unstructured.Unstructured {
 	}
 }
 
-func (i inventoryInfo) toWrapped() inventory.Info {
-	return inventory.WrapInventoryInfoObj(i.toUnstructured())
+func (i inventoryInfo) toWrapped() (inventory.Info, error) {
```
karlkfi, line 67: I don't really understand what this is doing.
Why does Info wrap an Unstructured?
Why does inventoryInfo contain anything other than metadata?
Why would this fail?

sdowell: I haven't fully dug into this portion of the test code, but this struct is used to set up the fake dynamic client for these tests: https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/apply/common_test.go#L292

I think this is supposed to be some glue between the dynamic client and the inventory object

cmd/apply/cmdapply.go
```diff
@@ -129,7 +129,10 @@ func (r *Runner) RunE(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	inv := inventory.WrapInventoryInfoObj(invObj)
+	inv, err := inventory.ConfigMapToInventoryInfoObj(invObj)
```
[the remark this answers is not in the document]
karlkfi: nit: prefer `ConfigMapToInventoryInfo`, since we don't refer to the Info as an Object elsewhere, and it's not a runtime.Object or client.Object.

pkg/apply/common_test.go
karlkfi: Is this going away in your rewrite?

cmd/apply/cmdapply.go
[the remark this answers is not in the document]
sdowell: Done

karlkfi (approved):
/lgtm

pkg/apply/applier_test.go
```diff
@@ -1949,7 +1953,8 @@ func TestReadAndPrepareObjectsNilInv(t *testing.T) {
 
 func TestReadAndPrepareObjects(t *testing.T) {
 	inventoryObj := testutil.Unstructured(t, resources["inventory"])
-	inventory := inventory.WrapInventoryInfoObj(inventoryObj)
+	invInfo, err := inventory.ConfigMapToInventoryInfo(inventoryObj)
+	assert.NoError(t, err)
```
karlkfi, line 1957: require maybe? prolly want to skip the rest of the tests if it errors

pkg/apply/destroyer_test.go
```diff
@@ -313,12 +313,15 @@ func TestDestroyerCancel(t *testing.T) {
 		t.Run(tn, func(t *testing.T) {
 			statusWatcher := newFakeWatcher(tc.statusEvents)
 
-			invInfo := tc.invInfo.toWrapped()
+			invInfo, err := tc.invInfo.toWrapped()
+			assert.NoError(t, err)
```
karlkfi, line 317: require

pkg/apply/applier_test.go
```diff
@@ -1431,7 +1431,9 @@ func TestApplier(t *testing.T) {
 			testCtx, testCancel := context.WithTimeout(context.Background(), testTimeout)
 			defer testCancel() // cleanup
 
-			eventChannel := applier.Run(runCtx, tc.invInfo.toWrapped(), tc.resources, tc.options)
+			invInfo, err := tc.invInfo.toWrapped()
+			assert.NoError(t, err)
```
karlkfi, line 1435: require

pkg/apply/applier_test.go
```diff
@@ -1872,7 +1874,9 @@ func TestApplierCancel(t *testing.T) {
 			testCtx, testCancel := context.WithTimeout(context.Background(), tc.testTimeout)
 			defer testCancel() // cleanup
 
-			eventChannel := applier.Run(runCtx, tc.invInfo.toWrapped(), tc.resources, tc.options)
+			invInfo, err := tc.invInfo.toWrapped()
+			assert.NoError(t, err)
```
karlkfi, line 1878: require

pkg/apply/applier_test.go
sdowell: Done

pkg/apply/destroyer_test.go
sdowell: done

pkg/apply/applier_test.go
sdowell: done

karlkfi (approved):
/lgtm

PR #660, opened 2025-03-13
test: fix inventory client tests

sdowell:
These tests were not actually verifying anything due to early exits and faulty conditions

### DIFF
--- a/pkg/inventory/inventory-client_test.go
+++ b/pkg/inventory/inventory-client_test.go
@@ -90,29 +90,23 @@ func TestGetClusterInventoryInfo(t *testing.T) {
 				WrapInventoryObj, InvInfoToConfigMap, tc.statusPolicy, ConfigMapGVK)
 			require.NoError(t, err)
 
-			var inv *unstructured.Unstructured
 			if tc.inv != nil {
-				inv = storeObjsInInventory(tc.inv, tc.localObjs, tc.objStatus)
+				wrapped := storeObjsInInventory(tc.inv, tc.localObjs, tc.objStatus)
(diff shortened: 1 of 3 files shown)


pkg/inventory/inventory-client_test.go
```diff
@@ -104,15 +105,13 @@ func TestGetClusterInventoryInfo(t *testing.T) {
 			if !tc.isError && err != nil {
```
karlkfi, line 105: This can just be if yes require.Error, else require.NoError

pkg/inventory/inventory-client_test.go
```diff
@@ -104,15 +105,13 @@ func TestGetClusterInventoryInfo(t *testing.T) {
...
-				clusterObjs, err := wrapped.Load()
-				if err != nil {
-					t.Fatalf("unexpected error received: %s", err)
-				}
-				if !tc.localObjs.Equal(clusterObjs) {
-					t.Fatalf("expected cluster objs (%v), got (%v)", tc.localObjs, clusterObjs)
-				}
+			require.NotNil(t, clusterInv)
+			wrapped := WrapInventoryObj(clusterInv)
+			require.NoError(t, err)
```
karlkfi, line 110: Don't need this here since the previous line doesn't produce an error

karlkfi (changes requested):

pkg/inventory/inventory-client_test.go
sdowell: done

pkg/inventory/inventory-client_test.go
sdowell: Ah missed that because the next PR has it return an error. Removed

PR #661, opened 2025-03-14
refactor: rename ClusterClientFactory to ConfigMapClientFactory

sdowell:
The previous name was somewhat misleading, as this factory is specifically for creating a ConfigMap client.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -63,7 +63,7 @@ func main() {
 	}
 
 	loader := manifestreader.NewManifestLoader(f)
-	invFactory := inventory.ClusterClientFactory{StatusPolicy: inventory.StatusPolicyNone}
+	invFactory := inventory.ConfigMapClientFactory{StatusPolicy: inventory.StatusPolicyNone}
 
 	names := []string{"init", "apply", "destroy", "diff", "preview", "status"}
 	subCmds := []*cobra.Command{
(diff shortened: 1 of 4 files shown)


cmd/main.go
```diff
@@ -63,7 +63,7 @@ func main() {
 	}
 
 	loader := manifestreader.NewManifestLoader(f)
-	invFactory := inventory.ClusterClientFactory{StatusPolicy: inventory.StatusPolicyNone}
```
soltysh, line 66: As I'm looking at the [client](https://github.com/kubernetes-sigs/cli-utils/blob/b6f5dbece92f7053344b91b99b62523176e5a734/pkg/inventory/inventory-client.go#L61-L89) it seems pretty versatile, allowing to work with all the resources (especially that it [embeds dynamic client](https://github.com/kubernetes-sigs/cli-utils/blob/b6f5dbece92f7053344b91b99b62523176e5a734/pkg/inventory/inventory-client.go#L67)). So imo, its current name is pretty accurate.

pkg/inventory/inventory-client-factory.go
```diff
@@ -8,19 +8,20 @@ import (
 )
 
 var (
-	_ ClientFactory = ClusterClientFactory{}
```
karlkfi, line 11: Wanna change the file name too? configmap-client-factory.go perhaps?

sdowell: done

karlkfi (approved):
/lgtm

PR #664, opened 2025-04-03
perf: Reduce inventory update request size

karlkfi:
## TL;DR
- Reduce inventory update request sizes by adding another `ToUnstructuredFunc` to the `inventory.UnstructuredClient`, for inventory resources that have a status subresource.
- Update `ConfigMapClientFactory` and `CustomClientFactory` to work with the modified `UnstructuredClient`
- Remove the `StatusPolicy`, which the `inventory.UnstructuredClient` no longer needs to know about.
- Fix `ConfigMapClientFactory` to include or exclude the status, as configured.
- Backfill `UnstructuredClient.CreateOrUpdate` tests to include status validation

## Motivation
The UnstructuredClient currently uses the same ToUnstructuredFunc to
generate the spec and the status. This is inefficient when status is
a subresource, because the ToUnstructuredFunc implementation always
needs to return both the spec and the status, which means they both
get sent over the wire to the Kubernetes API. The Kubernetes API will
ignore the spec when using the status endpoint, and ignore the status
when using the normal resource endpoint, so doing all the extra work
to generate, copy, and format the extra fields is inefficient on the
client side. In some edge cases this can also cause errors on the
server side, when the spec and status are larger than the maximum
API request size limit but smaller than the etcd object storage size
limit.

## Changes
To work around this issue, a new toStatus ToUnstructuredFunc can be
injected into the UnstructuredClient. When non-nil, this new func will
be used before calling UpdateStatus.

This has the added benefit that it makes it possible to use the
UnstructuredClient on custom inventory resources that have a status
field but not a status subresource.

To validate this change, additional test cases were added to
TestCreateOrUpdate, which exposed a bug in inventoryToConfigMap:
The object statuses were not being parsed. In order to fix this bug,
a new parseDataMap method was added.

With the UnstructuredClient having two ToUnstructuredFunc varients,
the UnstructuredClient no longer needs to know the StatusPolicy, so
StatusPolicy was removed. This clears up ambiguity about what
StatusPolicy meant when the status field was not a subresource. It also
means that only the ToUnstructuredFunc implementer needs to know whether
the status is enabled or disabled. Because of this, the
ConfigMapClientFactory was refactored to use a StatusEnabled bool,
instead of the StatusPolicy.

## Dependencies
- https://github.com/kubernetes-sigs/cli-utils/pull/666
- https://github.com/kubernetes-sigs/cli-utils/pull/667


pkg/inventory/inventory-client.go
```diff
@@ -363,6 +366,13 @@ func (cic *UnstructuredClient) CreateOrUpdate(ctx context.Context, inv Inventory
 		if err != nil {
 			return err
 		}
+		// When status is a sub-resource, remove the status from the spec update
+		// request. The status is ignored by the server anyway, so this reduces
+		// the size of the request, which allows spec updates to succeed when
+		// the status makes the object too big to apply.
+		if _, found := uObj.Object["status"]; found && cic.statusSubresource {
```
sdowell, line 373: The `cic.statusSubresource` check seems unnecessary, you could just always delete the `status` field if it exists (similar to the `spec` removal below)

pkg/inventory/inventory-client.go
```diff
@@ -383,7 +393,7 @@ func (cic *UnstructuredClient) CreateOrUpdate(ctx context.Context, inv Inventory
 	if err != nil {
 		return err
 	}
-	if cic.statusPolicy != StatusPolicyAll {
+	if !cic.statusSubresource || cic.statusPolicy != StatusPolicyAll {
```
sdowell, line 396: Why do we need `statusSubresource` and `statusPolicy`? In this implementation they seem redundant

pkg/inventory/inventory-client.go
karlkfi: This check safeguards against the case where the status needs to be sent in the Create or Update call, because it's not a subresource. 

This case does not affect the ConfigMap, because it has no status field. Nor does it affect the custom Inventory resource in the e2e tests, because it specifies status as a subresource. 

The case this _does_ affect is a custom inventory resource that _does not_ specify status as a subresource, which we don't have a test for today. In that case, if we remove the status, it will never be persisted, because UpdateStatus errors 404 when there's no /status endpoint on the resource API. That's why I also added a check of `!cic.statusSubresource` to exit before the UpdateStatus section.

sdowell: Ack. I wasn't aware it was possible to declare a top level `status` field but not implement it as a subresource

karlkfi: Subresources were added to k8s as a generic solution but the initial usage was specifically for status fields, to allow different controllers to manage spec vs status. 

These days, it's best practice for CRDs to always make the status a subresource, but it's not a requirement. Plus, not all built-in resources with a status implement status as a subresource. To make matters more confusing, some of the built-in resources, like `Endpoint` has a status but not a status subresource. This actually caused performance problems, but couldnt be fixed for reverse compatibility reasons, which required the introduction of `EndpointSlice`.

In a CRD, the subresources are defined with:
```
spec:
  versions:
  - name: v1alpha1
    ...
    subresources:
      status: {}
```

pkg/inventory/inventory-client.go
sdowell: In the current `UnstructuredClient`, `StatusPolicyAll` basically just means to call `UpdateStatus` to update the `status` subresource. That's the only thing the status policy is used for.

If the status is stored somewhere other than the `status` subresource, the `UnstructuredClient` doesn't know about it and it's an implementation detail of the conversion function

pkg/inventory/inventory-client.go
karlkfi: Actually, that's sort of revisionist of me. I think technically subresources were first added for the `/scale` endpoint, to make HPA perform better in K8s v1.8 (https://ronaknathani.com/blog/2024/01/understanding-the-kubernetes-scale-subresource/). 

Scale was sort of a hack to work around a design flaw in the ReplicaSet/ReplicationController API that embedded the scale in the spec. When they added HPA, they had two controllers managing the spec, which caused coordination and performance problems. IMO, they should have extracted the scale to its own resource with its own spec and status, but they added it as a subresource instead, because the best practice of using status.observedCeneration for controller coordination wasn't as well known at the time (not sure if it even existed yet when the ReplicationController was designed).

The status subresource was added later, in v1.10, specifically for CRDs, which more often needed separate controllers managing spec and status.

pkg/inventory/inventory-client.go
karlkfi: Right. This change would allow it to accommodate the extra use case. I realized we hadn't considered it before, and it seems easy to add.

I could add a test, to prevent regressions.

sdowell: I guess what I'm saying is we don't need both of `statusPolicy` and `statusSubresource` since they both mean the same thing for the `UnstructuredClient`. They both mean that `status` is a subresource and that `UpdateStatus` needs to be called after `Update`

karlkfi: `cic.toUnstructured` is called twice, once before updating the spec and once before updating the status.

How would the `ToUnstructuredFunc` implementation return once without status and once with status?

karlkfi: We don't need both to support only 2 use cases, sure.
We do need both to support 3 use cases.

1. no status
2. status as subresource
3. status without subresource

karlkfi: If you don't want to add the boolean, we could add more StatusPolicy enum options, like:

1. StatusPolicyNone
2. StatusPolicyEmbedded
3. StatusPolicySubresource
4. StatusPolicyField

karlkfi: Alternatively, we could pass two `ToUnstructuredFunc` impls to `NewUnstructuredClient`, one that includes the status and one that doesn't. That would make implementation a little more complicated but would allow an additional optimization so that the status doesn't get computed, populated, and copied when it's not needed.

sdowell: Why does it need more enums? From the perspective of `UnstructuredClient` the only thing it needs to know is whether the status is stored in a status subresource, and if so it calls `UpdateStatus` (and does the new `status`/`spec` removal logic added by this PR).

What would the `UnstructuredClient` do with those other enum values?

pkg/inventory/inventory-client.go
karlkfi: This thread is obsolete. PR was significantly refactored to remove the `statusSubresource` and add a `toUnstructuredStatus`.

pkg/inventory/inventory-client.go
karlkfi: This thread is obsolete. PR was significantly refactored to remove the `statusSubresource` and `StatusPolicy` and add a `toUnstructuredStatus` instead.

PR #665, opened 2025-04-04
refactor: rename UnstructuredInv to SingleObjectInv

sdowell:
After several iterations/refactors of the original change, the meaning of the UnstructuredInventory struct has changed. The name SingleObjectInventory is a more accurate name to describe the current struct.


karlkfi (approved):
/lgtm

PR #667, opened 2025-04-09
refactor: strategy, actuation, reconcile to string

karlkfi:
- Convert the ActuationStrategy, ActuationStatus, and ReconcileStatus to string enums, from int enums. This makes parsing, formatting, and debugging easier, especially to/from json and yaml, avoiding the need to implement MarshalJSON and UnmarshalJSON.
- Add Parse methods for parsing
- Add String methods for formatting
- Remove the Stringer generated String methods, to avoid circular dependency with the Parse methods.


seans3 (approved):
/lgtm

PR #668, opened 2025-05-10
kstatus update for suspended jobs

smiller333:
Suspended jobs are returning as in-progress which makes it seem like the job is running, but it should be considered as current since suspended jobs will not run.

### DIFF
--- a/pkg/kstatus/status/util.go
+++ b/pkg/kstatus/status/util.go
@@ -141,3 +141,21 @@ func GetIntField(obj map[string]interface{}, fieldPath string, defaultValue int)
 	}
 	return defaultValue
 }
+
+func GetBoolField(obj map[string]interface{}, fieldPath string, defaultValue bool) bool {
+	fields := strings.Split(fieldPath, ".")
+	if fields[0] == "" {
+		fields = fields[1:]
+	}
+
+	val, found, err := apiunstructured.NestedFieldNoCopy(obj, fields...)
+	if !found || err != nil {
+		return defaultValue
+	}
+
+	switch v := val.(type) {
+	case bool:
+		return v
+	}
+	return defaultValue
+}
--- a/pkg/kstatus/status/util.go
+++ b/pkg/kstatus/status/util.go
@@ -119,8 +119,10 @@ func GetStringField(obj map[string]interface{}, fieldPath string, defaultValue s
 	return rv
 }
 
-// GetIntField return field as string defaulting to value if not found
+// GetIntField return field as int defaulting to value if not found
 func GetIntField(obj map[string]interface{}, fieldPath string, defaultValue int) int {
+	var rv = defaultValue
+
 	fields := strings.Split(fieldPath, ".")
 	if fields[0] == "" {
 		fields = fields[1:]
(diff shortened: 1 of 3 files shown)


karlkfi:
Can you explain why suspended should be considered Current rather than Failed?

Seems kind of indeterminate. But I could understand wanting to continue if you applied it suspended to begin with.

karlkfi:
Looks like there's a spec field to suspend a job. So it might make sense to mark as Current only if both the spec field and condition match. Otherwise I would probably consider it InProgress, as it doesn't match the desired state yet.

smiller333:
There is an issue with Rancher Fleet where the deployments are not marked as completed when there is a suspended job in the deployment files.  This is because Rancher Fleet is using cli-utils to check the status of the resources it deploys in the cluster to ensure the deployments were successful and in the correct state.  Since a suspended job shows as in-progress by default, the Fleet deployment is never satisfied.

I thought about doing an MR for Fleet, but this still feels like the suspended status on jobs is not properly handled in the cli-utils.  If the job file is applied to the cluster and suspended, then the status of the deployment is that it is current, not in-progress.  Unless I am mistaken the in-progress status should mean that the resource is in some kind of a transition state, but a suspended job is never transitioning.

smiller333:
I only added the check for the status field as that was the existing pattern on the other parts of the function.  Is it possible for a job to be suspend: false in the spec, but have a suspended state in the conditions?

smiller333:
I updated the code to also check the `spec.suspend` field before treating a job with the Suspended status as current to ensure that there is no ambiguity.

I also did a quick look through the [kubernetes documentation](https://kubernetes.io/docs/concepts/workloads/controllers/job/#suspending-a-job) and based on this paragraph, only the Status field needs to be checked to determine if a job is suspended or not.

> The Job condition of type "Suspended" with status "True" means the Job is suspended; the lastTransitionTime field can be used to determine how long the Job has been suspended for. If the status of that condition is "False", then the Job was previously suspended and is now running. If such a condition does not exist in the Job's status, the Job has never been stopped.

smiller333:
@karlkfi 
I haven't heard back on my comments and commits, do you need any other information or commits?  I can move my explanation comments to the main MR comment for clarity if needed.  Just let me know.

pkg/kstatus/status/util.go
```diff
@@ -119,16 +119,18 @@ func GetStringField(obj map[string]interface{}, fieldPath string, defaultValue s
 	return rv
 }
 
-// GetIntField return field as string defaulting to value if not found
+// GetIntField return field as int defaulting to value if not found
 func GetIntField(obj map[string]interface{}, fieldPath string, defaultValue int) int {
+	var rv = defaultValue
```
karlkfi, line 124: Is this indirection necessary?

karlkfi:
Looks alright. Just one nit. 

The spec check is important. So thanks for adding it. Current means the status reflects the spec. So it needs to account for independent spec and status changes in both directions.

pkg/kstatus/status/util.go
smiller333: I was just trying to maintain a consistent pattern with what I assumed was the original function, GetStringField, in all of the functions.  If that isn't important and the extra indirection causes confusion, then I can remove it.

smiller333:
> The spec check is important. So thanks for adding it. Current means the status reflects the spec. So it needs to account for independent spec and status changes in both directions.

That makes sense.  Thanks for the clarification.

PR #671, opened 2025-07-05
Modernize

ash2k:
https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize.

There are two commits, please see them separately.


sdowell (commented):
/lgtm

Can this be plugged into the linter to enforce the constraint(s) on future changes?

ash2k:
Looks like it's not feasible to use the code directly: https://github.com/golangci/golangci-lint/discussions/5466

ash2k:
PTAL @seans3 @mortent

PR #673, opened 2025-07-15
Informer cleanup

ash2k:
Please see commits.

I'm proposing to make a breaking change to signatures of some functions to remove context since it is passed from the informer itself now.

@karlkfi @seans3 @mortent WDYT?


karlkfi:
FWIW, I'm not at Google any more, but the standing rule was basically that API changes are fine at the library level, as long as they result in compile errors that are easily corrected by downstream users, as long as the behavior is the same. 

This looks solid. Thanks.

PR #675, opened 2025-09-10
Move off deprecated code

ash2k:
/kind cleanup

### DIFF
--- a/pkg/apply/filter/current-uids-filter.go
+++ b/pkg/apply/filter/current-uids-filter.go
@@ -16,7 +16,7 @@ import (
 // if an object should not be pruned (deleted) because it has recently
 // been applied.
 type CurrentUIDFilter struct {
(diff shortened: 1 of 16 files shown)


pkg/apply/filter/current-uids-filter.go
```diff
@@ -16,7 +16,7 @@ import (
 // if an object should not be pruned (deleted) because it has recently
 // been applied.
 type CurrentUIDFilter struct {
-	CurrentUIDs sets.String // nolint:staticcheck
+	CurrentUIDs sets.Set[types.UID] // nolint:staticcheck
```
ash2k, line 19: Here and in a few other places I changed the types. This is a breaking change but I think it should be fine to make:

- Code will fail to compile, the fix is obvious.
- Moving off from the deprecated type.
- New code is more type safe (`string` -> `types.UID`)

sdowell (commented):
/lgtm

PR #676, opened 2025-10-13
feat: add support to prevent updates on objects

sdowell:
This change adds a new actuation feature to the applier which allows objects to be created but not updated. This is useful in cases where the user wants to create the object if it does not exists, but then allow other clients to make updates to the object that are not reverted by the applier.

The feature is used by providing the ignore mutation annotation on the declared object.


ash2k (approved):
/lgtm

PR #680, opened 2026-09-16
fix ParseObjMetadata rejecting rbac names with underscores

### LINKED ISSUE #487: [Bug]: pkg/object/ParseObjMetadata incorrectly assumes rbac resource cannot have underscore in its name

magic-peach:
Fixes #487

RBAC resource names are allowed to contain a literal underscore, but `ParseObjMetadata`'s extra field check did not account for that, only for the double underscore encoding used for colons. Any RBAC role or binding with a literal underscore in its name failed to parse with a "too many fields" error.

There was a previous attempt at this in #645, which went stale and was auto closed. karlkfi raised a concern there that skipping the check for RBAC objects could break round tripping for RBAC names that also contain a colon encoded as a double underscore. This change adds a test for exactly that case, a name with both an underscore and a colon, confirming it round trips correctly through `String()` and back through `ParseObjMetadata`, since the colon transcoding is resolved before the RBAC check runs.

Testing:
- go build ./...
- go test ./pkg/object/... (full package, all tests pass including the new RBAC underscore, colon, and round trip cases)


skitt:
See https://kubernetes.slack.com/archives/C01672LSZL0/p1789992857325439

/close

